PR #2, opened 2015-09-02
[SD-3031] Updates of images from Reuters causes media entries to beco…

marwoodandrew:
…me orphaned

It seems that the Reuters feed repeatedly ingests the same images, in this case the image would be downloaded again and the ingest item replaced in mongo. The old Grid FS entry would be orphaned.

I did some testing downloading the image with the update, calculating the md5 and downloading again if different. It never was so I simplified the code just to check if we already had the image.

### DIFF
--- a/superdesk/io/commands/update_ingest.py
+++ b/superdesk/io/commands/update_ingest.py
@@ -350,14 +350,14 @@ def ingest_item(item, provider, rule_set=None, routing_scheme=None):
         if item.get('ingest_provider_sequence') is None:
             ingest_service.set_ingest_provider_sequence(item, provider)
 
+        old_item = ingest_service.find_one(guid=item[GUID_FIELD], req=None)
+
         rend = item.get('renditions', {})
         if rend:
             baseImageRend = rend.get('baseImage') or next(iter(rend.values()))
             if baseImageRend:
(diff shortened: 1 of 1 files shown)


superdesk/io/commands/update_ingest.py
```diff
@@ -382,11 +382,32 @@ def ingest_item(item, provider, rule_set=None, routing_scheme=None):
...
     inserted = []
     try:
+        # If there is an existing set of renditions we keep those
+        if old_item:
+            media = old_item.get('renditions', {}).get('original', {}).get('media', {})
+            if media:
+                item['renditions'] = old_item['renditions']
+                item['mimetype'] = old_item.get('mimetype')
+                item['filemeta'] = old_item.get('filemeta')
+                logger.info("Reuters image not updated for GUID:{}".format(item[GUID_FIELD]))
```
amagdas: Is this specific only to Reuters or the message can be more generic?

amagdas:
:+1:

superdesk/io/commands/update_ingest.py
marwoodandrew: Only Reuters at the moment, but probably should be generic, it is core :-)

mdhaman:
:+1:

petrjasek:
:+1:

PR #4, opened 2015-09-02
Move Amazon S3 storage to core

amagdas:


mugurrus:
:+1:

petrjasek:
an idea - maybe we can put those test files into `tests` folder outside of `superdesk` folder, when you use the library you probably don't need those (i mean the actual unit tests, not the tests setup)

amagdas:
Yup, it makes sense, will do it in next PR.

amagdas:
Actually I will add to this PR.

amagdas:
@petrjasek done

petrjasek:
ok only not sure how it works with `find_packages` in setup, I think it should be updated

petrjasek:
to `find_packages(exclude=['tests'])`

amagdas:
Right, too sleepy

petrjasek:
:+1:

PR #5, opened 2015-09-04
(refactor) Modified content expiry to rely on the expiry date.

mdhaman:
There is no need to rely on the versioncreated for expiry.
All items ingested have expiry https://github.com/superdesk/superdesk-core/blob/master/superdesk/io/commands/update_ingest.py#L337

Test are in the Superdesk repo


amagdas:
:+1:

petrjasek:
:+1:
would be good to move that test in this repo at some point

PR #6, opened 2015-09-05
(feat) [SD-3036] Publishing embargo content

sivakuna-aap:


sivakuna-aap:
@amagdas can we merge this PR. I need this commit in Superdesk PR without which it will fail. Can we grant write permissions to Aus team on this repo?

PR #7, opened 2015-09-07
(fix) Fixing the issue where you cannot create/update ingest providers.

mdhaman:
Related to https://github.com/superdesk/superdesk/pull/1079

### DIFF
--- a/superdesk/io/ingest_provider_model.py
+++ b/superdesk/io/ingest_provider_model.py
@@ -10,6 +10,7 @@
 
 import logging
 from flask import g
+from eve.utils import config
 from superdesk.errors import SuperdeskApiError
 from superdesk.resource import Resource
 from superdesk.services import BaseService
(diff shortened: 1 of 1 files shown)


superdesk/io/ingest_provider_model.py
```diff
@@ -139,7 +140,7 @@ def _set_provider_status(self, doc, message=''):
     def on_create(self, docs):
         for doc in docs:
             if doc.get('content_expiry', 0) == 0:
-                doc['content_expiry'] = self.app.config['INGEST_EXPIRY_MINUTES']
+                doc['content_expiry'] = config.INGEST_EXPIRY_MINUTES
```
amagdas: So the key was not in app.config? If that is the case then I have to fix a lot more cases.

amagdas: @mdhaman any idea what is the root cause?

petrjasek: with `config` it would work also without application context, but that shouldn't happen imo.
maybe it could be more like `app.config` where `app` would be `flask.current_app`

amagdas: I don't see why the current implementation won't work...

petrjasek: having self.app is not thread safe in general, but you're right it shouldn't be an issue here

mdhaman:
We are getting attribute error. `IngestProviderService` object has no attribute `app`
Will change it to app.config

superdesk/io/ingest_provider_model.py
mdhaman: We are getting attribute error. `IngestProviderService` object has no attribute `app`
Will change it to app.config

amagdas:
Closing this as it's superseded by #9

PR #8, opened 2015-09-07
SD-3054 - add Http Push type of delivery to be used by Content API subscriber

mugurrus:


superdesk/publish/http_push.py
```diff
@@ -0,0 +1,156 @@
...
+    headers = {"Content-type": "application/json", "Accept": "application/json"}
+
+    def _transmit(self, queue_item, subscriber):
+        config = queue_item.get('destination', {}).get('config', {})
+
+        try:
+            self._copy_published_media_files(json.loads(queue_item['formatted_item']))
+
+            try:
+                self._post(config.get('resource_url'), queue_item['formatted_item'])
```
amagdas: It's better to post all the time, I don't think we should fetch the existing story from content api and try to patch it. This sort of checks should be in Content API.

superdesk/publish/http_push.py
```diff
@@ -0,0 +1,156 @@
...
+            except IdConflict:
+                response = self._get(config.get('item_url'), queue_item['item_id'])
+                read_item = json.loads(response.read().decode('utf_8'))
+                self._patch(config.get('item_url'), queue_item['item_id'], queue_item['formatted_item'], read_item['_etag'])
+        except Exception as ex:
+            raise PublishHTTPPushError.httpPushError(ex, queue_item.get('destination', {}))
+
+    def _get(self, url, item_id=None):
```
amagdas: Not needed

superdesk/publish/http_push.py
```diff
@@ -0,0 +1,156 @@
...
+        else:
+            raise PublishHTTPPushError.httpPushError(InvalidURL('Invalid HTTP URL %s' % url), config)
+
+    def _copy_published_media_files(self, item):
+        '''Copy the media files for the given item to the publish_items endpoint
+
+        @param item: the item object
+        '''
+        for k, v in item.get('renditions', {}).items():
+            if k == 'original_source':
```
amagdas: 'original_source' is removed, we always serve/host binaries from SD.

superdesk/publish/http_push.py
```diff
@@ -0,0 +1,156 @@
...
+                # if the source is AAP Multimedia then don't copy
+                continue
+
+            del v['href']
+            if not app.media.exists(v['media'], resource='publish_items'):
```
amagdas: This can no longer work, we can't use Mongo prefix to insert the binaries, we need to send them using Http POST.

superdesk/publish/http_push.py
```diff
@@ -0,0 +1,156 @@
...
+        request_client = self._request_client(url)
+        request_client.request('PATCH', url, formatted_item, headers)
+        response = request_client.getresponse()
+        if response.status == 404:
+            raise ItemNotFound()
+        if response.status < 200 or response.status >= 300:
+            raise Exception('Error pushing item %s' % response.reason)
+        return response
+
+    def _request_client(self, url):
```
petrjasek: imo we can use [requests](http://www.python-requests.org/en/latest/) here it would make it simpler

superdesk/publish/http_push.py
petrjasek: i'm not sure if we should post binaries, imo would be better to send urls where those can be fetched

superdesk/publish/http_push.py
petrjasek: yep, it could be also a `PUT` request, but that can be problematic with proxies, so `POST` is a safe bet.

superdesk/publish/http_push.py
amagdas: Totally agree :+1:

superdesk/publish/http_push_tests.py
```diff
@@ -0,0 +1,91 @@
...
+                                "versioncreated": "2015-03-09T16:32:23",
+                                "version": 1
+                                }
+        self.formatted_item2 = {"_id": "item1",
+                                "headline": "headline2",
+                                "versioncreated": "2015-03-21T13:43:51",
+                                "version": 2
+                                }
+        self.item = {'item_id': 'item1',
+                     'format': 'NITF',
```
amagdas: 'ninjs' instead of 'NITF'?
Also it would be good to test image publishing as well.

superdesk/publish/http_push_tests.py
```diff
@@ -0,0 +1,91 @@
...
+        '''
+        response = requests.get(self.getItemURL(item_id))
+        if response.status_code == requests.codes.not_found:  # @UndefinedVariable
+            return False
+        self.assertEqual(response.status_code, requests.codes.ok,  # @UndefinedVariable
+                         'Error retrieving item from the content API')
+        return True
+
+    def getItemURL(self, item_id):
+        '''Returns the URL for item read
```
sivakuna-aap: Doc String should start with `"`

superdesk/publish/http_push_tests.py
```diff
@@ -0,0 +1,91 @@
...
+                     'published_seq_num': 1,
+                     'formatted_item': json.dumps(self.formatted_item1),
+                     'destination': {"name": "test", "delivery_type": "http_push", "format": "ninjs",
+                                     "config": {"resource_url": self.resource_url}
+                                     }}
+
+    def is_item_published(self, item_id):
+        '''Return True if the item was published, False otherwise.
+        Raises Exception in case of server/communication error.
+        '''
```
sivakuna-aap: Doc String should start with `"`

superdesk/publish/http_push_tests.py
sivakuna-aap: What about packages?

superdesk/publish/http_push.py
```diff
@@ -0,0 +1,67 @@
...
+            raise PublishHTTPPushError.httpPushError(e, queue_item.get('destination', {}))
+
+        resource_url = queue_item.get('destination', {}).get('config', {}).get('resource_url')
+        response = requests.post(resource_url, json=item)
+        if response.status_code != requests.codes.created:  # @UndefinedVariable
+            raise PublishHTTPPushError.httpPushError(Exception('Error pushing item %s' % response.text),
+                                                     queue_item.get('destination', {}))
+
+    def _copy_published_media_files(self, item, assets_url):
+        '''Copy the media files for the given item to the publish_items endpoint
```
sivakuna-aap: Same as below. Doc Strings start with `"`

superdesk/publish/http_push.py
```diff
@@ -0,0 +1,67 @@
...
+from superdesk.publish.publish_service import PublishService
+
+
+errors = [PublishHTTPPushError.httpPushError().get_error_description()]
+
+
+class HTTPPushService(PublishService):
+    """HTTP Publish Service."""
+    headers = {"Content-type": "application/json", "Accept": "application/json"}
+
```
sivakuna-aap: Please add doc strings for `_transmit()` and `_media_exists()`

mugurrus: done

amagdas:
:+1:

sivakuna-aap:
:+1:

PR #12, opened 2015-09-11
(feat) Set Urgency and Priority for an item.

mdhaman:
SD-2627, SD-3159 and SD-3160
- Urgency values will be 1 to 5. 
- NewsValue is no longer needed.
- Priority will be as described in ticket SD-2627

### DIFF
--- a/superdesk/io/__init__.py
+++ b/superdesk/io/__init__.py
@@ -112,6 +112,18 @@ class Parser(metaclass=ParserRegistry):
         if text:
             item['dateline']['text'] = text
 
+    def map_priority(self, source_priority):
+        """
+        Maps the source priority to superdesk priority
+        :param str source_priority:
+        :return int: priority of the item
+        """
+        if source_priority and source_priority.isdigit():
+            if int(source_priority) in [1, 2, 3]:
+                return int(source_priority)
+
+        return 5
+
 
 def get_xml_parser(etree):
     """Get parser for given xml.
--- a/superdesk/io/newsml_1_2.py
+++ b/superdesk/io/newsml_1_2.py
@@ -39,7 +39,7 @@ class NewsMLOneParser(Parser):
                 item['ingest_provider_sequence'] = parsed_el.text
 
             parsed_el = tree.find('NewsEnvelope/Priority')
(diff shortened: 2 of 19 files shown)


superdesk/io/newsml_1_2.py
```diff
@@ -38,6 +38,9 @@ def parse_message(self, tree, provider):
             if parsed_el is not None:
                 item['ingest_provider_sequence'] = parsed_el.text
 
+            parsed_el = tree.find('NewsEnvelope/Priority')
+            item['Priority'] = self.map_priority(parsed_el.text if parsed_el else None)
```
sivakuna-aap: this should be `item['priority']` not `item['Priority']`. 'p' is capital, should be small.

amagdas:
:+1: except Siva's comment

mdhaman:
@amagdas @petrjasek this PR changes the Priority from `str` to `int`. 
Do you think we need to drop the database?

amagdas:
Might be a good idea.

superdesk/io/__init__.py
```diff
@@ -112,6 +112,18 @@ def set_dateline(self, item, city=None, text=None):
...
             item['dateline']['text'] = text
 
+    def map_priority(self, source_priority):
+        """
+        Maps the source priority to superdesk priority
+        :param str source_priority:
+        :return int: priority of the item
+        """
+        if source_priority and source_priority.isdigit():
+            if int(source_priority) in [1, 2, 3]:
```
amagdas: What's up with 4? :)

sivakuna-aap: Well the requirement for now is upto 3 only. :smirk:.

amagdas: Then what's up with 5 :smile:  ?

sivakuna-aap: anything which doesn't fall between 1-3 is 5 :smirk:

mdhaman: Verified with Lasky, everything in the cyber system goes out as 5. 1-3 priority level are introduced in Superdesk and rest is 5. :smirk:

PR #14, opened 2015-09-14
feat(service): add `service.find` method

petrjasek:
it can be used to query service resource python dict query,
without creating request and dumping json.


amagdas:
:+1:

sivakuna-aap:
:+1:

PR #15, opened 2015-09-15
[SD-3211] parse email with json body, part of ingesting from google f…

marwoodandrew:
…orms

This change is to allow the ingestion of email where the body is a JSON document constructed from a Google Form/Sheet and Email notification.

Unit tests will be included in a PR to the non core repo.

Maybe this is none core functionality, should it have been a separate none core ingestor?

### DIFF
--- a/superdesk/io/rfc822.py
+++ b/superdesk/io/rfc822.py
@@ -12,13 +12,13 @@
 import superdesk
 from superdesk.io import Parser
 import datetime
-from superdesk.metadata.item import ITEM_TYPE, CONTENT_TYPE, GUID_TAG
+from superdesk.metadata.item import ITEM_TYPE, CONTENT_TYPE, GUID_TAG, SIGN_OFF
 from superdesk.metadata.utils import generate_guid
 from superdesk.utc import utcnow
 from pytz import timezone
 from superdesk.media.media_operations import process_file_from_stream
 import io
-from flask import current_app as app
+from flask import current_app as app, json
 import email
 from email.header import decode_header
 import logging
@@ -26,7 +26,7 @@ from superdesk.errors import IngestEmailError
 from bs4 import BeautifulSoup, Comment, Doctype
 import re
 from eve.utils import ParsedRequest
-
+from .iptc import subject_codes
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


superdesk/io/rfc822.py
```diff
@@ -304,3 +308,96 @@ def safe_css(self, attr, css):
...
+                            if mail_item.get('Urgency') != '':
+                                item['urgency'] = int(mail_item.get('Urgency'))
+
+                            # We expect the username passed coresponds to a superdesk user
+                            lookup = superdesk.get_resource_service('users').find_one(
+                                req=ParsedRequest(), email=item['original_source'])
+                            if not lookup:
+                                raise UserNotRegisteredException()
+                            item['original_creator'] = lookup.get('_id')
+                            item['byline'] = lookup.get('display_name')
```
sivakuna-aap: This will ignore the byline if it's been set in user profile. should be `lookup.get('byline', lookup.get('display_name'))`.

superdesk/io/rfc822.py
```diff
@@ -304,3 +308,96 @@ def safe_css(self, attr, css):
...
+                            mail_item = dict((k, v[0]) for k, v in json.loads(json_str).items())
+
+                            self._expand_category(item, mail_item)
+
+                            item['original_source'] = mail_item.get('Username')
+                            item['headline'] = mail_item.get('Headline')
+                            item['abstract'] = mail_item.get('Abstract')
+                            item['slugline'] = mail_item.get('Slugline')
+                            item['body_html'] = mail_item.get('Body')
+                            item[SIGN_OFF] = mail_item.get('Sign off')
```
sivakuna-aap: Is `Sign Off` mandatory in Google Forms? If not then it should be based on user profile similar to whats being done for `byline`.

superdesk/io/rfc822.py
```diff
@@ -304,3 +308,96 @@ def safe_css(self, attr, css):
...
+
+                            if mail_item.get('Priority') != '':
```
sivakuna-aap: Explicit checking is not required. Empty string evaluates to False and non-empty to True. Hope the values are pre-defined in Google Forms.

marwoodandrew: The problem is the conversion to integer fails for empty strings

superdesk/io/rfc822.py
marwoodandrew: True

amagdas:
:+1:

amagdas:
Ping @superdesk/developers

mugurrus:
:+1:

PR #16, opened 2015-09-16
[SD-3212] - Preserve desk and stage in ingested items if exists

akintolga:


superdesk/metadata/item.py
```diff
@@ -308,6 +308,16 @@
         'type': 'string'
     },
 
+    # Desk and Stage Details
```
sivakuna-aap: This already exists in `common.py.item_schema()`. This will be overwritten by archive resource as we do `metadata_schema.update(item_schema())`.https://github.com/superdesk/superdesk/blob/master/server/apps/archive/common.py#L385

akintolga: I think that field will be actively used in "archive" collection whereas this field will be available in ingest. But what I can do is to move the 'task' field from archive common to metadata/item. But with this shape overwrite is fine.

sivakuna-aap: If ingest also needs these properties then it's better to move from common to metadata/item

amagdas:
:+1:

amagdas:
Ping @superdesk/developers @sivakuna-aap @marwoodandrew @mdhaman

sivakuna-aap:
I think `common.py.item_schema()` needs to be modified.

amagdas:
Yup, but that should be done in  another PR against SD repo.

sivakuna-aap:
:+1:

PR #18, opened 2015-09-18
fix(formatted email) do case insensitive user lookup on email and rep…

marwoodandrew:
…lace newlines with breaks in body

A couple of things that came out of running it in a real environment!!!

### DIFF
--- a/superdesk/io/rfc822.py
+++ b/superdesk/io/rfc822.py
@@ -380,7 +380,7 @@ class rfc822Parser(Parser):
                             item['headline'] = mail_item.get('Headline')
                             item['abstract'] = mail_item.get('Abstract')
                             item['slugline'] = mail_item.get('Slugline')
-                            item['body_html'] = mail_item.get('Body')
+                            item['body_html'] = mail_item.get('Body').replace('\n','<br />')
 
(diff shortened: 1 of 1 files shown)


superdesk/io/rfc822.py
```diff
@@ -388,8 +388,8 @@ def _parse_formatted_email(self, data, provider):
                                 item['urgency'] = int(mail_item.get('Urgency'))
 
                             # We expect the username passed coresponds to a superdesk user
-                            user = superdesk.get_resource_service('users').find_one(
-                                req=ParsedRequest(), email=mail_item.get('Username'))
+                            query = {'email': re.compile('^{}$'.format(mail_item.get('Username')), re.IGNORECASE)}
+                            user = superdesk.get_resource_service('users').find_one(req=None, **query)
                             if not user:
                                 logger.error('Failed to find user for email {}'.format(mail_item.get('Username')))
                                 raise UserNotRegisteredException()
```
sivakuna-aap: Better to include the same message for the exception as in logger.

sivakuna-aap:
:+1:

amagdas:
:+1:

PR #19, opened 2015-09-18
Update Ingest: include routing rules' content filters in fetched routing scheme

plamut:
This PR is needed for the upcoming refactoring of content filters (ticket [SD-3136](https://dev.sourcefabric.org/browse/SD-3136)).

~~**IMPORTANT:** this is a WIP, please do not merge it just yet.
( @amagdas can you please add WIP label for this repo, too? I don't have permissions, thanks!)~~

### DIFF
--- a/superdesk/io/commands/update_ingest.py
+++ b/superdesk/io/commands/update_ingest.py
@@ -104,8 +104,40 @@ def get_provider_rule_set(provider):
 
 
 def get_provider_routing_scheme(provider):
-    if provider.get('routing_scheme'):
-        return superdesk.get_resource_service('routing_schemes').find_one(_id=provider['routing_scheme'], req=None)
+    """Returns the ingests provider's routing scheme configuration.
+
+    If provider has a routing scheme defined (i.e. scheme ID is not None), the
+    scheme is fetched from the database. If not, nothing is returned.
+
+    For all scheme rules that have a reference to a content filter defined,
+    that filter's configuration is fetched from the database as well and
+    embedded into the corresponding scheme rule.
+
+    :param dict provider: ingest provider configuration
+
+    :return: fetched provider's routing scheme configuration (if any)
+    :rtype: dict or None
+    """
+
+    if not provider.get('routing_scheme'):
+        return None
+
+    schemes_service = superdesk.get_resource_service('routing_schemes')
+    filters_service = superdesk.get_resource_service('content_filters')
+
+    scheme = schemes_service.find_one(_id=provider['routing_scheme'], req=None)
+
+    # for those routing rules that have a content filter defined,
+    # get that filter from DB and embed it into the rule...
+    rules_filters = (
+        (rule, str(rule['filter']))
+        for rule in scheme['rules'] if rule.get('filter'))
+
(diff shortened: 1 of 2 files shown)


superdesk/io/commands/update_ingest.py
```diff
@@ -104,8 +104,40 @@ def get_provider_rule_set(provider):
...
+    :param dict provider: ingest provider configuration
+
+    :return: fetched provider's routing scheme configuration (if any)
+    :rtype: dict or None
+    """
+
+    if not provider.get('routing_scheme'):
+        return None
+
+    schemes_service = superdesk.get_resource_service('routing_schemes')
```
amagdas: Where are this services located? If they are referenced from core, they should be in core or this functionality has to be moved out.

sivakuna-aap: I agree with @amagdas

plamut: Still at the same place as before? The `routing_schemes` service has not been moved anywhere.

amagdas: Dunno, I asked first :smile:

plamut: Wasn't you who was doing the split? :laughing: That's why I asked you for the insight. ;)

If it is a small thing, I would like to do it in the same PR, but if not, it would be more manageable to open a separate task for it IMO - especially considering the fact that this is an already existing problem, and we can't really resolve everything in a single PR. :)

amagdas: Then let's do it :+1:

amagdas: Also I think it's a good time to do this kind of stuff when we touch specific regions of the code, the clean-up should be a continuous process, not something that happens only from time to time.

plamut: OK, good, I'll see how independent content filters are from the rest of the code, but at a first glance I don't expect too many problems (I cannot say much about the routing schemes, though, I might skip them since you say that incremental changes are also fine).

I will update the PR on Monday.

plamut:
@amagdas Hm, it seems like moving content filters to core cannot be done cleanly. Moving `ContentFilter-` and `FilterCondition-` services and resources (and helper classes) is easy, but not so for their tests.

The tests depend on `test_factory.SuperdeskTestCase`, which is not a part of the core yet, and switching to `superdesk.tests.TestCase` also does not work, even if removing `with self.app.app_context()` stuff from test cases - the tests have other dependencies on the stuff outside the core.

At that point the things started to look less clean and straightforward than anticipated, and moving to  core would represent a task on its own, thus I abandoned it for now.

Is there something that I am missing that would make the migration easier or should be skip moving `content_filters` to core for now?

amagdas:
Ok, let's get this merged as it's not adding mode code debt, we should create a separate task to handle the move of `ContentFilter` to core.
:+1:

plamut:
Thanks for the quick feedback!

I'll merge then, as soon as it gets another :+1:

akintolga:
:+1:

PR #23, opened 2015-09-24
[SD-3281] - Removing unreferenced files from S3 bucket

akintolga:


superdesk/commands/clean_images.py
```diff
@@ -35,13 +36,12 @@ def run(self):
             self.__add_existing_files(used_images, upload_items)
 
             print('Number of used files: ', len(used_images))
-            current_files = superdesk.app.media.fs('upload').find({'_id': {'$nin': list(used_images)}})
 
-            for file_id in (file._id for file in current_files if str(file._id) not in used_images):
-                print('Removing unused file: ', file_id)
-                superdesk.app.media.delete(file_id)
+            if isinstance(superdesk.app.media, AmazonMediaStorage):
```
amagdas: I think it's better to move each compare function to the belonging media storage.
The call would look like: `superdesk.app.media.compare_files(used_images)`, this will remove the need for type checking here.

superdesk/storage/amazon/amazon_media_storage.py
```diff
@@ -16,6 +16,7 @@
 import boto3
 
 logger = logging.getLogger(__name__)
+MAX_KEYS = 10000000
```
amagdas: Fetching so many items from S3 could be problematic, better to use batching.

sivakuna-aap:
:+1:

mugurrus:
Adrian's proposal makes for cleaner code

superdesk/commands/clean_images.py
akintolga: That's correct, I'll do that.

superdesk/storage/amazon/amazon_media_storage.py
akintolga: Yes this is ugly. Maybe it is better to use the date of the item to identify the next batch. Let me check how it is possible.

amagdas:
:+1:

PR #25, opened 2015-09-25
Fix iunique validator (handling input valus with special regex characters)

plamut:
The validator failed to escape the input before compiling it into a regular expression, and this PR fixes the issue.

Resolves [SD-3208](https://dev.sourcefabric.org/browse/SD-3208).


mdhaman:
:+1:

akintolga:
:+1:

PR #27, opened 2015-09-28
Retry failed published items

amagdas:


superdesk/publish/publish_content.py
```diff
@@ -51,7 +51,7 @@ def publish():
 
 
 def get_queue_items():
-    lookup = {'state': 'pending', 'destination.delivery_type': {'$ne': 'pull'}}
+    lookup = {'state': {'$in': ['pending', 'in-progress']}, 'destination.delivery_type': {'$ne': 'pull'}}
```
sivakuna-aap: Ideally if transmission has failed then the status should be changed to failed instead of keeping it in-progress. Keeping it in-progress the query will bring the items which haven't failed also and might end up transmitting it twice.

amagdas: I know but I still preferred this option because patching the item in `publish_queue` could fail so that item won't be retried at all.
I'll try your suggestion as well, although this flow is quite prone to race conditions.

amagdas:
@sivakuna-aap Updated

mugurrus:
:+1:

plamut:
This is just a refactoring commit, right, with tests already in place?

Looks good IMO. :+1:

sivakuna-aap:
:+1: Function deserves a doc string. :wink:

plamut:
I agree with that. :wink:

akintolga:
A quick question: It looks like failed items will keep trying every time. There might be a case where we may want to not to try again as there might be a known issue with one of the transmissions. One example would be: say we know we have an issue with our mail server and we don't want those email transmissions to keep trying. How do we do this?

plamut:
This is something similar to the socket reconnect interval in the other pull request. We can implement a logic that, after a few failed retries, gradually increases the retry interval (configurable), and resets it back to the initial interval value upon successful action.

amagdas:
Actually a proper solution would be to not copy the destination on each item but instead reference existing destinations.
By doing this, we can disable that `transport` until we have  a proper connection in place.
Another advantage of doing this is that all the untransmitted content will go to the proper place even if we change the target URI.

akintolga:
I agree with you @amagdas  and @plamut  they are potential solutions. I think this version is good for the time being and we can address this  at another time. Thanks.
:+1:

PR #32, opened 2015-10-07
[SD-3079] - Transmitter not found issue

akintolga:


sivakuna-aap:
:+1:

mdhaman:
:+1:

akintolga:
"transmitters" was creating a name clash so renamed the dictionary.

superdesk/publish/publish_content.py
```diff
@@ -71,17 +71,17 @@ def transmit_items(queue_items):
...
         except Exception as ex:
             logger.exception(ex)
-            failed_items.append(str(queue_item.get('_id')))
+            failed_items[str(queue_item.get('_id'))] = queue_item
 
     if len(failed_items) > 0:
-        for item_id in failed_items:
-            get_resource_service(PUBLISH_QUEUE).system_update(item_id, {'state': STATE_PENDING})
-        logger.error('Failed to publish the following items: {}'.format(failed_items))
+        for item_id in failed_items.keys():
```
plamut: We could also say `for item_id, item in failed_items.items()` and then use `item` in the loop body instead of retrieving it again (`failed_items[item_id]`) - but since it only occurs in one place, the readability would not really be improved that much, thus it's probably not worth updating the PR just for this...

plamut:
:+1:

PR #33, opened 2015-10-07
(fix): Fixing the mongo query so that query is created in the order of field specified.

mdhaman:


sivakuna-aap:
:+1:

plamut:
IMO a test should be added (but I'm not sure what the official project policy is).

:+1:

akintolga:
:+1:

PR #34, opened 2015-10-08
(feat) Adding the base enum.

mdhaman:


sivakuna-aap:
:+1:

superdesk/utils.py
```diff
@@ -35,6 +35,32 @@ class SortOrder(Enum):
...
+        except ValueError:
+            return None
+
+    @classmethod
+    def values(cls):
+        """
+        Returns list of values for an enum
+        :return: list of values for an enum
+        """
+        return [enum_member.value for enum_member in cls.__members__.values()]
```
plamut: Possible simplification: `cls.__members__.values()` --> `cls`

tests/enum_test.py
```diff
@@ -0,0 +1,32 @@
...
+# at https://www.sourcefabric.org/superdesk/license
+
+from superdesk.tests import TestCase
+from superdesk.utils import SuperdeskBaseEnum
+
+
+class EnumToTest(SuperdeskBaseEnum):
+    red = 'red'
+    blue = 'blue'
+    green = 'green'
```
plamut: Maybe just change the member values to something different than their names - just to make more sure that we are indeed testing the values themselves and not the corresponding member names.

plamut:
:+1:

superdesk/utils.py
```diff
@@ -35,6 +35,32 @@ class SortOrder(Enum):
     desc = 2
 
 
+@unique
```
petrjasek: imo this might be limiting on a base class, maybe we can leave this to the concrete implementation?

plamut: Can we think of any use cases where we would want non-unique values?
But there is a point in this, and we can as well remove this restriction from the base class, yes (though it's trivial to lift the restriction later, if needed).

Personally I'm indifferent here, both options seem fine...

tests/enum_test.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_enum_from_value(self):
+        self.assertEqual(EnumToTest.from_value(1), EnumToTest.red)
+        self.assertIsNone(EnumToTest.from_value(4))
+
+    def test_enum_values(self):
+        values = EnumToTest.values()
+        self.assertIn(1, values)
+        self.assertIn(2, values)
+        self.assertIn(3, values)
+        self.assertNotIn(4, values)
```
plamut: Idea: if we decide to keep the `@unique` restriction, we might as well want to add a test  for this, verifying that it is not possible to create a sub-enum with duplicate values.

superdesk/utils.py
mdhaman: @petrjasek it might be limiting. 

I will remove `@unique` decorator as it is trivial to remove.

tests/enum_test.py
mdhaman: As per the docs, `Subclassing an enumeration is allowed only if the enumeration does not define any members.` [See section 8.13.9](https://docs.python.org/3/library/enum.html)
@plamut do you think we should write a unit test to check uniqueness?

sivakuna-aap: In other comment it's been mentioned that `@unique` will be removed. After that this will not be valid any more.

plamut: @mdhaman Since the restriction is going to be removed from the base class, such a test would even be incorrect, thus no need to write it. :)

mdhaman: Fixed in PR https://github.com/superdesk/superdesk-core/pull/36.

@plamut @sivakuna-aap that was already implied in my comment. Point I was trying to get across is that, `@unique` is standard library functionality. If `unique decorator` is defined and `enum` has duplicate enumeration values you are going to get run-time errors. The @unique attribute `should` work as expected and no point testing the class definition in this scenario. :)

plamut: And just for completeness - another view was that using the `@unique` decorator is an implementation detail of the custom Enum subclass, meaning that a test would not test the correctness of the `@unique` validator itself, but rather whether the resulting class has the desired behavior (i.e. unique member values only). It just happens that this would be achieved with a built-in decorator in this hypothetical case. :)

In any case, both ways are valid IMO, it just depends how one looks on it.

PR #35, opened 2015-10-09
(feat & imp) SD-3183 and Vocabulary UI Improvements

sivakuna-aap:
1. [SD-3183](https://dev.sourcefabric.org/browse/SD-3183) Using keywords as labels to group content arbitrarily
2. Adding following features to Vocabularies:
      1. Adding "type" to each vocabulary which defines whether a vocabulary is system driven or manageable by user.
      2. Adding "display_name" to each vocabulary which is used in UI.
      3. Vocabulary UI displays only manageable type vocabularies
      4. Vocabulary UI displays both active and inactive vocabularies
3. Fixed the below issues in Vocabularies UI:
      1. Updating a vocabulary item removes is_active flag from collection
      2. No vocabulary should be allowed to delete as it might have been used in transactional data. So replaced delete with sd-switch

Required by https://github.com/superdesk/superdesk/pull/1232

### DIFF
--- a/superdesk/vocabularies/vocabularies.py
+++ b/superdesk/vocabularies/vocabularies.py
@@ -10,8 +10,9 @@
 
 
 import logging
+import json
 
-from flask import current_app as app
+from flask import request, current_app as app
 from eve.utils import config
 
 from superdesk import privilege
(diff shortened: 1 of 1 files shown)


akintolga:
:+1:

superdesk/vocabularies/vocabularies.py
```diff
@@ -53,9 +63,15 @@ def on_replace(self, document, original):
 
     def on_fetched(self, doc):
         """
-        Overriding to filter out inactive vocabularies and pops out 'is_active' property from the response.
+        Overriding to filter out inactive vocabularies and pops out 'is_active' property from the response if the
+        request wasn't for manageable vocabularies.
         """
 
+        if request and hasattr(request, 'args') and request.args.get('where'):
```
plamut: If you want, you can also write this more concisely as:
```py
if getattr(request, 'args', {}).get('where'):
```

plamut:
:+1:

PR #38, opened 2015-10-14
fix(websockets): remove client when connection is closed

petrjasek:
in case connection was closed by client (and not lost) it is using different callback which was not handled before. plus heartbeat is now part of server factory

### DIFF
--- a/superdesk/ws.py
+++ b/superdesk/ws.py
@@ -21,13 +21,11 @@ from autobahn.asyncio.websocket import WebSocketServerFactory
 from logging.handlers import SysLogHandler
 from logging import Formatter
 
-
 beat_delay = 30
 
 
 def configure_logging(config):
     debug_log_format = ('%(levelname)s:%(module)s:%(message)s\n')
-
     logger = logging.getLogger(__name__)
     logger.setLevel(logging.INFO)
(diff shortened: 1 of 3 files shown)


superdesk/ws.py
```diff
@@ -54,10 +52,15 @@ def onMessage(self, payload, isBinary):
...
     def connectionLost(self, reason):
-        """Unregister on connection drop"""
+        """Unregister client on lost connection."""
         WebSocketServerProtocol.connectionLost(self, reason)
         self.factory.unregister(self)
 
+    def onClose(self, wasClean, code, reason):
+        """Unregister client on closed connection."""
+        WebSocketServerProtocol.onClose(self, wasClean, code, reason)
+        self.factory.unregister(self)
```
amagdas: Should we unregister before calling `onClose` and `connectionLost`? In case of an error being thrown it could remain registered.

amagdas:
Looks good :+1: 
Not sure if my comment is a valid concern.

superdesk/ws.py
petrjasek: i think those are only logging and if those would throw error the whole thing would stop anyhow

amagdas: OK

ioanpocol:
:+1:

PR #43, opened 2015-10-16
Fix teletype formatter

amagdas:


mugurrus:
:+1:

sivakuna-aap:
:+1:

marwoodandrew:
How did I get that wrong?

amagdas:
There is a behave test in superdesk that is failing without this fix. I assume this is the proper use but I could be wrong.

ioanpocol:
:+1:

PR #44, opened 2015-10-19
fix(celery): handle errors with app.context

petrjasek:
when sending notifications on ingest error it was out
of application context and thus raising runtime error.

now all error handing is done in celery task `on_failure`
callback using application context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 
 +## [unreleased]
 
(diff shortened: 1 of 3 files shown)


amagdas:
:+1:

petrjasek:
had to rework this, was still failing on some ingest errors

amagdas:
OK :+1:

mdhaman:
:+1:

PR #49, opened 2015-10-22
[SD-3426] Order the results by name alphabetically:

sivakuna-aap:
[SD-3426](https://dev.sourcefabric.org/browse/SD-3426): Order the results by name alphabetically
Required by https://github.com/superdesk/superdesk/pull/1285

### DIFF
--- a/superdesk/publish/subscribers.py
+++ b/superdesk/publish/subscribers.py
@@ -28,7 +28,9 @@ class SubscribersResource(Resource):
         'name': {
             'type': 'string',
             'iunique': True,
-            'required': True
+            'required': True,
+            'nullable': False,
+            'empty': False
         },
         'media_type': {
             'type': 'string'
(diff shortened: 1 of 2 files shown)


superdesk/publish/subscribers.py
```diff
@@ -107,8 +109,6 @@ class SubscribersResource(Resource):
         }
     }
 
-    datasource = {'default_sort': [('_created', -1)]}
```
mugurrus: This removes default sorting, the results will be unsorted

superdesk/roles/roles.py
```diff
@@ -39,9 +41,7 @@ class RolesResource(Resource):
             'type': 'boolean'
         },
     }
-    datasource = {
```
mugurrus: Same as above

superdesk/publish/subscribers.py
mugurrus: Sorry, I just noticed in the superdesk PR that you do the sorting from the client side.

mugurrus:
:+1:

petrjasek:
:+1:

PR #50, opened 2015-10-22
refactor(notifications): refactor using websockets library

petrjasek:
so this should avoid memory leak, also now it knows when there
is server (pub) or client (sub) connection and only send messages
from servers to clients.

when server can't connect to websocket it will try next time it
will be sending a message.

SD-3385 SD-3480


amagdas:
:+1:

ioanpocol:
:+1:

amagdas:
@petrjasek We need to update the requirements in SD.

petrjasek:
right will do it

PR #51, opened 2015-10-22
refactor(locks): use mongo for task locking

petrjasek:
move away from redis and use mongo instead for locking.
so far it's a poc in `ingest_update` task.

seems to work, not sure if we would need any extra wrapping for the library.

cc @mdhaman @akintolga @amagdas

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -38,3 +38,4 @@ statsd==3.1
 httmock==1.2.3
 boto3==1.0.0
 websockets==2.6
+mongolock==1.3.4
--- a/superdesk/celery_task_utils.py
+++ b/superdesk/celery_task_utils.py
@@ -8,6 +8,7 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
+import os
 import logging
 from flask import current_app as app
 from superdesk.utc import utcnow, get_date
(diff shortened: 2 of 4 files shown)


sivakuna-aap:
IMO, it would be good if we abstract the locking and unlocking functions in a module/class and developers use those functions instead of directly invoking the Mongolock API methods. Also, if tomorrow we need replace Mongolock with some other framework then we can do it at one place rather touching all python modules.

akintolga:
Looks good. 
Do we need to include the hostname of the worker instance (to the 'lock_name' maybe) as there'll be few nodes?

:+1:

mdhaman:
Agree with Siva for abstracting the Mongolock API. 
:+1:

plamut:
+1 on the suggestion to abstract the locking mechanism into a utility module. A simple facade should do the trick, decoupling the application from the concrete locking implementation (Mongolock in our case).

petrjasek:
thx for feedback. @akintolga it's using task id instead of worker host for identification

amagdas:
Yup, +1 for moving to it's own module.
Also, lock is not really the proper name, as it does actually just a race condition detection, no critical region is acquired.
I think I prefer the names used already:`mark_task_as_not_running, is_task_running`

petrjasek:
updated, what do you think?
@amagdas it can be used for actual locking, so I wouldn't stick to that naming..

amagdas:
Sounds good :+1:

plamut:
Ideally this feature would be implemented as a context manager, probably by just using the [contextlib.contextmanager](https://docs.python.org/3.4/library/contextlib.html) decorator. This would allow the client code to simply say e.g.:
```py
with lock(lock_name, host_name, expire=1800) as acquired_lock:
    # do something
```
No need for wrapping the main code into a `try..finally` and explicitly releasing the lock in the latter, which might get forgotten.

If a lock cannot be created, some kind of an exception could be thrown, or perhaps silently ignored on attempts to acquire an already acquired lock.

It's still good as it is, though, just a bit more error prone and requiring a bit more work in the code using the locks.

mdhaman:
:+1:

akintolga:
:+1:

PR #52, opened 2015-10-22
fix(ci): remove exception logging when connecting to ws

petrjasek:
with stack traces for every unittest it has too long output on travis.

and there should be no need for details, there is another log entry when
message can't be delivered.

### DIFF
--- a/superdesk/notification.py
+++ b/superdesk/notification.py
@@ -43,9 +43,8 @@ def init_app(app):
     try:
         app.ws_client = loop.run_until_complete(websockets.connect('ws://%s:%s/server' % (host, port)))
         logger.info('websocket connected on=%s:%s' % app.ws_client.local_address)
-    except OSError as err:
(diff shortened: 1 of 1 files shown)


akintolga:
:+1:

mdhaman:
:+1:

plamut:
IMO stack traces are useful to quickly identify why a test has failed. Wouldn't it be better to keep them nevertheless, but with exception context suppressed? Then only the last stack frame would be printed out.

What do you guys think?

petrjasek:
@plamut in general yes, but here it's obvious what went wrong - socket is closed - and we handle it via reconnecting later - so I wouldn't try to put it there

amagdas:
LGTM :+1:

plamut:
@petrjasek Aha, OK, in that case stack traces are indeed not needed, no additional value in them.
:+1:

PR #57, opened 2015-10-27
feat(ninjs): add profile and slugline fields

petrjasek:
- adding profile field to item metadata
- making ninjs formatter keep profile and slugline

not sure how gracefuly can content-api handle schema changes, does it require an update?
ping @amagdas @mugurrus @plamut

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 
 +## [unreleased]
 
+- feat(ninjs): output profile and slugline in ninjs
 - fix(celery): use application context for task error handling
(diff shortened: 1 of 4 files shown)


amagdas:
:+1: 
No updates needed to ContentAPI, it will render the new fields correctly.

petrjasek:
hm not sure about it, got 422 response code

amagdas:
Yeah, you are right :(
This needs to be updated:
https://github.com/superdesk/superdesk-content-api/blob/master/content_api/items/resource.py#L14

mugurrus:
:+1:

PR #58, opened 2015-10-28
refactor(subjectcodes): make iptc newscodes optional

petrjasek:
it introduces `app.subjects` for subjects registration
and makes iptc newscodes usage optional

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *#
 .#*
 *TAGS
+*.egg-info
 .ropeproject
 /data
 static/*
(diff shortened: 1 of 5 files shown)


sivakuna-aap:
:+1: LGTM. Can we compute `app.subjects.get_items()` before hand. This avoid from computing it every time  the server receives request for `/subjectcodes`?

amagdas:
:+1:

petrjasek:
@sivakuna-aap we could but then in case there are some changes in those files you need to restart, so I was trying to avoid it. but this whole thing is more like quickfix, imo it should work like other vacabularies later on.

PR #64, opened 2015-11-02
fix(notification): make mock client send method coroutine

petrjasek:
to work like the original method and avoid TypeError


amagdas:
:+1:

vied12:
:ok_hand:

PR #65, opened 2015-11-03
(fix) Reuters ingest routing

marwoodandrew:
This is a fix to stop Reuters content being routed each time the same version of an item is received. There will be a PR against the Superdesk Repo with a test and further changes


sivakuna-aap:
:+1:

akintolga:
:+1:

petrjasek:
:+1:

PR #66, opened 2015-11-04
(fix) [SD-3599] Disabled users should not get error notifications

sivakuna-aap:


mugurrus:
:+1:

superdesk/activity.py
```diff
@@ -215,9 +215,10 @@ def notify_and_add_activity(activity_name, msg, resource=None, item=None, user_l
     add_activity(activity_name, msg=msg, resource=resource, item=item,
                  notify=[user.get("_id") for user in user_list] if user_list else None, **data)
     if user_list:
-        recipients = [user.get('email') for user in user_list if
-                      superdesk.get_resource_service('preferences').
-                      email_notification_is_enabled(preferences=user.get('user_preferences', {}))]
+        recipients = [user.get('email') for user in user_list
+                      if user.get('is_enabled', False) and user.get('is_active', False) and
```
mugurrus: A test for this would be helpful

ioanpocol:
:+1:  (is needed a small test to prove that notifications are not sent anymore to disabled users)

superdesk/activity.py
petrjasek: maybe this filtering should be done in `get_users_by_user_type` method?

sivakuna-aap: `get_users_by_user_type()` is being used in multiple places. Not sure if they need inactive users. That's the reason doing it in recipients.

sivakuna-aap:
The fix has superdesk changes also. I will cover tests in superdesk PR.

superdesk/activity.py
petrjasek: yep I was wondering about that

petrjasek:
:+1:

PR #67, opened 2015-11-04
(fix) Adding genre to the aggregates and remove broadcast subscriber type.

mdhaman:
- Changes index for genre field.
- Removes subscriber type broadcast as they behave same as wire.

### DIFF
--- a/superdesk/metadata/item.py
+++ b/superdesk/metadata/item.py
@@ -136,6 +136,7 @@ metadata_schema = {
     'genre': {
         'type': 'list',
         'mapping': {
+            'type': 'object',
             'properties': {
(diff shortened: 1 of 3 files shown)


akintolga:
:+1:

superdesk/metadata/item.py
```diff
@@ -136,6 +136,7 @@
     'genre': {
         'type': 'list',
         'mapping': {
+            'type': 'object',
```
sivakuna-aap: Should it be `dict` here instead of `object`?

sivakuna-aap:
:+1:

mugurrus:
:+1:

PR #68, opened 2015-11-04
feat(ninjs): add service using `anpa_category` value

petrjasek:
it's working more or less the same, so like this could be more generic

this is for pa atm, but eventually we could rename `anpa_category` to `service` in metadata / ui,
would this work for you? @sivakuna-aap @mdhaman @akintolga @marwoodandrew


petrjasek:
and btw - fixing subject to have `code` not `qcode` to match ninjs spec

akintolga:
:+1:

mugurrus:
:+1:

PR #70, opened 2015-11-05
feat(Amazon S3): Make direct href links public and organize media fil…

ioanpocol:
…e by date

### DIFF
--- a/superdesk/storage/amazon/amazon_media_storage.py
+++ b/superdesk/storage/amazon/amazon_media_storage.py
@@ -14,6 +14,7 @@ import logging
 import json
 from io import BytesIO
 import boto3
+import time
 
 logger = logging.getLogger(__name__)
 MAX_KEYS = 1000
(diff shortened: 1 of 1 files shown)


plamut:
Do we test these methods?
The change itself seems fine.

superdesk/storage/amazon/amazon_media_storage.py
```diff
@@ -59,11 +60,20 @@ def url_for_media(self, media_id):
...
+
+        return '%s/%s' % (time.strftime('%Y%m%d'), media_id)
+
     def read_from_config(self):
         self.region = self.app.config.get('AMAZON_REGION', 'us-east-1') or 'us-east-1'
         username = self.app.config['AMAZON_ACCESS_KEY_ID']
         api_key = self.app.config['AMAZON_SECRET_ACCESS_KEY']
         self.container_name = self.app.config['AMAZON_CONTAINER_NAME']
+        self.kwargs = {}
+        if self.app.config.get('AMAZON_SERVE_DIRECT_LINKS', False):
```
petrjasek: maybe this should be `PUBLIC_LINKS` not `DIRECT_LINKS` it would be more obvious imo

petrjasek:
:+1: some tests would be great (as long as it would talk to real s3)

superdesk/storage/amazon/amazon_media_storage.py
ioanpocol: Public links are just first version but the intention is to have direct links but not public.

ioanpocol:
I think that for behave tests we have to make the amazon account credentials visible on github, so because of this there are no tests for this.

petrjasek:
so I suggest define tests talking to s3 using environment vars - in case it's not set do nothing - and we can set those credentials on bamboo.. but that's task of its own, so wouldn't block this meanwhile

ioanpocol:
Added a new task for behave tests: SDPA-41

petrjasek:
@ioanpocol are you going to merge this to pa-poc branch too?

ioanpocol:
I will update the superdek-core version on requirements.txt on both master and pa-poc branches.

PR #73, opened 2015-11-10
feat(crop): return size from crop_image

petrjasek:
returned stream has now `width` and `height` attributes


akintolga:
:+1:

sivakuna-aap:
:+1:

PR #74, opened 2015-11-11
crop updates

petrjasek:
- moving `archive_crop` module into core as `superdesk.crop`
- let `crop_image` also resize to given width/height if provided


sivakuna-aap:
:+1:  Looks Good.  Any specific reason for not having `crop.py` in `superdesk.media`?

petrjasek:
didn't think about that module, it should be there, will update this

ioanpocol:
:+1:

mdhaman:
:+1:

PR #75, opened 2015-11-11
Pa poc headline limit

eozan:


ioanpocol:
:+1:

superdesk/metadata/item.py
```diff
@@ -332,7 +332,8 @@
         'type': 'datetime',
         'versioned': False
     },
-    'lock_session': Resource.rel('auth')
+    'lock_session': Resource.rel('auth'),
+    'template_id': Resource.rel('content_templates', embeddable=False, nullable=True)
```
petrjasek: can you call this `template` ? we already set that field for item created on schedule so it would work together

plamut:
Looks good, just consider @petrjasek's recommendation if it applies.

:+1:

petrjasek:
obsolete

PR #79, opened 2015-11-15
media updates

petrjasek:
- make it possible to specify media id when putting media object
- format renditions when publishing ninjs - remove extra fields

### DIFF
--- a/superdesk/storage/amazon/amazon_media_storage.py
+++ b/superdesk/storage/amazon/amazon_media_storage.py
@@ -11,6 +11,7 @@
 ''' Amazon media storage module'''
 import boto3
 import json
+import bson
 import time
 import logging
 from io import BytesIO
@@ -57,15 +58,15 @@ class AmazonMediaStorage(MediaStorage):
 
     def url_for_media(self, media_id):
         if not self.app.config.get('AMAZON_SERVE_DIRECT_LINKS', False):
-            return upload_url(media_id)
+            return upload_url(str(media_id))
         protocol = 'https' if self.app.config.get('AMAZON_S3_USE_HTTPS', False) else 'http'
-        endpoint = 's3-%s.amazonaws.com' % self.region
-        return '%s://%s.%s/%s' % (protocol, self.container_name, endpoint, self.name_for_media(media_id))
+        endpoint = 's3-%s.amazonaws.com' % self.app.config.get('AMAZON_REGION')
+        return '%s://%s.%s/%s' % (protocol, self.container_name, endpoint, media_id)
 
-    def name_for_media(self, media_id):
+    def media_id(self, filename):
         if not self.app.config.get('AMAZON_SERVE_DIRECT_LINKS', False):
-            return media_id
-        return '%s/%s' % (time.strftime('%Y%m%d'), media_id)
+            return bson.ObjectId()
+        return '%s/%s' % (time.strftime('%Y%m%d'), filename)
 
     def fetch_rendition(self, rendition):
         stream, name, mime = download_file_from_url(rendition.get('href'))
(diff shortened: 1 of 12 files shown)


tests/storage/gridfs_media_storage_test.py
```diff
@@ -0,0 +1,53 @@
...
+        self.app.config['DOMAIN'] = {'upload': {}}
+        self.app.data = SuperdeskDataLayer(self.app)
+        self.media = SuperdeskGridFSMediaStorage(self.app)
+        self.app.register_blueprint(bp)
+
+    def test_media_id(self):
+        filename = 'some-file'
+        media_id = self.media.media_id(filename)
+        self.assertIsInstance(media_id, bson.ObjectId)
+        self.assertEqual(media_id, self.media.media_id(filename))
```
mugurrus: You should supply a calculated value here instead of the result of the call to `self.media.media_id(filename)`

mugurrus:
Needs rebase, otherwise good except the minor comment.

tests/storage/gridfs_media_storage_test.py
petrjasek: it's testing that if you run it 2x, you get the same value

superdesk/storage/amazon/amazon_media_storage.py
```diff
@@ -149,17 +150,17 @@ def put(self, content, filename=None, content_type=None, resource=None, metadata
...
+        _id = _id or self.media_id(filename)
+        found = self._check_exists(_id)
         if found:
-            return filename
+            return _id
 
         try:
             file_metadata = self.transform_metadata_to_amazon_format(metadata)
-            self.client.put_object(Key=filename, Body=content, Bucket=self.container_name,
+            self.client.put_object(Key=_id, Body=content, Bucket=self.container_name,
```
plamut: Why are these keyword parameter capitalized? Is `boto3.client` an external library that defines them like this?

superdesk/utils.py
```diff
@@ -154,3 +155,11 @@ def compare_preferences(original, updates):
     removed = original_keys - updates_keys
     modified = {o: (original[o], updates[o]) for o in intersect_keys if original[o] != updates[o]}
     return added, removed, modified
+
+
+def sha(text):
```
plamut: Maybe rename this method to `sha1` to make it clear in the calling code which SHA hash this function calculates.
We should also **strongly** reconsider if we indeed want to use SHA-1 in the first place - it's considered broken (or at least will be in the near future) based on the most recent findings. We should better move to a different hashing function sooner rather than later.

plamut:
Looks good to me in general, though using SHA1 makes me uneasy. We should address this potential issue, but probably in a separate PR?

superdesk/storage/amazon/amazon_media_storage.py
petrjasek: yep it's external

superdesk/utils.py
petrjasek: I've changed it to use `sha256` instead

plamut: That's great! Perhaps consider renaming the function itself, too, but it looks good.

plamut:
Oh, BTW: :+1:

superdesk/utils.py
petrjasek: I keept sha so that we can change it later to more secure version without affecting all the code..

mugurrus:
:+1:

superdesk/utils.py
plamut: Aha, I see, it does that kind of make sense.
(unless the whole SHA family gets broken :stuck_out_tongue_winking_eye: )

PR #80, opened 2015-11-16
(feat) [SD-3626] Modified schema and added elasticsearch settings to query with phrase prefix

mdhaman:
This PR is dependent on [EVE-ELASTIC PR](https://github.com/petrjasek/eve-elastic/pull/48).

This allows configureing elasticsearch index settings. For example, different analyser and filters.

### DIFF
--- a/tests/elasticsearch_settings_test.py
+++ b/tests/elasticsearch_settings_test.py
@@ -1,0 +1,125 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+import json
+from eve.utils import ParsedRequest
+from superdesk.tests import TestCase
+from superdesk import get_resource_service
+
+
+class ElasticSearchSettingsTest(TestCase):
+    items = [
+        {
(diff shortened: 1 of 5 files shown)


mugurrus:
:+1:

sivakuna-aap:
:+1:

tests/elasticsearch_settings_test.py
```diff
@@ -0,0 +1,130 @@
...
+            '_id': '654', 'headline': 'Test 5', 'slugline': 'England Test Soccer', 'body_html': 'Test'
+        },
+        {
+            '_id': '984', 'headline': 'Test 6', 'slugline': 'Soccer Germany', 'body_html': 'Test'
+        }
+    ]
+
+    def setUp(self):
+        super().setUp()
+        with self.app.app_context():
```
plamut: For convenience, it would probably make sense to push application context on the stack in the `setUp()` and pop it in the `tearDown()` method, avoiding the need to use the same `with` statement over and over again in every test method.

plamut:
Looks good IMO, with a non-essential recommendation.

:+1:

petrjasek:
:+1:

mdhaman:
@petrjasek you are right init_app will fail as it tries to put_mapping. This update petrjasek/eve-elastic#50 should fix your production issues.

On our local instance, I closed the index, applied settings and open index using elasticsearch API. an then did rebuild index. IMO, open/close index should not be done everytime we make software change. I think elasticsearch index changes should be done out of the superdesk application context.

petrjasek:
> I think elasticsearch index changes should be done out of the superdesk application context.
:+1: right

PR #82, opened 2015-11-18
[SD-3539] -  Desk Mentions

akintolga:
Refactoring activity to make it indexable in mongo.

### DIFF
--- a/superdesk/activity.py
+++ b/superdesk/activity.py
@@ -17,6 +17,8 @@ from superdesk.errors import SuperdeskApiError, add_notifier
 import superdesk
 from bson.objectid import ObjectId
 from superdesk.emails import send_activity_emails
+import datetime
+from superdesk.utc import utcnow
 
 log = logging.getLogger(__name__)
 
@@ -118,13 +120,9 @@ class ActivityResource(Resource):
             'schema': {
                 'type': 'dict',
                 'schema': {
-                    'user_id': {'type': 'string',
-                                'required': False,
-                                'nullable': True},
+                    'user_id': Resource.rel('users'),
                     'read': {'type': 'boolean', 'default': False},
-                    'desk_id': {'type': 'string',
-                                'required': False,
-                                'nullable': True}
+                    'desk_id': Resource.rel('desks')
                 }
             }
         },
@@ -135,7 +133,8 @@ class ActivityResource(Resource):
     }
     exclude = {endpoint_name, 'notification'}
     datasource = {
-        'default_sort': [('_created', -1)]
+        'default_sort': [('_created', -1)],
+        'filter': {'_created': {'$gte': utcnow() - datetime.timedelta(days=1)}}
     }
     superdesk.register_default_user_preference('email:notification', {
         'type': 'bool',
@@ -158,7 +157,7 @@ class ActivityService(BaseService):
         user = getattr(g, 'user', None)
         if not user:
             raise SuperdeskApiError.notFoundError('Can not determine user')
-        user_id = str(user.get('_id'))
+        user_id = user.get('_id')
 
         # make sure that the user making the read notification is in the notification list
         if not self.is_recipient(updates, user_id):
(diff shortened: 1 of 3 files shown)


superdesk/activity.py
```diff
@@ -113,7 +113,15 @@ class ActivityResource(Resource):
         'name': {'type': 'string'},
         'message': {'type': 'string'},
         'data': {'type': 'dict'},
-        'read': {'type': 'dict'},
+        'recipients': {
+            'type': 'list',
+            'mapping': {
```
sivakuna-aap: Should be schema as mapping is used by elasticsearch

superdesk/activity.py
```diff
@@ -147,23 +156,38 @@ def on_update(self, updates, original):
...
+        """
+        return any(r for r in activity.get('recipients', []) if r['user'] == user_id)
+
+    def is_read(self, activity, user_id):
+        """
+        Returns the read value for the given user
+        """
+        try:
+            return next(r['read'] for r in activity.get('recipients', []) if r['user'] == user_id)
+        except StopIteration:
```
petrjasek: this part seems extra here given that it checks if user is in list before, maybe it should just return `False`, raising an error from such method is a bit surprising imo, not sure what others think

petrjasek:
:+1:

superdesk/activity.py
```diff
@@ -147,23 +156,38 @@ def on_update(self, updates, original):
...
         # make sure the transition is from not read to read
-        if not (updates.get('read')[user_id] == 1 and original.get('read')[user_id] == 0):
+        if not self.is_read(updates, user_id) and self.is_read(original, user_id):
             raise SuperdeskApiError.forbiddenError('Can not set notification as read')
 
         # make sure that no other users are being marked as read
-        for read_entry in updates.get('read'):
-            if read_entry != user_id:
-                if updates.get('read')[read_entry] != original.get('read')[read_entry]:
+        for recipient in updates.get('recipients'):
```
plamut: Is the `recipients` field always present? If not, a default value should be passed to `get()` as well (e.g. an empty list) to avoid an error.

superdesk/activity.py
plamut: Let's just pass `False` as the second argument to `next()` avoiding the need for catching  the `StopIteration` error.
Returning `False` is IMO better than raising an error, at least that's what I would expect based on the method name. On the other hand, if we do want to raise an error, it should be documented in the docstring, and maybe the method renamed to `check_is_read()` or something.

tests/activity_tests.py
```diff
@@ -0,0 +1,34 @@
...
+        self.assertFalse(ActivityService().is_recipient(activity, '3'))
+
+    def test_is_read(self):
+        activity = {'recipients': [{'user': '1', 'read': True},
+                                   {'user': '2', 'read': False}]}
+
+        self.assertTrue(ActivityService().is_read(activity, '1'))
+        self.assertFalse(ActivityService().is_read(activity, '2'))
+        with assert_raises(SuperdeskApiError):
+            ActivityService().is_read(activity, '3')
```
plamut: I would split these tests into a few more tests, with each testing only _one_ execution path, so that in a case of failure it is known exactly which specific case has failed.

plamut:
Looks good, with a few remarks (one potentially essential - the (non)existence of the `recipients` field).

:+1:

tests/activity_tests.py
ahilles107: :+1:

mdhaman:
:+1:

mugurrus:
:+1:

superdesk/activity.py
```diff
@@ -121,7 +133,8 @@ class ActivityResource(Resource):
     }
     exclude = {endpoint_name, 'notification'}
     datasource = {
-        'default_sort': [('_created', -1)]
+        'default_sort': [('_created', -1)],
+        'filter': {'_created': {'$gte': utcnow() - datetime.timedelta(days=1)}}
```
sivakuna-aap: This filter is applied across all activities but 24hrs filter is specific to Desk Mentions not for other activities.

superdesk/notification.py
```diff
@@ -15,9 +15,9 @@
 import asyncio
 import websockets
 
-from flask import json
 from flask import current_app as app
 from datetime import datetime
+from bson.json_util import dumps
```
sivakuna-aap: any specific reason for moving from flask.json to bson.json_util?

akintolga: I used this to serialize the ObjectId for the user and the desk. For flask.json we'd need an encoder to handle the ObjectId but with this one it is built in.

superdesk/activity.py
akintolga: I have been told that only last 24h of activities are relevant so return to the users so I have put this  in. Also Michael applied capped collection script to Activity collection so that script will be shared with team Europe.

superdesk/notification.py
mdhaman: You can use this `flask.json` with `json_serialize_datetime_objectId` 
see :
https://github.com/superdesk/superdesk-core/blob/master/superdesk/utils.py#L142
https://github.com/superdesk/superdesk/blob/master/server/apps/publish/formatters/aap_bulletinbuilder_formatter.py#L51

PR #85, opened 2015-11-19
RSS ingest: try to populate body text from the "content" field as a fallback

plamut:
Resolves [SD-3561](https://dev.sourcefabric.org/browse/SD-3561).


petrjasek:
:+1:

sivakuna-aap:
:+1:

PR #88, opened 2015-11-24
[SD-3646](feat) Adding functionality to register jinja filters.

mdhaman:
There is requirement for Kill template where we need custom date format.


mugurrus:
:+1:

petrjasek:
:+1:

PR #89, opened 2015-11-25
(imp) [SD-3387] Ingest refactoring

sivakuna-aap:
The basic idea is "to allow user select a delivery method (ftp/email/file/rss/http/etc) and then a data format (nitf.newsml/teletype/anpa/custom/etc) as independent options while creating a Ingest Channel". 

In Superdesk we call delivery method as `Feeding Service` and data format as `Feed Parser`. This PR makes the changes and they have been depicted in the below UML diagrams:

feedingservice
feedingparser
ingestsequence


petrjasek:
looks good, for example - is reuters `ReutersFeedingService` with generic `NewsMLG2FeedParser` ?

mdhaman:
LGTM. 
@petrjasek I think `ReutersFeedingService` can have a generic parser or its own `NewsMLG2` parser. @sivakuna-aap can correct me on this?

sivakuna-aap:
@petrjasek @mdhaman I haven't fully gone through existing reuters. But am planning to rename it as HTTPFeedingService. In case of reuters provider this is how the Ingest Provider configuration should look like: `name: reuters, source: reuters, feeding_service: http, feed_parser: newsml2, config:{} ... `

petrjasek:
I'm just wondering which part is handling its api - I guess it's the feeding service handling authentication, fetching channels, fetching lists of items - right?

sivakuna-aap:
yes @petrjasek, FeedingService will handle the functions you mentioned.

petrjasek:
sounds good. and btw those meta classes - you want to use it for registration? I would rather add some `register_x` methods to make it explicit, it might be hard to disable some of them if it's imported somewhere even if not used..

sivakuna-aap:
Even I prefer to remove the meta classes but didn't want to disturb the existing setup. I prefer to mark the super classes as abstract instead of having a meta class for registration. 

Shall I remove registration meta class and mark the super classes as ABCMeta?

petrjasek:
imo yes, but lets see what others think

akintolga:
:+1: Looks pretty good. For meta classes, I'm also in favor of removing them, that would be "cherry on top".

mdhaman:
+1 for registry

sivakuna-aap:
Closing this PR. Realized that there's more which needs to be moved from superdesk. Will create a new PR with the refactoring and also without meta classes for registration of feeding services and feed parsers.

PR #94, opened 2015-12-03
(fix) [SD-3387] Fixing for Ingest Provider Creation failure. 

sivakuna-aap:
Reason: Registration of Feeding Services and Feed Parsers happen later after registering the Ingest Provider Resource. Thus causing the value of 'allowed' property to be empty.

### DIFF
--- a/superdesk/io/ingest_provider_model.py
+++ b/superdesk/io/ingest_provider_model.py
@@ -18,7 +18,7 @@ from superdesk import get_resource_service
 from superdesk.activity import ACTIVITY_CREATE, ACTIVITY_EVENT, ACTIVITY_UPDATE, notify_and_add_activity, \
     ACTIVITY_DELETE
 from superdesk.errors import SuperdeskApiError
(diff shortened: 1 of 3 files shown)


superdesk/io/ingest_provider_model.py
```diff
@@ -43,13 +43,11 @@ def __init__(self, endpoint_name, app, service, endpoint_schema=None):
             'source': required_string,
             'feeding_service': {
                 'type': 'string',
-                'required': True,
-                'allowed': list(allowed_feeding_services)
```
mugurrus: Why allow invalid values?

sivakuna-aap: Handling invalid values in `on_create` and `on_update` instead of Cerberus.

mugurrus: Sorry, I saw later that you do the validation in on_ methods

mugurrus:
:+1:

akintolga:
:+1:

mdhaman:
:+1:

PR #95, opened 2015-12-04
(fix) [SD-3387] This should fix the create/update of an Ingest Provider.

sivakuna-aap:
1. Replaced Dictionary view objects for allowed_feeding_services and allowed_feed_parsers with normal list as Cerberus doesn't support Dictionary view objects.

2. There might be cases where a feeding service doesn't need a feed parser to parse the content. RSS is an example. That's the reason moved the validation from on_create/on_update to schema declaration instead of handling it in event methods, so that Cerberus can handle it before triggering the service.


akintolga:
:+1:

mdhaman:
:+1:

mugurrus:
:+1:

PR #97, opened 2015-12-04
[SD-3755] Dont use redis keys

actionless:
https://dev.sourcefabric.org/browse/SD-3755

### DIFF
--- a/superdesk/celery_app.py
+++ b/superdesk/celery_app.py
@@ -183,18 +183,3 @@ def _update_subtask_progress(task_id, current=None, total=None, done=None):
         redis_db.delete(done_key)
 
     return task_id, crt_current, crt_total
-
-
-def set_key(key, value=0, db=None):
-    """
-    @TODO: deprecate this?
-    Sets the value of a key in Redis
-    :param key: Name of the Key
-    :param value: Value to be set
-    :param db: if None the Redis db object is constructed again
-    """
-
-    if db is None:
-        db = app.redis
-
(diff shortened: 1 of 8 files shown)


actionless:
i am not sure what to do with the tests

petrjasek:
imo it would good to have some helper methods for those sequences so you don't duplicate this `find_and_modify` query every time we use this, can you maybe introduce sequences module with a few methods to manipulate it?
then you could remove this `set_key` method and its tests

actionless:
@petrjasek, like that?
```
def get_next_sequence_number(resource, id, min_number, max_number, field_name='sequence_number'):
```

actionless:
@petrjasek done

actionless:
so about the tests, should i do them for this new sequences helper?

petrjasek:
right there should be a test for this new helper.
otherwise looks good.

superdesk/celery_app.py
```diff
@@ -187,6 +187,7 @@ def _update_subtask_progress(task_id, current=None, total=None, done=None):
 
 def set_key(key, value=0, db=None):
     """
+    @TODO: deprecate this?
```
ioanpocol: Yes, please delete set_key method.

ioanpocol:
:+1:  (except last 2 comments)

actionless:
@ioanpocol @petrjasek done

ioanpocol:
:+1:

PR #99, opened 2015-12-08
fix(dpa and rss ingest)

marwoodandrew:
For RSS the parser and the service are tightly bound, so has no separate parser.
For DPA can parse always returned false


mugurrus:
:+1:

sivakuna-aap:
:+1:

PR #100, opened 2015-12-08
notification: create a new event loop for send_message

vied12:
I needed to do that in order to prevent this error:
```
17:19:42 rest.1 | Traceback (most recent call last):
17:19:42 rest.1 |   File "env/src/superdesk-core/superdesk/notification.py", line 83, in push_notification
17:19:42 rest.1 |     _notify(event=name, extra=kwargs)
17:19:42 rest.1 |   File "env/src/superdesk-core/superdesk/notification.py", line 61, in _notify
17:19:42 rest.1 |     loop = asyncio.get_event_loop()
17:19:42 rest.1 |   File "/usr/lib/python3.4/asyncio/events.py", line 576, in get_event_loop
17:19:42 rest.1 |     return get_event_loop_policy().get_event_loop()
17:19:42 rest.1 |   File "/usr/lib/python3.4/asyncio/events.py", line 522, in get_event_loop
17:19:42 rest.1 |     % threading.current_thread().name)
17:19:42 rest.1 | RuntimeError: There is no current event loop in thread 'Thread-2'.
```

### DIFF
--- a/superdesk/notification.py
+++ b/superdesk/notification.py
@@ -58,7 +58,7 @@ def _notify(**kwargs):
     def send_message():
         yield from app.notification_client.send(message)
 
-    loop = asyncio.get_event_loop()
+    loop = asyncio.new_event_loop()
     loop.run_until_complete(send_message())
 
 
--- a/superdesk/notification.py
+++ b/superdesk/notification.py
@@ -40,7 +40,11 @@ def init_app(app):
     host = app.config['WS_HOST']
     port = app.config['WS_PORT']
     try:
-        loop = asyncio.get_event_loop()
+        try:
+            loop = asyncio.get_event_loop()
+        except:
+            loop = asyncio.new_event_loop()
+      
         app.notification_client = loop.run_until_complete(websockets.connect('ws://%s:%s/server' % (host, port)))
         logger.info('websocket connected on=%s:%s' % app.notification_client.local_address)
     except (RuntimeError, OSError):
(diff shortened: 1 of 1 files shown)


mugurrus:
:+1:

vied12:
Perhaps it would be possible to use `set_event_loop(loop)` when the thread is created instead of creating an event loop in `_notify()`.
Maybe someone who knows better the life cycle of notifications can answer :)

ioanpocol:
@petrjasek  Any feedback on this?

petrjasek:
it's created by gunicorn, not sure we can handle that. maybe via some app callback - before first request - set `app.loop` and use that in notifications?

superdesk/notification.py
```diff
@@ -58,7 +58,7 @@ def _notify(**kwargs):
     def send_message():
         yield from app.notification_client.send(message)
 
-    loop = asyncio.get_event_loop()
+    loop = asyncio.new_event_loop()
```
ioanpocol: Without this change at least locally the notification is not working at all.
With this change the notification don't work if app.py is started after ws.py.

I added here and to https://github.com/superdesk/superdesk-core/blob/fix/notification-loop-multi-thread/superdesk/notification.py#L43 the following code:
```
        try:
            loop = asyncio.get_event_loop()
        except:
            loop = asyncio.new_event_loop()
```

and at least locally it is working well.

vied12:
Cool ! So you think you will update this PR ?

ioanpocol:
Please update you :)

vied12:
I'm not sure to understand weel you comment and what to do with current modifications.
It's better if you do it yourself :relaxed:
(You have the quick edit mode in github if you are in a hurry)

petrjasek:
:+1:

vied12:
:ok_hand:

PR #102, opened 2015-12-09
refactor(sequences): make it a service; add method for key-based sequences

actionless:


superdesk/sequences.py
```diff
@@ -1,37 +1,88 @@
...
-            upsert=False
-        )
+def init_app(app):
+    endpoint_name = 'sequences'
+    service = SequencesService(endpoint_name, backend=superdesk.get_backend())
+    SequencesResource(endpoint_name, app=app, service=service)
+
+
+class SequencesResource(Resource):
+    schema = {
```
actionless: will it create elastic indexes for it by default?
i think it would be better to avoid it

ioanpocol: No, will create elastic collection only if you set 'search_backend': 'elastic' in datasource

ioanpocol:
:+1:

superdesk/sequences.py
actionless: btw, should i set any privileges here?

by default it will be writeable for all or only for authenticated users?

superdesk/sequences.py
```diff
@@ -1,37 +1,90 @@
...
+            'iunique': True
+        },
+        'sequence_number': {
+            'type': 'number',
+            'default': 0,
+        },
+    }
+    etag_ignore_fields = ['sequence_number', 'name']
+    resource_methods = []
+    item_methods = []
```
actionless: is that enough to make it accessible only internally?

ioanpocol: You have to set internal_resource = True like it is set forArchiveVersionsResource.

actionless: why ArchiveVersionsResource have 
```
class ArchiveVersionsResource(Resource):
    resource_methods = []
    internal_resource = True
```
?

actionless:
@ioanpocol  comment got hidden after i updated the code:

>why ArchiveVersionsResource have both
```
    resource_methods = []
    internal_resource = True
```
?

superdesk/sequences.py
```diff
@@ -1,37 +1,91 @@
...
+        :param max_seq_number: default $MAX_VALUE_OF_PUBLISH_SEQUENCE, maximal possible value
+        :param min_seq_number: default 0, minimal possible value
+        :param sequence_field: default 'sequence_number', name of field to store sequence number in the target item
+        :returns: sequence_number
+        """
+        if not max_seq_number:
+            max_seq_number = app.config['MAX_VALUE_OF_PUBLISH_SEQUENCE']
+
+        target_resource = get_resource_service(resource_name)
+        target_item = target_resource.find_one(**{
```
mugurrus: You don't need to create a dict here, the call should look like:
`find_one(req=None, query_field=query_value)`

superdesk/sequences.py
```diff
@@ -1,37 +1,91 @@
...
+        },
+    }
+    etag_ignore_fields = ['sequence_number', 'name']
+    internal_resource = True
+
+
+class SequencesService(BaseService):
 
-    return sequence_number
+    def get_next_sequence_number_for_item(
```
mugurrus: For consistency all sequences should be stored in the sequences collection, there's no need for customizing the collection where the sequence should be stored. This is also error prone. This would simplify the API to:
`def get_next_sequence_number(self, key_name, max_seq_no=None, min_seq_no=0)`

superdesk/sequences.py
actionless: no, `query_field!='query_field'`

superdesk/sequences.py
actionless: that will require cleaning up the sequences on item delete, so not simplifying api at all, but i will preset the changes as a separate commit to e possible to rollback

actionless:
also this one got hidden but comment but the question is still actual: 
https://github.com/superdesk/superdesk-core/pull/102#discussion_r47202216

mugurrus:
:+1:

superdesk/sequences.py
```diff
@@ -1,37 +1,65 @@
...
-    return sequence_number
+def init_app(app):
+    endpoint_name = 'sequences'
+    service = SequencesService(endpoint_name, backend=superdesk.get_backend())
+    SequencesResource(endpoint_name, app=app, service=service)
+
+
+class SequencesResource(Resource):
+    schema = {
+        'key': {
```
petrjasek: it could use `_id` field which has all the characteristics and has an index on top..

PR #104, opened 2015-12-10
(fix)(elasticsearch) use analyzer instead index_analyzer in item schema

ahilles107:
Fix `analyzer on field [phrase] must be set when search_analyzer is set` error on schema build


ioanpocol:
:+1:

actionless:
:+1:

PR #106, opened 2015-12-11
Fix for sequence numbers and legal prefix

akintolga:


sivakuna-aap:
:+1:

mdhaman:
:+1:

superdesk/io/ingest_provider_model.py
```diff
@@ -126,8 +126,7 @@ def __init__(self, endpoint_name, app, service, endpoint_schema=None):
             },
             'sequence_number': {
```
mugurrus: Should be removed, not needed anymore.

PR #110, opened 2015-12-17
account creation email message updated

vied12:
to let able to use more variable, like `user` and the instance url in custom templates, like we need in liveblog


petrjasek:
:+1:

ioanpocol:
:+1:

PR #120, opened 2016-01-04
[sd-3808] Ruleset not working

marwoodandrew:
if a rule_set was passed to apply_rule_set then it would never be applied.


sivakuna-aap:
:+1:

mugurrus:
:+1:

PR #122, opened 2016-01-06
(fix) [SD-3866] there is no prompting re-login notification received for the newly added user to a desk(s)

sivakuna-aap:
Required by https://github.com/superdesk/superdesk/pull/1592


mugurrus:
:+1:

petrjasek:
:+1:

PR #124, opened 2016-01-07
Modified logging so that is controlled via .yml file in the superdesk repo

mdhaman:
This is dependent on [PR 1600](https://github.com/superdesk/superdesk/pull/1600)


mdhaman:
@petrjasek , @ioanpocol , @mugurrus can please a look at this change. Maybe it might be affecting other projects.

sivakuna-aap:
:+1:

petrjasek:
:+1:

PR #127, opened 2016-01-12
[SD-2825] Fixed notification for granting privileges to a role.

mdhaman:


superdesk/roles/roles.py
```diff
@@ -93,11 +93,12 @@ def __send_notification(self, updates, role):
...
-                             'role {{role}} is granted new privileges: Please re-login.',
-                             self.datasource,
-                             notify=notified_users,
-                             role=role.get('name'))
+            if len(added) > 0 or (0, 1) in modified.values():
+                activity = add_activity(ACTIVITY_UPDATE,
+                                        'role {{role}} is granted new privileges: Please re-login.',
+                                        self.datasource,
+                                        notify=notified_users,
+                                        role=role.get('name'))
```
sivakuna-aap: Do we need 2 push notifications? If not can we do the same thing what we are doing in desks.py?

sivakuna-aap:
:+1: except the minor one

superdesk/roles/roles.py
mdhaman: not required.

marwoodandrew:
:+1:

PR #132, opened 2016-01-15
fix(WENN parser) handle ftp retrieve errors and markup body

marwoodandrew:
The FTP retrbinary sometimes throws an exception, since the file has been opened, an empty file is created and there is never another attempt to retrieve it.


ioanpocol:
:+1:

mugurrus:
:+1:

PR #136, opened 2016-01-19
[SD-3817] Make the the locators lookup more comprehensive

marwoodandrew:
Generated a list of cities with populations > 50,000, It comes to more than 8,000 cities.

Might be a bit over the top?


petrjasek:
:+1:
seems like it's not fetched by client, should be fine imo

akintolga:
:+1:

PR #138, opened 2016-01-21
refactor(apps): move apps into core

petrjasek:
so idea is to put more or less everything into core repo (and we might rename this to -server at some point). then in main repo you import this with pip and with settings you enable/disable things


ioanpocol:
:+1: 
Are we moving also the frontend part on superdesk-core?

petrjasek:
@ioanpocol into it's own repo, but same thing

mugurrus:
Why put AAP code in the central repo?

petrjasek:
@mugurrus can you be more specific?

mugurrus:
e.g. https://github.com/superdesk/superdesk/tree/master/server/apps/publish/formatters

mdhaman:
:+1: All Good. Formatters can be moved later on.

akintolga:
:+1: Agree.

PR #143, opened 2016-02-03
fix(anpa): fix formatting  for empty dateline.text

petrjasek:
it was checking if text was set but it could be set to null

SD-3469


akintolga:
:+1:

mdhaman:
:+1:

PR #146, opened 2016-02-03
[SD-3977] Kill articles that have been imported into archived from outside superdesk

marwoodandrew:


superdesk/tests/steps.py
```diff
@@ -714,9 +729,6 @@ def step_impl_when_upload_with_crop(context):
 @when('upload a file "{file_name}" to "{destination}" with "{guid}"')
 def step_impl_when_upload_image_with_guid(context, file_name, destination, guid):
     upload_file(context, destination, file_name, 'media', {'guid': guid})
-    if destination == 'archive':
```
mdhaman: these lines might be required. check the behave test.

apps/archive/common.py
```diff
@@ -428,6 +428,10 @@ def item_schema(extra=None):
                 'marked_for_legal': {
                     'type': 'boolean',
                     'default': False
+                },
+                'marked_archived_only': {
```
sivakuna-aap: Since this attribute exists only in `archived` repo, it should be moved to `ArchivedResource` instead of `ArchiveResource`

marwoodandrew: Probably should but can't see a nice way, since we are only needing to add a flag to the schema within a already defined field. I'd have to pass the definition of all flags in extra_fields.

Can you see a better way?

sivakuna-aap: Got your point, grouping all the flags together makes sense. :+1:

mdhaman:
:+1:

petrjasek:
:+1:

PR #147, opened 2016-02-04
[SD-3971] If source is defined at the desk level then apply the source for manually created articles on the desk.

mdhaman:
If `source` is defined on the desk then manually created article use the desk `source` setting.
If `source` is `not` defined on the desk then manually created article use the `DEFAULT_SOURCE_VALUE_FOR_MANUAL_ARTICLES` setting.  

`The same logic applies for items created via duplicate action.`


petrjasek:
:+1:

akintolga:
:+1:

PR #151, opened 2016-02-04
SD-3067 Publish stories asynchronously

mugurrus:


apps/publish/content/common.py
```diff
@@ -167,56 +114,57 @@ def update(self, id, updates, original):
...
+
+                self._update_archive(original, updated, should_insert_into_versions=auto_publish)
+                self.update_published_collection(published_item_id=original[config.ID_FIELD], updated=updated)
+                """
+                This sequence is for test purposes, it will be removed once the feature is finished.
+                """
+#                 item = get_resource_service('published').find_one(req=None, item_id=original[config.ID_FIELD])
+#                 enqueue_item(item)
+
+            self._process_publish_updates(original, updates)
```
mdhaman: imo, call to `self._process_publish_updates` should be moved to before https://github.com/mugurrus/superdesk-core/blob/async_publishing/apps/publish/content/common.py#L159.

apps/publish/content/common.py
```diff
@@ -391,52 +250,17 @@ def _get_digital_id_for_package_item(self, package_item):
...
-        if not preserve_state:
-            self.set_state(original, updates)
-
-        updates.setdefault(config.LAST_UPDATED, last_updated)
+    def _process_publish_updates(self, original, updates):
+        """ Common updates for published items """
+        desk = None
+        if original.get('task', {}).get('desk'):
+            desk = get_resource_service('desks').find_one(req=None, _id=original['task']['desk'])
+        if not original.get('ingest_provider'):
```
mdhaman: Agree that it is part of the old code. 
But not sure we should change the source of the manually created articles on publishing action. Have to chat with POs.

apps/publish/content/common.py
```diff
@@ -446,35 +270,13 @@ def _set_item_expiry(self, updates, original):
         stage_id = original.get('task', {}).get('stage')
         offset = updates.get(EMBARGO, original.get(EMBARGO)) or \
             updates.get('publish_schedule', original.get('publish_schedule'))
-
         updates['expiry'] = get_expiry(desk_id, stage_id, offset=offset)
 
-    def _update_archive(self, original, updates, versioned_doc=None, should_insert_into_versions=True):
+    def _is_take_item(self, item):
```
mdhaman: Use the `get_take_package_id` function of the `TakesPackageService`.

akintolga:
:+1:  Looks good as the initial 'async' implementation.

ioanpocol:
:+1:

PR #152, opened 2016-02-05
Feat macros interactive

petrjasek:
let macros return diff which can be used by client to display all replacements with find&replace like interface


akintolga:
:+1:

ioanpocol:
:+1:

PR #153, opened 2016-02-05
fix(apps: auth: db: commands): make --admin boolean

actionless:
resolves https://github.com/superdesk/superdesk/pull/1729

(argparse should do converting job for us)

### DIFF
--- a/apps/auth/db/commands.py
+++ b/apps/auth/db/commands.py
@@ -10,7 +10,7 @@
 
 import logging
 from base64 import b64encode
-from flask import current_app as app, json
+from flask import current_app as app
 import superdesk
 from superdesk.utils import get_hash, is_hashed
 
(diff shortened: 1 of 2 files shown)


petrjasek:
:+1:

mdhaman:
:+1:

apps/auth/db/commands.py
```diff
@@ -27,13 +27,12 @@ class CreateUserCommand(superdesk.Command):
         superdesk.Option('--username', '-u', dest='username', required=True),
         superdesk.Option('--password', '-p', dest='password', required=True),
         superdesk.Option('--email', '-e', dest='email', required=True),
-        superdesk.Option('--admin', '-a', dest='admin', required=False),
+        superdesk.Option('--admin', '-a', dest='admin', required=False, default=False),
```
petrjasek: ideally it should have also `action='store_true'` so that it can be used without any value, like `--admin`

actionless: isn't it behave like that for bools by default?

petrjasek: don't think so

actionless: https://github.com/smurfix/flask-script/blame/master/docs/index.rst#L249-L269

petrjasek: that probably works if you don't use `Option` and it does config automatically
btw `dest` should be `is_admin`

actionless: under the hood it creates `Option` instances for them anyway

renamed it back to `admin` to avoid any confusion

petrjasek: yep it does and it does some detection, but when you set it explicitly it doesn't do it.
I've tested it - if you run it with just `--admin` you get `manage.py users:create: error: argument --admin/-a: expected one argument`

actionless: weird, after the sleep i will try to figure that out

actionless:
so adding `, action='store_true'` as Petr suggested did help, i tested it works as expected now

petrjasek:
it will require some updates in scripts because not it will fail with `--admin=true`

actionless:
prepared: https://github.com/superdesk/superdesk/pull/1742

petrjasek:
:+1:

actionless:
(will get back to it after the server migration which will be on the weekend)

PR #157, opened 2016-02-09
fix(sentry): make it work with new sentry

petrjasek:
in case sentry is not configured it throws an error in newer versions.
this adds a proxy which will only make an instance if configured.


mdhaman:
:+1:

mugurrus:
:+1:

PR #159, opened 2016-02-09
fix(expiry): fix expiry of items created from template

petrjasek:
those items were missing current version field which caused
an exception in expiry task and thus preventing other items
from being removed.

SD-3912


mdhaman:
:+1:

marwoodandrew:
:+1:

mugurrus:
:+1:

PR #160, opened 2016-02-09
fix(workqueue): When an item is opened for correction it will show up on the bottom panel (SD-3556)

ioanpocol:


mugurrus:
:+1:

petrjasek:
looks ok, but can you nelu explain a bit the solution? just wondering how it fixes the issue

sivakuna-aap:
I think this solution might be a performance hit as archive repository in Mongo can have million(s) of docs and since workqueue is visible always, every operation on the article hits the workqueue endpoint.

Is there a corresponding PR in superdesk-client-core?

ioanpocol:
Changed to use elastic instead of mongodb

PR #161, opened 2016-02-10
fix(publish): trigger enqueue publish after publishing

petrjasek:
@mugurrus seems like the issue publishing e2e was missing `enqueue_published` call.
I've added it to publishing handler - not sure if it could be an issue - this way it reacts faster and e2e tests work without any changes. otherwise it would have to wait for the queue. what do you think?


akintolga:
:+1:

mdhaman:
:+1:

PR #165, opened 2016-02-15
[SD-3936] Items missing from Reuters ingest

marwoodandrew:
Changes the Reuters ingest to :- 

- keep the pollToken rather than relying on times to identify new content. 

- Request the content id's rather than guids (guids exclude the version component)


mugurrus:
:+1: (I highly suggest writing a regression test)

mdhaman:
:+1:

ioanpocol:
:+1:

PR #173, opened 2016-02-18
SDPA-152 All uploaded, cropped or manipulated images should be referenced by an S3 HTTPS url

mugurrus:


vied12:
:+1:

ioanpocol:
We can add a few behave tests that will run only if the amazon related settings are set.
We have to store original urls to amazon and create the proxy url only on fetch.
Right now the files on amazon are public. Do we intend to implement any security scheme on them?

ioanpocol:
:+1:  (except comments)

mugurrus:
Behave tests and generate URL on fetch in a separate PR

PR #175, opened 2016-02-22
[SD-4069] Add on stage macro on move and a publishing macro

marwoodandrew:
The PO wants to be able to publish content by just moving it to a stage.

The reason for this is that they can transition users who create content to Superdesk and have them use the normal workflow (i.e creating content and moving it to another stage). WITHOUT  having them publish using the normal publishing mechanism.


mugurrus:
:+1:

mdhaman:
:+1:

PR #176, opened 2016-02-23
[SD-4080] Modifying the expiry process to ensure that item are in legal before they are expired.

mdhaman:
This change are to ensure that if the legal archive is down then items are not deleted from the production archive after expiry.

- [x] Add a flag `moved_to_legal` to published item
- [x] Add a flag `moved_to_legal` to publish queue item
- [x] Modify legal archive import to set the appropriate flag when items are moved
- [x] Abort the expiry process if items are not moved.
- [x] Command to move `stray item` into legal archive.
- [x] Unit and Behave tests


akintolga:
:+1:

mdhaman:
@superdesk/developers - Need Review

apps/legal_archive/commands.py
```diff
@@ -25,95 +26,124 @@
...
-            version_doc.pop(config.ETAG, None)
-
-        if versions_to_insert:
-            legal_archive_versions_service.post(versions_to_insert)
-            logger.info('Inserted de-normalized version history for article {}'.format(log_msg))
-
-        logger.info('Upsert completed for article ' + log_msg)
+        try:
+            # used for unit of ImportLegalArchiveCommand
+            if app.config.get('Import_LegalArchive_Command_Testing', False):
```
petrjasek: this could be avoided with MagicMock

apps/archive/commands.py
```diff
@@ -80,19 +81,26 @@ def _remove_expired_items(self, expiry_datetime, log_msg):
...
+                logger.info('{} Removing item. {}'.format(self.log_msg, expiry_msg))
+                logger.info('{} Items to be removed. {}'.format(self.log_msg, processed_items))
                 items_to_be_archived = items_to_be_archived | processed_items
 
+        items_to_expire = items_to_be_archived | set([item.get(config.ID_FIELD) for item in spiked_killed_items
+                                                      if item.get(ITEM_STATE) == CONTENT_STATE.KILLED])
+
+        if not self.check_if_items_imported_to_legal_archive(list(items_to_expire)):
+            logger.exception('{} CANNOT EXPIRE ITEMS AS ITEMS ARE NOT MOVE TO LEGAL ARCHIVE.'.format(self.log_msg))
+            return
```
petrjasek: just not sure it should stop here or remove items that are in legal archive. but we can probably merge this and see how it works..

petrjasek:
:+1:

apps/legal_archive/commands.py
mdhaman: will look into it.

apps/archive/commands.py
mdhaman: Agree on that it will no remove items if they are in legal archive. I didn't want to complicate things, due to package/broadcast dependencies . Hence, I am checking for all items for now.

petrjasek:
:+1:

PR #177, opened 2016-02-23
Merge pa

petrjasek:
same thing like on client. adding alchemy support, some changes regarding image cropping, and there is one breaking change - renaming description to description_text in schema. we should probably figure out how to migrate data before merging this to master


akintolga:
:+1:

PR #178, opened 2016-02-24
[SD-4066] fix category mapping fo AP feeds

marwoodandrew:


superdesk/io/feed_parsers/ap_anpa.py
```diff
@@ -0,0 +1,135 @@
...
+                category['qcode'] = 's'
+
+    def map_sluglines_to_subjects(self, item):
+        """
+        The first few characters of the slugline may match AP supplimetal categories
+        this is used to set the subject code.
+        :param item:
+        :return:
+        """
+        if len(item['slugline']) > 4:
```
mdhaman: use `item.get('slugline','')` so that it works in case of missing slugline

mdhaman:
:+1:  except minor comment.

akintolga:
:+1:

PR #185, opened 2016-02-26
fix(service): fix delete by multiple ids

petrjasek:
it was using mongo query when doing elastic find_one,
which was failing on master instance.

also as long as you removed more than 1 item it wasn't calling
on_update/on_updated methods for removed items.

fixes SD-4005 I hope

### DIFF
--- a/apps/archive/archive_test.py
+++ b/apps/archive/archive_test.py
@@ -11,6 +11,7 @@
 
 from bson import ObjectId
 
+from unittest.mock import MagicMock
 from superdesk import get_resource_service
 from test_factory import SuperdeskTestCase
 from eve.utils import date_to_str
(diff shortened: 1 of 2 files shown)


petrjasek:
@mdhaman at this point it's failing on master every time expiry is running, we will see after that, couldn't reproduce the error locally

mugurrus:
:+1:

akintolga:
:+1:

mdhaman:
:+1: :relieved: 
@petrjasek yes it is causing error at line. You can only reproduce the error if you have more than 50 (approx) spiked or killed items. The error seems to be related to the parsing the URL. For smaller number of records, elastic returns 404 (not found error is swallowed by eve-elastic) and rest of the code goes an deletes the records from archive collection.

PR #186, opened 2016-02-26
feat(auth: db: commands): add --no-update option to users:create script

actionless:
https://dev.sourcefabric.org/browse/LBSD-842

### DIFF
--- a/apps/auth/db/commands.py
+++ b/apps/auth/db/commands.py
@@ -28,9 +28,10 @@ class CreateUserCommand(superdesk.Command):
         superdesk.Option('--password', '-p', dest='password', required=True),
         superdesk.Option('--email', '-e', dest='email', required=True),
         superdesk.Option('--admin', '-a', dest='admin', required=False, action='store_true'),
+        superdesk.Option('--no-update', '-n', dest='no_update', required=False, action='store_true'),
     )
 
-    def run(self, username, password, email, admin):
+    def run(self, username, password, email, admin=False, no_update=False):
 
         # force type conversion to boolean
         user_type = 'administrator' if admin else 'user'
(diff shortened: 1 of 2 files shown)


apps/auth/db/commands.py
```diff
@@ -52,6 +53,9 @@ def run(self, username, password, email, admin):
             user = superdesk.get_resource_service('users').find_one(username=userdata.get('username'), req=None)
 
             if user:
+                if no_update:
+                    logger.info('user already exists %s' % (userdata))
+                    return userdata
                 logger.info('updating user %s' % (userdata))
                 superdesk.get_resource_service('users').patch(user.get('_id'), userdata)
```
ahilles107: Personally i would even remove updating part from that command as it's `CreateUserCommand` not `CreateOrUpdateUserCommand`. It would also fix issue from ticket.

mugurrus:
:+1:

ahilles107:
@actionless it will be not better to remove call (on upgrade) for create user command instead adding that --no-update option there?

actionless:
it could lead to unnecessary duplication of some code parts or creation of some additional helper functions so i would like to leave it as is to avoid both

ahilles107:
It ***could***, duplication of ***some*** code parts, creation of ***some*** additional helper. 

I know that i'm not core developer in Superdesk team but You can still write to me with details. I believe in clear code responsibilities, and this PR only obscures code and command. Much better would be just creating separate console command for updating user and not doing this here. We have `users` service for that task, so only think what can be duplicated will be command input validation. 

Adding parameters like that don't make project easier for non core developers.

actionless:
ok, i see you mean creating another command, not another function

so yes, it's possible to create two different commands which will be calling the same function with some other argument to change that create/update behavior

but i still don't getting why that create/update behavior is so important to create a separate command just for that

would like to see some feedback from @petrjasek

mugurrus:
There is no update users command. We only need create command. This is an option that is specific for a case.

ahilles107:
We don't do software for us but for users/other developers and adding ```option that is specific for a case``` without rethinking how it could be done if it would be designed properly shouldn't have place.

petrjasek:
imo it shouldn't try to update at all if it's called create, so if user exists it can just log an info message and quit

mugurrus:
Agree with Petr

actionless:
ok, i've submitted alternative implementation ^

PR #198, opened 2016-03-01
fix(settings): don't set SERVER_NAME

actionless:
it's causing issues when attempting to set SUPERDESK_URL with the port other than 80


actionless:
it's not needed for flask itself and i haven't found it being used by any other superdesk parts

however if it's still required somewhere i could replace that `.endswith(':80'):` with something more universal

ioanpocol:
:+1:

actionless:
oops, it broken flask's `url_for` will do the alternate fix then

actionless:
fixed on the upper level, https://github.com/mitsuhiko/werkzeug/pull/872

idk how soon it will be merged, mb we have to use werkzeug fork then

PR #205, opened 2016-03-04
feat(correction): Store the current correction sequence (SDNTB-31)

ioanpocol:


mdhaman:
:+1:

apps/publish/content/correct.py
```diff
@@ -40,6 +40,7 @@ def on_update(self, updates, original):
         super().on_update(updates, original)
         updates[ITEM_OPERATION] = ITEM_CORRECT
         updates['versioncreated'] = utcnow()
+        updates['correction_sequence'] = original.get('correction_sequence', 0) + 2
```
akintolga: Maybe this should be `original.get('correction_sequence', 2) + 1`

PR #206, opened 2016-03-04
[SD-4134] Change genre vocabulary to use qcode rather than value

marwoodandrew:
As a an initial part of making the vocabularies constant, change the use of value to qcode in genre.


mdhaman:
:+1:

akintolga:
:+1:

ahilles107:
:+1:

petrjasek:
just wondering if we should call it qcode, because without having some id of the vocabulary it isn't a qcode really, you need to know where it comes from. it's more like a code in ninjs
> The code for the subject in a scheme (= controlled vocabulary) which is identified by the scheme property

marwoodandrew:
I think that the aim is to make the structure of the vocabulary entries constant to simplify/unify the handling of them.

ioanpocol:
please rebase

petrjasek:
:+1: just wondering what should be the property name, agree with the changes

mdhaman:
do we need to store the scheme definition with `qcode|code` in the mongo (anpa_category:a) or it is a concern for output formatters?

petrjasek:
I think in vocabulary it should be `code` and scheme is vocabulary id, but on an item we should have either `code` + `scheme` or `qcode` which combines both.

akintolga:
Merging as it is needed for e2e tests...

petrjasek:
no worries

PR #216, opened 2016-03-09
fix(celery-app): disconnect redis all connections after updating task…

actionless:
… progress

### DIFF
--- a/superdesk/celery_app.py
+++ b/superdesk/celery_app.py
@@ -168,18 +168,20 @@ def update_key(key, flag=False, db=None):
 
 def _update_subtask_progress(task_id, current=None, total=None, done=None):
     redis_db = redis.from_url(celery.conf['CELERY_RESULT_BACKEND'])
+    try:
+        current_key = 'current_%s' % task_id
+        total_key = 'total_%s' % task_id
+        done_key = 'done_%s' % task_id
 
-    current_key = 'current_%s' % task_id
-    total_key = 'total_%s' % task_id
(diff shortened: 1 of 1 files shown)


actionless:
@mugurrus seems like you haven't updated the unittest test for amazon upload:
https://travis-ci.org/superdesk/superdesk-core/builds/114672057

vied12:
:+1:

petrjasek:
:+1:

superdesk/celery_app.py
```diff
@@ -168,18 +168,20 @@ def update_key(key, flag=False, db=None):
 
 def _update_subtask_progress(task_id, current=None, total=None, done=None):
     redis_db = redis.from_url(celery.conf['CELERY_RESULT_BACKEND'])
```
naomiaro: should this be in the try/catch as well?

actionless: it won't change the behaviour so it's more matter of codestyle

petrjasek: it can't be in try/finally because then it would try to close it even if it fails on that line

naomiaro: Yeah but it could b=maybe throw some other exception if it can't open a connection properly. Can check if it's set when trying to close it in finally.

petrjasek: yep if it fails then it stops the program and that's probably good thing to do

actionless: if it fails to connect when it's nothing to disconnect from

PR #220, opened 2016-03-09
fix(date): Change the date format to HH:MM:SS in order to be consiste…

ioanpocol:
…nt over the application

### DIFF
--- a/features/steps/template_steps.py
+++ b/features/steps/template_steps.py
@@ -21,5 +21,4 @@ def then_next_run_is_on_monday(context, time):
     next_run = parse_date(data.get('next_run'))
     assert isinstance(next_run, datetime)
     assert next_run.weekday() == 0
-    assert next_run.strftime('%H%M') == time
-    assert next_run.strftime('%S') == '00', 'there should be no seconds, but it was %s' % (next_run.strftime('%S'), )
+    assert next_run.strftime('%H:%M:%S') == time
--- a/superdesk/utc.py
+++ b/superdesk/utc.py
@@ -73,17 +73,17 @@ def utc_to_local(local_tz_name, utc_datetime):
         return local_tz.normalize(local_dt)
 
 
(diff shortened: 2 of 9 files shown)


petrjasek:
:+1:

features/steps/template_steps.py
```diff
@@ -21,5 +21,5 @@ def then_next_run_is_on_monday(context, time):
     next_run = parse_date(data.get('next_run'))
     assert isinstance(next_run, datetime)
     assert next_run.weekday() == 0
-    assert next_run.strftime('%H%M') == time
+    assert next_run.strftime('%H:%M:%S') == time
     assert next_run.strftime('%S') == '00', 'there should be no seconds, but it was %s' % (next_run.strftime('%S'), )
```
naomiaro: prob need to get rid of this assert now for no seconds.

superdesk/utc.py
```diff
@@ -76,14 +76,14 @@ def utc_to_local(local_tz_name, utc_datetime):
...
     :param datetime current_datetime
     :param string timestr
     :param int second
     """
     if timestr is None:
-        timestr = '0000'
-    time = datetime.datetime.strptime(timestr, '%H%M')
+        timestr = '00:00:00'
+    time = datetime.datetime.strptime(timestr, '%H:%M:%S')
     return current_datetime.replace(hour=time.hour, minute=time.minute, second=second)
```
naomiaro: should use time.second here as well, now that the time can have seconds.

superdesk/utc.py
```diff
@@ -76,14 +76,14 @@ def utc_to_local(local_tz_name, utc_datetime):
 def set_time(current_datetime, timestr, second=0):
```
naomiaro: don't need this second=0

naomiaro:
There's a few time things I've posted on a couple lines.

PR #221, opened 2016-03-10
[SD-3097] - Extra carriage returns cleared when a dictionary is uploaded

akintolga:


petrjasek:
just thinking how is this related to sd-3097?

akintolga:
When a custom dictionary is loaded, it may have `\r` as well as `\n`. if not stripped those words are saved with `\r` at the end (esp if the file is created from a windows based editor). And no words on the editor matches them.

petrjasek:
:+1: makes sense wasn't sure about the case sensitivity

mugurrus:
:+1:

mdhaman:
:+1:

PR #222, opened 2016-03-10
(fix) Fix e2e test for highlight

mdhaman:
@ioanpocol , @petrjasek , @mugurrus  fix for the highlight test.

### DIFF
--- a/features/highlights.feature
+++ b/features/highlights.feature
@@ -336,3 +336,70 @@ Feature: Highlights
         Given highlights
         When we create highlights package
         Then we get new package with items
+
+
+    @auth
+    Scenario: Generate text item from highlights using filter
+        Given "desks"
+        """
+        [{"name": "desk1"}]
+        """
+        Given "archive"
+        """
+        [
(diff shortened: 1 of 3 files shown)


ioanpocol:
:+1:

petrjasek:
just wondering - would it be possible to handle undefined param in that filter? not to complicate the template users might need to deal with?

akintolga:
Yes, I think it is possible, I'll make a change for that today.

mdhaman:
@petrjasek done the change.

petrjasek:
:+1: nice

features/highlights.feature
```diff
@@ -336,3 +336,70 @@ Feature: Highlights
...
+        When we get "/archive"
+        Then we get list with 2 items
+
+        When we post to "generate_highlights"
+        """
+        {"package": "package", "preview": true}
+        """
+        Then we get new resource
+        """
+        {"type": "text"}
```
ahilles107: missing new line at end of file

PR #225, opened 2016-03-11
fix(email): stop sending emails in test

petrjasek:
it doesn't work anyhow


naomiaro:
:+1:

ioanpocol:
:+1:

petrjasek:
seems like we use that varible when doing some tests and mocking email.

PR #227, opened 2016-03-14
SD-4208 Update Amazon S3 proxy URL to include version

mugurrus:


akintolga:
:+1:

actionless:
mb only better smth like `AMAZON_PROXY_VERSION`

superdesk/storage/amazon/amazon_media_storage.py
```diff
@@ -61,7 +61,8 @@ def url_for_media_default(app, media_id, content_type=None):
     extension = str(_guess_extension(content_type)) if content_type else ''
     url = '%s.%s/%s%s' % (app.config['AMAZON_CONTAINER_NAME'], endpoint, media_id, extension)
```
actionless: extension won't work without proxy so you have to add it under `if app.config.get('AMAZON_PROXY_SERVER'):`

actionless: or that extension should be added when uploading, so it won't be added by proxy

PR #228, opened 2016-03-14
fix(errors): send error emails only once every 24 hours

petrjasek:
by default, can be set via setting `EMAIL_NOTIFICATION_RESEND` in hours


akintolga:
:+1:

mdhaman:
:+1:

PR #231, opened 2016-03-15
feat(sign-off): Map signoff and add edit signoff for item(SDNTB-24, S…

ioanpocol:
…D-NTB-25)

Those changes are available only if sign_off mapping is set on
settings.py and index.html
When an user is added/edited, the sign-off is not shown and is set to
mapped value
On authoring the user is able to set a new value for sign-off by
selecting only one user from user list

### DIFF
--- a/apps/archive/common.py
+++ b/apps/archive/common.py
@@ -186,6 +186,7 @@ def set_sign_off(updates, original=None, repo_type=ARCHIVE, user=None):
     Set sign_off on updates object. Rules:
         1. updates['sign_off'] = original['sign_off'] + sign_off of the user performing operation.
         2. If the last modified user and the user performing operation are same then sign_off shouldn't change
+        3. If sign_off is received on updates, this value will be preserved
     """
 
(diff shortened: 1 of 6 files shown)


apps/archive/common.py
```diff
@@ -186,6 +186,7 @@ def set_sign_off(updates, original=None, repo_type=ARCHIVE, user=None):
     Set sign_off on updates object. Rules:
         1. updates['sign_off'] = original['sign_off'] + sign_off of the user performing operation.
         2. If the last modified user and the user performing operation are same then sign_off shouldn't change
+        3. If sign_off is received on updates, this value will be preserved
```
akintolga: A new behave test in archive.feature would be good for this behavior.

akintolga:
:+1:

mdhaman:
:+1:

apps/archive/common.py
ioanpocol: Done

PR #234, opened 2016-03-15
'alt_text' added into metadata SDPA-332

vied12:


superdesk/metadata/item.py
```diff
@@ -298,6 +298,11 @@
     'associations': {
         'type': 'dict',
     },
+    'alt_text': {
+        'type': 'text',
```
akintolga: Should this type `string`?

superdesk/metadata/item.py
```diff
@@ -298,6 +298,11 @@
     'associations': {
         'type': 'dict',
     },
+    'alt_text': {
+        'type': 'string',
+        'nullable': True,
+        'mapping': not_analyzed,
```
petrjasek: are you sure about this? it's used for ids, not text fields

vied12: I don't know...
It was a suggestion from @akintolga, I think because we are already using `string` for `usageterms` and `copyrightnotice` and others...
Should we change it ?
what is the main difference ?

petrjasek: so we use it for eg. ids, because when analyzed it will split `localhost:2012-12:123` into `localhost` `2012` `12` and `123`, so then that id would match search for localhost, 2012, etc.
if it's some text which is not a single value, it should be analyzed in general. unless you want to ignore it (not to be able to search by that), but there is different setting for that

vied12: I will change it, thanks

akintolga: My comment was to change from "text" to "string" as the type of the mapping. I agree with Petr that "not_analysed" shouldn't be there either.

PR #235, opened 2016-03-15
fix(publish): fix user in item history for kill/correct

petrjasek:
for all publish actions record who did them.

SD-4055


ioanpocol:
:+1:

akintolga:
:+1:

PR #236, opened 2016-03-15
fix(init): use init_index instead of put_mapping

petrjasek:
put_mapping would fail in case index is not there,
while init_index can create it if missing.

this was in issue on travis, after a failure when gunicorn
created a new process it could call put_mapping during the time
when index was removed by other gunicorn process, and a failure
in app init stops gunicorn and thus honcho.


akintolga:
:+1:

mdhaman:
:+1:

PR #239, opened 2016-03-16
fix(storage: amazon): set extension in media_id, not url generator

actionless:


actionless:
they're passing inside docker so merging

PR #241, opened 2016-03-17
fix(elastic): use locks when setting up elastic

petrjasek:
we start multiple processes most of the time so this needs some handling


akintolga:
:+1:

mdhaman:
:+1:

PR #246, opened 2016-03-22
SD-4070 HTTP Push - send published package and its items in a single request in specified format

mugurrus:
Embed package items in the sent package for HTTP push publish
destination.

### DIFF
--- a/apps/packages/package_service.py
+++ b/apps/packages/package_service.py
@@ -23,6 +23,7 @@ from superdesk.metadata.packages import LINKED_IN_PACKAGES, PACKAGE_TYPE, TAKES_
 from apps.archive.common import insert_into_versions
 from apps.archive.archive import SOURCE as ARCHIVE
 from superdesk.utc import utcnow
+from superdesk.factory.default_settings import VERSION
 
 logger = logging.getLogger(__name__)
 package_create_signal = superdesk.signals.signal('package.create')  # @UndefinedVariable
@@ -300,8 +301,10 @@ class PackageService():
         non_root_groups = (group for group in package.get(GROUPS, []) if group.get(GROUP_ID) != ROOT_GROUP)
         for g in (ref for group in non_root_groups for ref in group.get(REFS, [])):
             if g.get(RESIDREF, '') == old_ref_id:
+                new_item = get_resource_service('archive').find_one(req=None, _id=new_ref_id)
                 g[RESIDREF] = new_ref_id
                 g['guid'] = new_ref_id
+                g[VERSION] = new_item[VERSION]
 
     def update_field_in_package(self, package, ref_id, field, field_value):
         """
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -41,7 +41,8 @@ from apps.item_autosave.components.item_autosave import ItemAutosave
 from apps.legal_archive.commands import import_into_legal_archive
 from apps.packages import TakesPackageService
 from apps.packages.package_service import PackageService
-from apps.publish.published_item import LAST_PUBLISHED_VERSION, PUBLISHED
+from apps.publish.published_item import LAST_PUBLISHED_VERSION, PUBLISHED,\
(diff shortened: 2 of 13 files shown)


apps/packages/package_service.py
```diff
@@ -300,8 +301,10 @@ def replace_ref_in_package(self, package, old_ref_id, new_ref_id):
         non_root_groups = (group for group in package.get(GROUPS, []) if group.get(GROUP_ID) != ROOT_GROUP)
         for g in (ref for group in non_root_groups for ref in group.get(REFS, [])):
             if g.get(RESIDREF, '') == old_ref_id:
+                new_item = get_resource_service('archive').find_one(req=None, _id=new_ref_id)
                 g[RESIDREF] = new_ref_id
                 g['guid'] = new_ref_id
+                g[VERSION] = new_item[VERSION]
```
akintolga: Do we need to save the version of items in a unpublished package? Until it gets published a package always points to the last version of the referenced item.

apps/publish/enqueue/enqueue_service.py
```diff
@@ -277,6 +283,24 @@ def queue_transmission(self, doc, subscribers):
...
+        """ Embeds all package items in the package document
+        """
+        for group in package.get(GROUPS, []):
+            if group[GROUP_ID] == ROOT_GROUP:
+                continue
+            for ref in group[REFS]:
+                if RESIDREF not in ref:
+                    continue
+                package_item = get_resource_service('published').find_one(req=None, item_id=ref['residRef'],
+                                                                          _current_version=ref['_current_version'])
```
akintolga: VERSION enum would be good here...

apps/publish/enqueue/enqueue_service.py
```diff
@@ -266,8 +270,10 @@ def queue_transmission(self, doc, subscribers):
                             publish_queue_item['content_type'] = doc.get('type', None)
                             publish_queue_item['headline'] = doc.get('headline', None)
                             publish_queue_item['publishing_action'] = self.published_state
+                            if doc.get(PUBLISHED_IN_PACKAGE):
+                                publish_queue_item[PUBLISHED_IN_PACKAGE] = doc[PUBLISHED_IN_PACKAGE]
                             publish_queue_item.pop(ITEM_STATE, None)
-                            get_resource_service('publish_queue').post([publish_queue_item])
+                            get_resource_service('publish_queue').post([publish_queue_item])[0]
```
akintolga: Are we doing anything with the first item in this line?

apps/publish/content/common.py
```diff
@@ -131,8 +133,8 @@ def update(self, id, updates, original):
                 self._update_archive(original, updated, should_insert_into_versions=auto_publish)
                 self.update_published_collection(published_item_id=original[config.ID_FIELD], updated=updated)
 
-                from apps.publish.enqueue import enqueue_published
-                enqueue_published.apply_async()
+            from apps.publish.enqueue import EnqueueContent
+            EnqueueContent().run()
```
mdhaman: why synchronous queuing?

apps/publish/enqueue/enqueue_service.py
sivakuna-aap: String constant for `residRef` also

apps/publish/enqueue/enqueue_service.py
mugurrus: Will remove, used it to debug

apps/packages/package_service.py
mugurrus: When embedding items in a package they are read from published collection. In this collection there can be multiple records with the same item id so I need to be able to select the one that corresponds to the published package.

ioanpocol:
:+1:

petrjasek:
would this version work for you @akintolga @mdhaman or is this something that should be configurable?

petrjasek:
my bad, config is there. :+1:

akintolga:
:+1:

PR #248, opened 2016-03-23
fix(preferences): add privileges to preferences

petrjasek:
on preferences you can set privileges and in case user doesn't
have any of them it won't show the preference in user preferences

SD-4137


mugurrus:
:+1:

akintolga:
:+1:

PR #249, opened 2016-03-23
fix(templates): remove html from headline in items created from template

petrjasek:
if you copy&paste in headline, line breaks can get there, so make sure
it doesn't make it to item which is created from template.

SD-3999


akintolga:
:+1:

mdhaman:
:+1:

PR #250, opened 2016-03-24
fix(schema): move body_footer spec to item meta

petrjasek:
it's used by both item and template, so must be defined
in base schema

SD-3980


ioanpocol:
:+1:

PR #261, opened 2016-03-31
[SD-4270] - Auto publish should not be visible on frontend

akintolga:
1. Created a new attribute: `action_type` to differentiate the return type of the macros
  1. `interactive`: macro returns a diff
  2. `direct`: story is updated directly
  3. `background-thread` is where the macro can run and no response is expected

2. Modified the `get` function to accept parameter to return `backend` macros


mdhaman:
:+1:

marwoodandrew:
:+1:

ioanpocol:
:+1:

PR #262, opened 2016-04-01
[SD-3952] Modify websocket notification to use broker for HA.

mdhaman:
If `USE_PUB_SUB_FOR_WEBSOCKETS` is set then socket notification are sent to broker (RabbitMQ/Redis). The `ws.py` process is the consumer of the messages delivered by the exchange and then the `ws.py` process broadcast the message to all connected websockets. 

@petrjasek

**This PR would require change superdesk repo in `ws.py` [Superdesk PR 1906](https://github.com/superdesk/superdesk/pull/1906).

### DIFF
--- a/superdesk/notification.py
+++ b/superdesk/notification.py
@@ -61,12 +61,12 @@ def push_notification(name, **kwargs):
         init_app(app)
 
     if not app.notification_client.open:
-        logger.info('No connection to broker. Dropping event %s' % name)
+        logger.warning('No connection to broker. Dropping event %s' % name)
         return
 
     try:
         message = _create_socket_message(event=name, extra=kwargs)
-        logger.info('Sending the message to the broker...')
+        logger.debug('Sending the message: {} to the broker.'.format(message))
         app.notification_client.send(message)
     except Exception as err:
         logger.exception(err)
--- a/superdesk/notification.py
+++ b/superdesk/notification.py
@@ -10,21 +10,23 @@
 
 """Superdesk push notifications"""
 
-import os
 import logging
-import asyncio
-import websockets
+import os
+import json
 
-from flask import current_app as app, json
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


petrjasek:
:+1: looks good
i was thinking about avoiding those ifs and having some class for the communication, but imo would be easier to eventually replace the old one altogether, what do you think?

mdhaman:
I will remove those`ifs`. I can replace the old one and use consumer/producer for socket notification. and it should work with `redis` as well.

mdhaman:
@petrjasek removed the switch between direct socket connect and broker. It uses broker directly.

akintolga:
:+1:

superdesk/notification.py
```diff
@@ -34,57 +36,37 @@ def close(self):
...
     """
     logger.info('pushing event {0} ({1})'.format(name, json.dumps(kwargs, default=json_serialize_datetime_objectId)))
-
     if not app.notification_client.open:
         app.notification_client.close()
         init_app(app)
 
     if not app.notification_client.open:
-        logger.info('No connection to websocket server. Dropping event %s' % name)
+        logger.info('No connection to broker. Dropping event %s' % name)
```
petrjasek: maybe this one should be warning

superdesk/notification.py
```diff
@@ -34,57 +36,37 @@ def close(self):
...
         return
 
     try:
-        _notify(event=name, extra=kwargs)
+        message = _create_socket_message(event=name, extra=kwargs)
+        logger.info('Sending the message to the broker...')
```
petrjasek: and here a debug message

petrjasek:
:+1: nice

mdhaman:
@petrjasek  done.

marwoodandrew:
:+1:

PR #267, opened 2016-04-05
[SD-4231] Images with some specific EXIF are not uploading to Amazon S3

jerome-poisson:
Removed metadata uploading to Amazon S3 as it is limited to 2048 bytes
(keys + values), and if the limit is exceeded the file won't upload
(MetadataTooLarge error).

Furthermore, the metadata are not needed here: they are aleady stored in
MongoDB and are still present in the file anyway.

SD-4231


petrjasek:
:+1:

akintolga:
:+1:

marwoodandrew:
:+1:

PR #269, opened 2016-04-06
fix(websockets): push some events only every x seconds

petrjasek:
it will not send ingest events more often to clients, atm it's set
to do it every 5s

SD-4013

### DIFF
--- a/superdesk/websockets_comms.py
+++ b/superdesk/websockets_comms.py
@@ -10,15 +10,18 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 
+import arrow
 import logging
 import asyncio
 import websockets
 import signal
 
+from datetime import timedelta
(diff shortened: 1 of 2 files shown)


mdhaman:
:+1:

akintolga:
:+1:

mdhaman:
Discussed these changes with @akintolga and came up with following suggestions:
- throttle `publish_queue:update`
- content expiry generates `content:update` notification for each item deleted and generate a 'separate' event `content:expired` after process is completed. Maybe completely remove the delete and use only the `content:expired` or change the `content:update` event for delete to `item:delete` and handle `item:delete` differently.

petrjasek:
added `publish_queue:update`, but this content expiry will rather handle in a new PR

PR #272, opened 2016-04-07
SDPA-366 : Clicking "Update" on an item creates a new item without th…

naomiaro:
…e old item's content type profile

### DIFF
--- a/apps/archive/archive_rewrite.py
+++ b/apps/archive/archive_rewrite.py
@@ -126,9 +126,19 @@ class ArchiveRewriteService(Service):
         fields = ['family_id', 'event_id', 'flags']
 
         if new_file:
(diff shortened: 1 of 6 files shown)


mdhaman:
Takes also create new item. You would have to change the https://github.com/superdesk/superdesk-core/blob/master/apps/packages/takes_package_service.py.

naomiaro:
OK cool, I might make a separate ticket to update takes, as this update fix is needed today for PA.

mdhaman:
:+1:  look good. But it would good to have a behave test.

mugurrus:
It would be good to have a test for this.

naomiaro:
OK I'll work on a test to make sure the key profile has been added etc.

mugurrus:
:+1:

PR #273, opened 2016-04-07
[SD-3996] Running app:prepopulate command fails

jerome-poisson:
When on_create is called with no valid request object (which is the
case when using app:prepopulate command), the "user" key must be
present in the doc dict to avoid calling request.view_args.get('user').

This was implemented by using setdefault, but it was leading to an
evaluation of request.view_args.get('user') during parsing, resulting
in a crash.

Note that the traceback fixed here is not the initial one reported in
the issue, so an other patch may follow.

SD-3996


jerome-poisson:
oups I've just seen a mistake in the patch, I'll fix it

jerome-poisson:
bad patch

PR #275, opened 2016-04-07
SDPA-348 PoI in the API ninjs response; SDPA-370 Generate renditions automatically

mugurrus:


apps/archive/archive.py
```diff
@@ -222,6 +223,21 @@ def on_update(self, updates, original):
...
 
+        updates_feature_image = updates.get('associations', {}).get('feature_image')
+        if updates_feature_image and 'poi' in updates_feature_image:
+            original_feature_image = original.get('associations', {}).get('feature_image', {})
+            if original_feature_image and original_feature_image.get('poi', {}) == updates_feature_image['poi']:
+                return
+            _id = updates_feature_image[config.ID_FIELD] if config.ID_FIELD in updates_feature_image \
+                    else original_feature_image[config.ID_FIELD]
+            archive_service = get_resource_service(ARCHIVE)
+            image_item = archive_service.find_one(req=None, _id=_id)
```
mdhaman: a minor one: `get_resource_service(ARCHIVE)` not needed. Class is Archive service.

features/content_crop.feature
```diff
@@ -39,12 +40,7 @@ Feature: Cropping the Image Articles
       }
       """
       When we publish "123" with "publish" type and "published" state
-      Then we get response code 400
-      """
-      {
-          "_issues": {"validator exception": "[['RENDITIONS is a required field']]"}
-      }
-      """
+      Then we get response code 200
```
vied12: :heart:

features/content_crop.feature
```diff
@@ -193,6 +189,59 @@ Feature: Cropping the Image Articles
...
+      And we get existing resource
+      """
+      {
+      	"associations": {
+      		"feature_image": {
+      			"_id": "bike",
+      			"poi": {"x": 0.2, "y": 0.3},
+      			"renditions": {
+      				"16-9" : {
+      					"poi" : {"y" : 216, "x" : 108}
```
vied12: poi is in pixel in rendition, but it's a proportion in the parent object. It that on purpose ?

mugurrus: Yes. The UI calculates PoI in percentage, not absolute values. The PoI outside renditions stores this value which is used to calculate the absolute values for each rendition.

features/content_crop.feature
mugurrus: Yes, vocabulary was missing from the scenario so the renditions that needed to be validated were not generated - which raised the 400 error code.

ioanpocol:
:+1:

PR #277, opened 2016-04-11
chore(eve): update eve to 0.6.3

petrjasek:
it raises now `OriginalChanged` error when there is no update done in db,
which could be the case with system update (other updates change etag).

this case is now handled in eve backend layer.

### DIFF
--- a/superdesk/tests/__init__.py
+++ b/superdesk/tests/__init__.py
@@ -10,7 +10,6 @@
 
 
 import os
-import pymongo
 import unittest
 import elasticsearch
 import logging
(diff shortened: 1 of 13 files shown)


ioanpocol:
:+1:

superdesk/tests/__init__.py
```diff
@@ -195,6 +203,9 @@ def setUp(self):
     def tearDown(self):
         if hasattr(self, 'ctx'):
             self.ctx.pop()
+        with self.app.app_context():
+            mongoclient = self.app.data.mongo.pymongo().cx
```
mugurrus: These 2 lines can be compacted into a single line

mugurrus:
:+1:

ioanpocol:
:+1:

PR #279, opened 2016-04-11
publish renditions

vied12:
Image rendition has changed and can take now a ratio as parameter.
Also only one dimension can be provided (width or height)


mugurrus:
:+1:

PR #281, opened 2016-04-11
save POI: disable exception before real fix

vied12:


apps/archive/archive.py
```diff
@@ -231,10 +231,13 @@ def on_update(self, updates, original):
                 else original_feature_image[config.ID_FIELD]
             image_item = self.find_one(req=None, _id=_id)
             if not image_item:
```
mugurrus: This if can be deleted

PR #282, opened 2016-04-12
fix(prepopulate): ignore conflicts in prepopulate

petrjasek:
in case data is there already there is nothing to prepopulate,
so ignore the resource

### DIFF
--- a/apps/prepopulate/app_prepopulate.py
+++ b/apps/prepopulate/app_prepopulate.py
@@ -11,9 +11,10 @@
 import os
 import json
 import flask
-from apps.archive.common import ITEM_OPERATION
 import superdesk
+import werkzeug.exceptions
 
+from apps.archive.common import ITEM_OPERATION
 from superdesk import get_resource_service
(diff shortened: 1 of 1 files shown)


ioanpocol:
:+1:

apps/prepopulate/app_prepopulate.py
```diff
@@ -79,7 +80,10 @@ def prepopulate_data(file_name, default_user=get_default_user()):
                 if not res:
                     raise Exception()
             else:
-                ids = service.post([data])
+                try:
+                    ids = service.post([data])
+                except werkzeug.exceptions.Conflict:
+                    ids = [data.get('_id')]  # data with given id is there already
```
jerome-poisson: why using data.get here? data['_id'] would not be better if we know that _id exists?

petrjasek: refactored this, there were items without id with conflicts, so handling this case now too

jerome-poisson:
:+1:

PR #286, opened 2016-04-13
feat(search): Add the option to search by service, category, subject …

ioanpocol:
…and genre (SDNTB-59)

### DIFF
--- a/apps/archive_broadcast/broadcast.py
+++ b/apps/archive_broadcast/broadcast.py
@@ -32,7 +32,7 @@ from superdesk.utc import utcnow
 logger = logging.getLogger(__name__)
 # field to be copied from item to broadcast item
 FIELDS_TO_COPY = ['urgency', 'priority', 'anpa_category', 'type',
-                  'subject', 'dateline', 'slugline', 'place']
+                  'subject', 'cvs', 'dateline', 'slugline', 'place']
 ARCHIVE_BROADCAST_NAME = 'archive_broadcast'
 
 
--- a/superdesk/metadata/item.py
+++ b/superdesk/metadata/item.py
@@ -141,6 +141,17 @@ metadata_schema = {
             }
         }
     },
+    'cvs': {
+        'type': 'list',
+        'nullable': True,
+        'mapping': {
+            'type': 'object',
+            'properties': {
+                'qcode': not_analyzed,
+                'name': not_analyzed,
+            }
+        }
+    },
     'genre': {
         'type': 'list',
         'nullable': True,
--- a/apps/archive/archive_rewrite.py
+++ b/apps/archive/archive_rewrite.py
@@ -134,7 +134,7 @@ class ArchiveRewriteService(Service):
             else:
                 extended_fields = [
                     'abstract', 'anpa_category', 'pubstatus', 'slugline', 'urgency',
-                    'subject', 'priority', 'byline', 'dateline', 'headline', 'place',
+                    'subject', 'cvs', 'priority', 'byline', 'dateline', 'headline', 'place',
                     'genre', 'body_footer', 'company_codes', 'keywords'
                 ]
 
--- a/apps/content_filters/filter_condition.py
+++ b/apps/content_filters/filter_condition.py
@@ -44,6 +44,7 @@ class FilterConditionResource(Resource):
                         'body_html',
                         'genre',
                         'subject',
+                        'cvs',
                         'desk',
                         'sms'],
         },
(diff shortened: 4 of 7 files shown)


apps/archive_broadcast/broadcast.py
```diff
@@ -32,7 +32,7 @@
 logger = logging.getLogger(__name__)
 # field to be copied from item to broadcast item
 FIELDS_TO_COPY = ['urgency', 'priority', 'anpa_category', 'type',
-                  'subject', 'dateline', 'slugline', 'place']
+                  'subject', 'cvs', 'dateline', 'slugline', 'place']
```
mdhaman: You might have to do the same thing in `archive_rewrite.py`.

May be move `common` fields for `take, broadcast and update` can be moved common.py?

superdesk/metadata/item.py
```diff
@@ -141,6 +141,15 @@
             }
         }
     },
+    'cvs': {
```
mdhaman: if you are planing to have aggregations on `cvs` then it would be better to make it similar to `genre` or `anpa_category`. 
```
       'mapping': {
             'type': 'object',
             'properties': {
                 'qcode': not_analyzed,
                 'name': not_analyzed
             }
         }
```

ioanpocol: Done

apps/archive_broadcast/broadcast.py
ioanpocol: Added cvs on archive_rewrite.py.

superdesk/metadata/item.py
petrjasek: why new field? seems like subject codes

petrjasek:
why new field? seems like subject codes

ioanpocol:
Looks like subject is something that is used by all implementation and also we have a search for it on advanced search page. From client to client only the list of subjects codes is different. 
The intention is to put on cvs those values that are custom for different clients.

petrjasek:
I don't see any reason why those should be separated. what we should do is adding `scheme` attribute to subject items, so you know from which cv they are

ioanpocol:
In this case it should be fixed the search issue on global search.

petrjasek:
yep

ioanpocol:
Subject is not the right name for cvs, probably better to be renamed it as cvs.
Subject is just one on many possible cvs.

ioanpocol:
If you search for 'subject' in the code you will find it in a lot of tests and there is also a kind on logic related to locator. So again I think that it will be better/cleaner if we make subject a standalone field and create cvs for new dynamic fields.

petrjasek:
thing is - you're not storing there cvs (if those mean controlled vocabularies), but values from there and that's why there is qcode (scheme + code). if those vocabularies are custom or not doesn't matter imo, behaviour should be the same. also how should this be represented on output? in eg ninjs or newsml?

ioanpocol:
For newsml seams that those values can be saved to property. Below is the text from specification(  https://www.iptc.org/std/NewsML/1.2/specification/NewsML_1.2-spec-functionalspec_9.html  )
////////////////
5.9.3 Descriptive Metadata
.......
Finally, the optional and repeatable Property element can be used to provide any additional descriptive metadata that is not explicitly provided for within the NewsML DTD.
///////////

To put all cvs values in subject will just create an invalid newsml document.
For ninjs probably the best place to put those values is "object".

But again not the export format is the issue. We use "subject" in our code, we have even an endpoint for it, we have a lot of tests, we have a kind of logic with locator and we have search on it. I think that it have enough identity to be a field and not to mix other things on it.
A different field named cvs or property or any other name will make the code cleaner.

petrjasek:
> To put all cvs values in subject will just create an invalid newsml document.
this probably depends on version, it would be valid in newsmlg2

> For ninjs probably the best place to put those values is "object".
here it depends on the values, _object_ is "Something material, excluding persons", while _subject_ is "A concept with a relationship to the content"

`subjects` being and endpoint is imo obsolete, is should be another cv, but this was done before we introduced controlled vocabularies. having tests for it is imo a plus

> A different field named cvs or property or any other name will make the code cleaner.
imo this would only introduce duplication, not sure how it would make code cleaner?

ioanpocol:
>subjects being and endpoint is imo obsolete, is should be another cv, but this was done before we introduced controlled vocabularies. having tests for it is imo a plus

From export point of view it can be used search for some export format.
Currently 'subject' related code looks not updated to that new functionality.
The endpoint, search and tests are done taken in account that 'subject' will contains one or more selection from the catalog provided by subject endpoint.

PR #287, opened 2016-04-13
[SD-4301] Fixing the problem where archive kill not working.

mdhaman:
@petrjasek for eve-elastic if `req.args` is not passed and lookup is used then it throws exception at https://github.com/petrjasek/eve-elastic/blob/master/eve_elastic/elastic.py#L358. 

I have done [PR Eve-Elastic](https://github.com/petrjasek/eve-elastic/pull/60) for that as well.

### DIFF
--- a/apps/archive/common.py
+++ b/apps/archive/common.py
@@ -19,7 +19,7 @@ from pytz import timezone
 
 import superdesk
 from superdesk.users.services import get_sign_off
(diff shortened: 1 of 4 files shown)


petrjasek:
you can use `eve-elastic==0.3.7` now

mugurrus:
:+1:

mdhaman:
@petrjasek done

petrjasek:
:+1:

PR #288, opened 2016-04-13
fix(products): fix default product id

petrjasek:
number can't be converted to objectid, so that product couldn't be used
by subscriber


akintolga:
:+1:  Thanks @petrjasek for fixing this, sorry this must have slipped...

PR #292, opened 2016-04-14
(fix) fixing the replace words keys names.

mdhaman:
Key name were not appropriate for `replace_words` vocab.

imo, later on move this vocab to some other config settings.


akintolga:
:+1:

marwoodandrew:
:+1:

petrjasek:
:+1:

PR #294, opened 2016-04-14
rename _id to guid in ninjs output

mugurrus:
As requested by PA

### DIFF
--- a/superdesk/publish/formatters/ninjs_formatter.py
+++ b/superdesk/publish/formatters/ninjs_formatter.py
@@ -54,7 +54,7 @@ class NINJSFormatter(Formatter):
 
     def _transform_to_ninjs(self, article, recursive=True):
         ninjs = {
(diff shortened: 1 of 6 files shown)


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -54,7 +54,7 @@ def format(self, article, subscriber):
 
     def _transform_to_ninjs(self, article, recursive=True):
         ninjs = {
-            '_id': article[GUID_FIELD],
+            'guid': article[GUID_FIELD],
```
petrjasek: not sure this will work, seems like in http push service `guid` is set using `_id` value and `_id` is removed anyhow

mugurrus: updated http push service

petrjasek:
:+1:

PR #299, opened 2016-04-15
fix(spike): fix push notification data

petrjasek:
it was missing item id and instead of user id it sent all user data

SD-4343


mdhaman:
:+1:  @sjunaid was complaining about this.

ioanpocol:
:+1:

PR #301, opened 2016-04-15
(fix)[SD-4068] Transfer Kill Text from take to other takes and broadcast

mdhaman:
Kill for multiple items is triggered
- for all takes and takes_package if one of the take is killed.
- for broadcast items if master item is killed.

In case of the multiple items the kill header text will be different but rest of the body_html will be same.
For example:

If there are 2 takes and Take1 is killed and `body_html` of the take1 is `Story killed due to legal reason.`
Take1: {'slugline': 'test1', 'versioncreated': '2016-04-04T00:00:00+0000',  'dateline': {'text': 'London, PA May 4'}}

Take2: {'slugline': 'test2', 'versioncreated': '2016-04-05T00:00:00+0000', 'dateline': {'text': 'London, PA May 5'}}

Th `body_html` for `Take1` will be:
```
`body_html`:<p>Please kill story slugged test1 ex London, PA May 4 at 04 May 2016 10:00 AEDT<p>
                   <p>Story killed due to legal reason.</p>
```
The `body_html` for `Take2` will be:
```
`body_html`:<p>Please kill story slugged test2 ex London, PA May 5 at 05 May 2016 10:00 AEDT<p>
                   <p>Story killed due to legal reason.</p>
```

### DIFF
--- a/apps/archive_broadcast/broadcast.py
+++ b/apps/archive_broadcast/broadcast.py
@@ -349,6 +349,7 @@ class ArchiveBroadcastService(BaseService):
                         if refs:
                             correct_service.patch(package.get(config.ID_FIELD), package_updates)
                         else:
+                            package_updates['body_html'] = updates.get('body_html', '')
                             kill_service.patch(package.get(config.ID_FIELD), package_updates)
 
                         processed_packages.add(package.get(config.ID_FIELD))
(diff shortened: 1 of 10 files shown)


petrjasek:
if we use static templates, we might need to add there more info, so maybe xml would be better?
```
<template><headline>kill</headline><body_html><p>Story killed ...</p></body_html></template>
```
but in general looks good :+1:

mdhaman:
@petrjasek modified the template to json.

petrjasek:
:+1: json looks even better

akintolga:
:+1:

PR #303, opened 2016-04-15
get_renditions_spec function added SDPA-402

vied12:
returns the internal SD renditions and the custom ones defined in vocabulary


petrjasek:
:+1:

PR #304, opened 2016-04-15
SDPA-407 Feature image is not published in composite items

mugurrus:


ioanpocol:
:+1:

tests/publish/ninjs_formatter_test.py
```diff
@@ -311,7 +311,7 @@ def test_item_with_usable_associations(self):
         formatted = json.loads(doc)
         self.assertIn('associations', formatted)
         self.assertIn('image', formatted['associations'])
-        image = formatted['associations']['image']
+        image = formatted['associations']['image'][0]
```
petrjasek: not sure this works with the specs, could it be `formatted['associations']['image0']` instead?

mugurrus: Each value in the associations dict is a list of dicts (see here: https://github.com/superdesk/superdesk-core/blob/master/superdesk/publish/formatters/ninjs_formatter.py#L146)

apps/packages/takes_package_service.py
```diff
@@ -125,6 +125,8 @@ def package_story_as_a_take(self, target, takes_package, link):
         for field in fields_for_creating_takes_package:
             if field in target:
                 takes_package[field] = target.get(field)
+        if target.get('associations', {}).get('featureimage'):
+            takes_package['associations'] = {'featureimage': target['associations']['featureimage']}
```
vied12: Is there anything to do for embedded images too ?

tests/publish/ninjs_formatter_test.py
naomiaro: I don't think they really use lists in the standard:

http://www.iptc.org/std/ninjs/ninjs-schema_1.0.json

apps/packages/takes_package_service.py
mugurrus: In a different PR

tests/publish/ninjs_formatter_test.py
mugurrus: Initially I implemented the structure according to the standard. Later @petrjasek changed it: from dict values to list of dict values because it was easier to transform from the item group structure. So currently our (not fully ninjs compatible) structure has lists of dicts values for associations dict. If we want to change this structure, that is another topic for another PR.

apps/packages/takes_package_service.py
vied12: I can't find it. It has been merged already ?

mugurrus: I mean I can do it in a different PR, it's not done yet

PR #305, opened 2016-04-18
SDPA-406 : Cannot add role to AD user

naomiaro:


naomiaro:
Together with https://github.com/superdesk/superdesk-client-core/pull/343

superdesk/users/users.py
```diff
@@ -45,22 +44,20 @@ def __init__(self, endpoint_name, app, service, endpoint_schema=None):
...
-                'type': 'string',
-                'readonly': self.readonly
+                'type': 'string'
             },
             'email': {
                 'unique': True,
                 'type': 'email',
-                'required': True
+                'required': True,
+                'readonly': self.readonly
```
mdhaman: You could create users in LDAP (for testing) that might not have email. Maybe allow the email to editable.

naomiaro: For testing if it's LDAP, a valid user should be created anyways really. Email seems to be the most important part.

petrjasek: can it be both `readonly` and `required` at the same time?

naomiaro: hmm yeah this might be a problem, it was ok for me but I should test bringing in a new user via LDAP

naomiaro: Might be better even to change sometime to a different schema depending on if it's POST/PATCH

mdhaman: Making `phone` field as editable is fine. But as per the video attached in the ticket, I cannot understand why the user does not have the `default` role assigned to the user. Does PA setup have `default` role.

mdhaman:
:+1:

PR #306, opened 2016-04-19
fix(schedule): fix publish scheduling

petrjasek:
it was using datetime in utc as if it was naive, but it gets
utc datetime from client.

SD-4377

### DIFF
--- a/apps/archive/common.py
+++ b/apps/archive/common.py
@@ -460,13 +460,12 @@ def update_schedule_settings(updates, field_name, value):
 
     schedule_settings = updates.get(SCHEDULE_SETTINGS, {}) or {}
 
-    if field_name:
+    if field_name and value:
         tz_name = schedule_settings.get('time_zone')
-        if tz_name:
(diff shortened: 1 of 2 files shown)


petrjasek:
@akintolga client is sending `publish_schedule` field in utc using the timezone so switching to timezone using that value as naive doesn't work. not sure what was the idea, maybe it should be fixed on client?

petrjasek:
so at the end fixed client to send time using local time zone

akintolga:
@petrjasek the original idea was to send datetime without any timezone info to server, as well as the selected timezone separately. On the server side we were calculating the utc by given timezone and the datetime. 

In archive schema publish_schedule field initially thought to keep the datetime (without any conversion)  as user set and utc_publish_schedule was keeping the utc version of it. And in all actions utc_publish_schedule was being used.

I think as long as server and client is consistent this problem will go away. Your approach with sending the datetime with timezone to server is looking good.

PR #307, opened 2016-04-19
fix(utc): fix local_to_utc method

petrjasek:
it was using local datetime as if it was utc

goes together with https://github.com/superdesk/superdesk-client-core/pull/348 and replaces https://github.com/superdesk/superdesk-core/pull/306


ioanpocol:
:+1:

akintolga:
:+1:

PR #311, opened 2016-04-20
Fix http transmitter exception

actionless:


superdesk/publish/transmitters/http_push.py
```diff
@@ -61,10 +61,18 @@ def _copy_published_media_files(self, item, assets_url):
             del renditions[name]['href']
             if not self._media_exists(rendition['media'], assets_url):
                 media = app.media.get(rendition['media'], resource='upload')
-                files = {'media': (rendition['media'], media, rendition.get('mimetype') or rendition['mime_type'])}
-                response = requests.post(assets_url, files=files, data={'media_id': rendition['media']})
+                files = {'media': (
+                    rendition['media'], media,
+                    rendition.get('mimetype') or rendition['mime_type']
```
mugurrus: AFAIK only mimetype is used in renditions.

mugurrus:
:+1:

superdesk/publish/transmitters/http_push.py
actionless: should i update it here then?
that commit was changing only the style

mugurrus: I would update it in this commit as it's a very small change

actionless: should i change it to just `rendition['mimetype']`?

mugurrus: yep

PR #312, opened 2016-04-20
SDPA-418 : Can no longer embed images from API

naomiaro:
Problem was when stages weren't being populated correctly because desks.json were loaded first and prevented stages.json data to be entered in mongo db.


petrjasek:
:+1:

PR #314, opened 2016-04-21
(fix)Fixing the embargo and scheduled publishing.

mdhaman:
- For `publish_schedule` and `embargo` validation is done using utc dates.
- Moved `deschedule` to `update` method so that correct state of `schedule_settings` are stored.
- `Embargo` can be modified while correcting the story.
- Remove `embargo` setting if embargo as lapsed or killed.

### DIFF
--- a/apps/publish/content/correct.py
+++ b/apps/publish/content/correct.py
@@ -11,13 +11,13 @@
 import logging
 from superdesk import get_resource_service
 from superdesk.media.crop import CropService
-from superdesk.metadata.item import ITEM_STATE, CONTENT_STATE, EMBARGO
+from superdesk.metadata.item import ITEM_STATE, CONTENT_STATE, EMBARGO, SCHEDULE_SETTINGS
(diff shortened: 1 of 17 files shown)


apps/publish/content/correct.py
```diff
@@ -35,6 +35,18 @@ class CorrectPublishService(BasePublishService):
...
     published_state = 'corrected'
 
+    def set_state(self, original, updates):
+        updates[ITEM_STATE] = self.published_state
+
+        if original.get(EMBARGO) or updates.get(EMBARGO):
+            # embargo time elapsed
+            if get_utc_schedule(original, EMBARGO) < utcnow():
+                # remove embargo information. so the next correction is without embargo.
+                updates[EMBARGO] = None
```
naomiaro: Nice! This was annoying.

akintolga:
:+1:

ioanpocol:
:+1:

petrjasek:
:+1:

PR #316, opened 2016-04-22
Ensure to do not overwrite crop area when updating PoI SDPA-420

vied12:
This remove the hard coded handler for `featureimage` and treat all the associations in the same way.

### DIFF
--- a/features/content_crop.feature
+++ b/features/content_crop.feature
@@ -232,7 +232,6 @@ Feature: Cropping the Image Articles
       }
       """
       Then we get OK response
-
       And we get existing resource
       """
       {
@@ -270,6 +269,9 @@ Feature: Cropping the Image Articles
         }
       }
       """
+      When we get "/archive/bike"
+      Then we get OK response
+      And we get no "poi"
       When we patch "/archive/feature_image"
       """
       {
--- a/features/content_crop.feature
+++ b/features/content_crop.feature
@@ -270,7 +270,38 @@ Feature: Cropping the Image Articles
         }
       }
       """
-
+      When we patch "/archive/feature_image"
+      """
+      {
+        "associations": {
+          "featureimage": {
+            "_id": "bike",
+            "poi": {"x": 0.1, "y": 0.1},
+            "renditions": {
+              "16-9": {"CropLeft":0,"CropRight":1280,"CropTop":0,"CropBottom":720}
+            }
+          }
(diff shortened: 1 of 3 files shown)


mugurrus:
:+1:

vied12:
Actually this doesn't work. By patching the associated item, we are forgetting that an associated object must be disconnected from its original. An update on the associated object shouldn't be reflected on its original.
I need to find another way to validate associated objects.

vied12:
@mugurrus can you check latest update please ?

mugurrus:
The way the associated item was implemented is wrong as it:
- duplicates content
- breaks the item structure: we do not store multiple items in a document - this is the first time we do it for the feature image. This embedding brings other issues, e.g. we have to duplicate validation code specifically for the embedded item.
Right now we don't have time to refactor, so I suggest to leave it this way, update the original document and we'll refactor later.

features/content_crop.feature
```diff
@@ -270,7 +269,41 @@ Feature: Cropping the Image Articles
...
+      And we get no "poi"
+      When we patch "/archive/feature_image"
+      """
+      {
+        "associations": {
+          "featureimage": {
+            "_id": "bike",
+            "poi": {"x": 0.1, "y": 0.1},
+            "renditions": {
+              "16-9": {"CropLeft":0,"CropRight":1280,"CropTop":0,"CropBottom":720}
```
naomiaro: It's a little strange I think for REST to PATCH something like this crop data and receive the transformation requested below.

vied12: This PR doesn't introduce this feature. And the response bellow is truncated. Only what I want to check appears.
But it's a good remark. How can it be more RESTfull from your point of view ?

mugurrus: Renditions are generated and the resulted json is an extension of the one send in the user request. I don't see a way around this difference between the user request and the resulted json in our case.

vied12: Maybe by isolated the image renditions process in another endpoint, then we don't have to patch any more the archive in order to create renditions ?

mugurrus: You could do that but what's the gain? You would have to do 2 separate requests to update an item. You would need to do extra processing to send only the relevant renditions fields to the renditions endpoint.

petrjasek:
:+1:

PR #321, opened 2016-04-25
stop testing on travis python 3.5.0

petrjasek:
that version is not in any distro, so it's only slowing down travis


akintolga:
:+1:

mdhaman:
:+1:

PR #322, opened 2016-04-26
SDPA-422 HTTP publish handling gracefully non-media configuration

mugurrus:
transmit success on accepted HTTP code

### DIFF
--- a/superdesk/publish/transmitters/http_push.py
+++ b/superdesk/publish/transmitters/http_push.py
@@ -42,7 +42,7 @@ class HTTPPushService(PublishService):
         if not queue_item.get(PUBLISHED_IN_PACKAGE) or not destination.get('config', {}).get('packaged', False):
             resource_url = destination.get('config', {}).get('resource_url')
             response = requests.post(resource_url, data=json.dumps(item), headers=self.headers)
-            if response.status_code != requests.codes.created:  # @UndefinedVariable
+            if response.status_code not in (requests.codes.created, requests.codes.accepted):  # @UndefinedVariable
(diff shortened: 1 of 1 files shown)


superdesk/publish/transmitters/http_push.py
```diff
@@ -42,7 +42,7 @@ def _transmit(self, queue_item, subscriber):
         if not queue_item.get(PUBLISHED_IN_PACKAGE) or not destination.get('config', {}).get('packaged', False):
             resource_url = destination.get('config', {}).get('resource_url')
             response = requests.post(resource_url, data=json.dumps(item), headers=self.headers)
-            if response.status_code != requests.codes.created:  # @UndefinedVariable
+            if response.status_code not in (requests.codes.created, requests.codes.accepted):  # @UndefinedVariable
```
vied12: which variable requires `@UndefinedVariable` ?

ioanpocol:
:+1:

petrjasek:
:+1:

superdesk/publish/transmitters/http_push.py
mugurrus: .created and .accepted
see https://github.com/kennethreitz/requests/blob/master/requests/status_codes.py#L14
and https://github.com/kennethreitz/requests/blob/master/requests/status_codes.py#L85

mugurrus: @UndefinedVariable is only for Eclipse to remove red underline when the code is OK but Eclipse can't figure it out.

vied12: ok, thanks !

PR #323, opened 2016-04-26
SDPA-427 : Publishing shouldn't retry delivering on 4xx http responses

naomiaro:


marwoodandrew:
Does it matter that most transmit function will not return http error codes?

superdesk/publish/transmitters/http_push.py
```diff
@@ -44,7 +44,7 @@ def _transmit(self, queue_item, subscriber):
             response = requests.post(resource_url, data=json.dumps(item), headers=self.headers)
             if response.status_code not in (requests.codes.created, requests.codes.accepted):  # @UndefinedVariable
                 message = 'Error pushing item %s: %s' % (response.status_code, response.text)
-                raise PublishHTTPPushError.httpPushError(Exception(message, destination))
+                raise PublishHTTPPushError.httpPushError(Exception(message, destination, response.status_code))
```
actionless: shouldn't it be `raise PublishHTTPPushError.httpPushError(Exception(message), destination, response.status_code)` ?

actionless: i would you recommend you to install `rainbow-parenthesis` plugin to your editor to avoid such typos in the future

naomiaro: Oh yeah, it totally should. I pretty much cowboy coded this to start while I figure out these python mocking tests...

naomiaro:
@marwoodandrew what do you mean by that? Most calls to http push won't send a response code? Why wouldn't it?

superdesk/publish/transmitters/http_push.py
naomiaro: Yeah need to move both of those out of that exception call.

superdesk/publish/publish_content.py
```diff
@@ -107,14 +107,17 @@ def transmit_subscriber_items(self, queue_items, subscriber):
...
+
             max_retry_attempt = app.config.get('MAX_TRANSMIT_RETRY_ATTEMPT')
             retry_attempt_delay = app.config.get('TRANSMIT_RETRY_ATTEMPT_DELAY_MINUTES')
             try:
                 orig_item = publish_queue_service.find_one(req=None, _id=queue_item['_id'])
                 updates = {config.LAST_UPDATED: utcnow()}
-                if orig_item.get('retry_attempt', 0) < max_retry_attempt:
+
+                # only retry on 5xx responses.
+                if orig_item.get('retry_attempt', 0) < max_retry_attempt and e.http_code >=500 and e.http_code < 600:
```
mugurrus: You have to verify if the http_code attribute exists

naomiaro: Yep, updating to check if it's a http publish error

marwoodandrew: Presumably you are catching exceptions that are thrown in the call to "transmitter.transmit(queue_item)", Most implementations of transmit will not throw exceptions that contain a http error code. Could you swallow the exceptions that you don't want to retry in the transmit function?

naomiaro: I think I'm going to split PublishHTTPPushError into 2 classes, PublishHTTPClientPushError (4xx) and PublishHTTPServerPushError (5xx) and check for PublishHTTPClientPushError when retrying.

naomiaro: Other Exceptions will be thrown like they were previous though.

superdesk/publish/publish_content.py
```diff
@@ -107,14 +107,18 @@ def transmit_subscriber_items(self, queue_items, subscriber):
...
+
             max_retry_attempt = app.config.get('MAX_TRANSMIT_RETRY_ATTEMPT')
             retry_attempt_delay = app.config.get('TRANSMIT_RETRY_ATTEMPT_DELAY_MINUTES')
             try:
                 orig_item = publish_queue_service.find_one(req=None, _id=queue_item['_id'])
                 updates = {config.LAST_UPDATED: utcnow()}
-                if orig_item.get('retry_attempt', 0) < max_retry_attempt:
+
+                if orig_item.get('retry_attempt', 0) < max_retry_attempt and \
+                        e.__class__.__name__ != 'PublishHTTPPushClientError':
```
akintolga: Here `!isinstance(e, PublishHTTPPushClientError)` would be better imo.

naomiaro: ah yes, I wanted something like that. For some reason missed this python method.

superdesk/errors.py
```diff
@@ -558,5 +558,25 @@ def httpPushError(cls, exception=None, destination=None):
...
+    def httpPushError(cls, exception=None, destination=None):
+        return PublishHTTPPushClientError(14001, exception, destination)
+
+
+class PublishHTTPPushServerError(PublishHTTPPushError):
+    _codes = {
+        14002: "HTTP push publish server error",
+    }
+
+    @classmethod
```
petrjasek: now I see why you said that those static methods are weird. imo the idea was to have them on the base class, so you can raise more specific errors without importing them all, to have it on those specific error classes is indeed useless..

naomiaro: Yeah, think it's best to just import what is needed though @petrjasek

naomiaro: All the errors are thrown with the exact same class, so it's pretty useless for catch/ isinstance

petrjasek: yep there are some errors with different codes, then it makes sense, imo the rest is there for consistency

actionless: exception handling should be refactored indeed, it started to piss me off first when i were working on some ingest stuff for VPP.

apart from the things @naomiaro mentioned already, it's really annoying what the original context of exception is usually lost, it should print trace to the logs.

and regarding the discussed topic -- why not to make exceptions inherit from some base class instead of doing this classmethod hack?

naomiaro: Yeah, I'm stuck having to pass a code as a first argument though to a SuperdeskPublishError class, so creating a static method to hold this code like the other classes was the best I think for now. (Also all these args these notifiers are looking for in the base class.

naomiaro: I've also had to cancel out these notifiers in my tests here so that I can even catch an exception in the test context.

PR #326, opened 2016-04-27
Replaced use of deprecated 'filtered', 'and', 'not' and 'query' filter (needed for SD-4323)

jerome-poisson:
The 'bool' query is used to replace them, see
https://www.elastic.co/guide/en/elasticsearch/reference/2.0/breaking_20_query_dsl_changes.html#_literal_filtered_literal_query_and_literal_query_literal_filter_deprecated

SD-3996
SD-4323

### DIFF
--- a/apps/archive/user_content.py
+++ b/apps/archive/user_content.py
@@ -25,10 +25,12 @@ class UserContentResource(Resource):
         'source': 'archive',
         'aggregations': aggregations,
         'elastic_filter': {
-            'and': [
-                {'not': {'exists': {'field': 'task.desk'}}},
-                {'not': {'term': {'version': 0}}},
-            ]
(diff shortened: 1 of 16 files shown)


ioanpocol:
@petrjasek @jerome-poisson  Can we fix this PR and update elastic search to a newer version?

jerome-poisson:
We can, but it will take time I think, as this P.R. was done a couple of months ago.

We can take it as basis to see what needs to be changed, and make a new one. Also extensive tests need to be done (usual automatic tests + lot of manual testing), as nearly everything depends on this.

jerome-poisson:
OK checked it, the thing was to replace the "and" and "filtered" queries by "bool", it should be that hard to change.

PR #328, opened 2016-04-28
fix(content_type): Add the option to check on subject the required st…

ioanpocol:
…atus of custom fields like category

### DIFF
--- a/apps/validate/validate.py
+++ b/apps/validate/validate.py
@@ -16,10 +16,34 @@ from superdesk.metadata.item import ITEM_TYPE
 logger = logging.getLogger(__name__)
 
 
+def check_json(doc, field, value):
+    if isinstance(doc, dict):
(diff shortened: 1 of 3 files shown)


apps/validate/validate.py
```diff
@@ -16,11 +16,35 @@
...
+    else:
+        return False
+
+
 class SchemaValidator(cerberus.Validator):
     def _validate_type_picture(self, field, value):
         """Allow type picture in schema."""
         pass
 
+    def _validate_mandatory_in_list(self, mandatory, field, value):
```
petrjasek: this won't be triggered when using `validator.validate` right? so how it should be used?

ioanpocol: I added the behave tests to be more clear how is used

petrjasek:
:+1:

PR #330, opened 2016-04-29
fix(ingest): raise specific error when generating token fails

petrjasek:
and ignore schedule when running ingest update via manage.py

SD-4434

### DIFF
--- a/apps/io/feeding_services/reuters.py
+++ b/apps/io/feeding_services/reuters.py
@@ -56,12 +56,8 @@ class ReutersHTTPFeedingService(HTTPFeedingService):
             provider_config = {}
             provider['config'] = provider_config
 
-        if 'url' not in provider_config:
-            provider_config['url'] = 'http://rmb.reuters.com/rmd/rest/xml'
-
-        if 'auth_url' not in provider_config:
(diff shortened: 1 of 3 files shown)


superdesk/io/commands/update_ingest.py
```diff
@@ -188,14 +188,14 @@ class UpdateIngest(superdesk.Command):
...
-            if not is_closed(provider) and is_service_and_parser_registered(provider) and is_scheduled(provider):
+            if not is_closed(provider) and is_service_and_parser_registered(provider):
                 kwargs = {
                     'provider': provider,
                     'rule_set': get_provider_rule_set(provider),
                     'routing_scheme': get_provider_routing_scheme(provider)
                 }
 
-                update_provider.apply_async(expires=get_task_ttl(provider), kwargs=kwargs)
+                update_provider(**kwargs)
```
ioanpocol: This will serialize all ingest providers updates for a single celery worker

petrjasek: it's not the celery task, but the command you can run via `python manage.py`

mdhaman: It will serialize all ingest providers.

petrjasek: right, reverted this

PR #336, opened 2016-05-03
fix(templates): Desk templates should not check for existing name com…

ioanpocol:
…pared to personal templates (SD-4105)


akintolga:
:+1:

mdhaman:
:+1:

PR #341, opened 2016-05-04
chore(requirements): reorganize and start depending on fixed werkzeug

actionless:


petrjasek:
not sure why you did this "refactoring", those libs should be set in setup.py, renaming requirements to requirements-dev is ok, but there should be no requirements file then

requirements-dev.txt
```diff
@@ -0,0 +1,7 @@
+flake8
+nose
+pep8
+pyflakes<=1.1
+wooper
+
+-r requirements.txt
```
petrjasek: no need for this on travis, it installs the library when doing `pip install .`

petrjasek:
now I see the issue, but imo it's problem with nginx config, not werkzeug

actionless:
since werkzeug wasn't the issue i can remove it

but it still seems more readable and handy during the local development to have requirements separated from setup.py

petrjasek:
dev-requirements are fine, but don't see why we should have requirements at all. you can use setup from dev-requirements via `-e .`

actionless:
will it install package itself too or only its requirements?

petrjasek:
both

PR #342, opened 2016-05-04
(fix)Fixing the prepopulate data for spellchecking and template name to be unique

mdhaman:
@ioanpocol had to change the code so that template name is unique (case insensitive).
Also, depending on the order of the keys in the payload, the behave test "Validate unique name on desk template" used to fail.
i.e. in case of error, if `is_public` field is validated first then response from the server is
```
_issues": {"template_name": "Template Name is not unique"}
```
else the response is:
```
_issues": {"template_name": ["Template Name is not unique", "Template Name is not unique"]}
```

### DIFF
--- a/apps/prepopulate/app_prepopulate_data.json
+++ b/apps/prepopulate/app_prepopulate_data.json
@@ -41,7 +41,7 @@
                     "type" : "bool",
                     "enabled" : true
                 },
-                "spellcheck:status" : {
+                "spellchecker:status" : {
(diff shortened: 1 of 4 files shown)


petrjasek:
:+1:

akintolga:
:+1:

mdhaman:
@ioanpocol  can you please review?

mugurrus:
:+1:

PR #345, opened 2016-05-05
fix(email-parser) - Preserve desk fix and field rename

akintolga:
- Desk name search is made case-insensitive
- Field urgency is renamed to News Value


mdhaman:
:+1:

PR #347, opened 2016-05-09
[SD-4457] - Resending a broadcast story to digital subscribers

akintolga:


apps/publish/enqueue/enqueue_service.py
```diff
@@ -231,6 +231,9 @@ def resend(self, doc, subscribers):
             self._resend_to_subscribers(doc, wire_subscribers, subscriber_codes)
 
         if len(digital_subscribers) > 0:
+            if is_genre(doc, BROADCAST_GENRE):
+                raise SuperdeskApiError.badRequestError('Only wire subscribers can receive broadcast stories!')
```
petrjasek: just not sure it should raise an error, because it might still send it to those wire subscribers. but not sure how a warning status code would look like

akintolga: That's true. Maybe it would be better if I check this condition before sending anything.

akintolga:
Moved the validation logic to `resend.py` module  `_validate_subscribers` function

petrjasek:
:+1:

petrjasek:
sorry closed by mistake

mdhaman:
:+1:

PR #348, opened 2016-05-09
feat(superdesk: factory: default settings): read username and passowrd env vars for mongo

actionless:
that should partly resolve https://forum.sourcefabric.org/discussion/comment/34317#Comment_34317

however i have know idea how to feed username and password to Eve-Elastic, it seems like some update there required to include auth to request if it's provided in the config: 
https://www.elastic.co/guide/en/shield/current/enable-basic-auth.html

ping @petrjasek


petrjasek:
you can specify username and password as part of `MONGO_URI`, there is no need to add extra env vars. [docs](https://docs.mongodb.com/manual/reference/connection-string/#standard-connection-string-format)

actionless:
i see

what about elastic?

petrjasek:
afaik there is no authentication built in elastic, it can be done via proxy and then it's also part of url

actionless:
but what about this one?
https://www.elastic.co/guide/en/shield/current/enable-basic-auth.html

or it should accept creds as part of url too?

petrjasek:
not sure shield works with elastic <2.0, and we don't support yet 2+

PR #352, opened 2016-05-12
new endpoint to create renditions

vied12:


apps/picture_renditions.py
```diff
@@ -0,0 +1,50 @@
+
+import superdesk
+
+from flask import current_app as app
+from superdesk.media.renditions import generate_renditions, get_renditions_spec
+
+
+def get_file(rendition, item):
+    if item.get('fetch_endpoint'):
```
mugurrus: Some docstrings would be nice

mugurrus:
Some behave tests for the endpoint would be nice to have, otherwise OK :+1:

vied12:
Just added comments and tests

petrjasek:
:+1:

PR #354, opened 2016-05-12
feat(profile): add updated_by to content profile

petrjasek:
and set it automatically when creating/updating a profile

SD-3872


mugurrus:
:+1:

mdhaman:
:+1:

PR #355, opened 2016-05-12
fix(highlights): After export highlights, the generated text content …

ioanpocol:
…item will be deleted on close item(SD-3798)


mugurrus:
:+1:

petrjasek:
:+1:

PR #357, opened 2016-05-13
feat(mail): Add support to use email servers without authentication

ioanpocol:
This PR is from Anca / liveblog. Is not affecting and existent deployment.


petrjasek:
:+1:

PR #358, opened 2016-05-13
fix(elastic): set elastic retry_on_timeout setting

petrjasek:
when there is a timeout, it will try again before raising an error


mdhaman:
:+1:

PR #361, opened 2016-05-18
fix(aggregations) - Removed unused aggregations

akintolga:
- Stages are never used in the facets
- State had been removed few months ago from UI
- Date facets were not necessary as those number were never used


mugurrus:
:+1:

mugurrus:
One scenario needs updates

mdhaman:
:+1:

PR #363, opened 2016-05-18
[SD-4491],[SD-4509] keep content that is not HTML in a <pre> tag and …

marwoodandrew:
…mark it with a format field

Removes the use of the preformatted type to indicate that the content is not strictly marked up.
Now use a new field "formatted"


apps/publish/formatters/aap_ipnews_formatter.py
```diff
@@ -79,9 +79,9 @@ def format(self, article, subscriber, codes=None):
                     odbc_item['genre'] = article['genre'][0].get('name', None)
                 else:
                     odbc_item['genre'] = 'Current'  # @genre
-                if article.get(ITEM_TYPE, CONTENT_TYPE.TEXT) == CONTENT_TYPE.TEXT:
+                if article.get(FORMAT, FORMATS.HTML) == FORMATS.HTML:
```
akintolga: Can we merge this `if` condition with the one in line: 66?

superdesk/io/feed_parsers/anpa.py
```diff
@@ -75,8 +75,8 @@ def parse(self, file_path, provider=None):
 
                 # text
                 body_lines = [l.strip() for l in text if l.startswith('\t')]
-                if item[ITEM_TYPE] == CONTENT_TYPE.PREFORMATTED:
-                    item['body_text'] = '\n'.join(body_lines)
+                if item[FORMAT] == FORMATS.PRESERVED:
```
akintolga: `item.get(FORMAT)` would be better

ioanpocol:
Please rebase

mdhaman:
:+1:

petrjasek:
:+1:

PR #365, opened 2016-05-18
feat(schema): add schema:migrate command

petrjasek:
it can detect schema updates and rebuild elastic indexes if needed

LBSD-897

### DIFF
--- a/superdesk/commands/schema.py
+++ b/superdesk/commands/schema.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2016 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+#
+
+"""Schema utils.
+"""
+
+import superdesk
+
+from flask import current_app as app
+from superdesk.lock import lock, unlock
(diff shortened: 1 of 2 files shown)


petrjasek:
@ancafarcas what do you think?
Idea would be that whenever you make changes which should trigger schema change, you would increment `SCHEMA_VERSION` in `superdesk.__init__.py`.
and that command should run when deploying changes.

akintolga:
:+1: 

One question: It looks like this will re-index all the collections including text-archive. Is that right?

ancafarcas:
@petrjasek I think it is a good idea. I would certainly adopt it.

petrjasek:
@akintolga it depends on config, if you have it in separate index it won't touch it, only the main one

superdesk/commands/schema.py
```diff
@@ -0,0 +1,82 @@
...
+from superdesk.lock import lock, unlock
+from superdesk.commands.rebuild_elastic_index import RebuildElasticIndex
+
+
+VERSION_ID = 'schema_version'
+
+
+def _get_version_db():
+    """Get db used for storing version information."""
+    return app.data.mongo.pymongo().db['superdesk']
```
actionless: db name is hardcoded

petrjasek: it's collection name not db

mugurrus:
:+1:

PR #366, opened 2016-05-19
SDPA-456 Blocked and then repeat publishing caused by "Failed to move to archived." error

mugurrus:
code profiling functionality

### DIFF
--- a/superdesk/factory/default_settings.py
+++ b/superdesk/factory/default_settings.py
@@ -252,6 +252,7 @@ INSTALLED_APPS.extend([
     'superdesk.activity',
     'superdesk.vocabularies',
     'apps.comments',
+    'superdesk.profiling',
 
     'superdesk.io',
     'superdesk.io.feeding_services',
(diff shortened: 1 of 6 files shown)


superdesk/factory/default_settings.py
```diff
@@ -415,3 +416,7 @@ def env(variable, fallback_value=None):
 
 # max transmit items to be fetched from mongo
 MAX_TRANSMIT_QUERY_LIMIT = int(env('MAX_TRANSMIT_QUERY_LIMIT', 500))
+
+# Code profiling for performance analysis
+ENABLE_PROFILING = True
```
petrjasek: imo this should be disabled by default

vied12:
:+1:  a little note about how to use it could help me to understand

vied12:
Never mind, there is an UI. I start to understand now :)

petrjasek:
would this also profile workers? or is this supposed to work only for rest api processes

mugurrus:
@petrjasek Profiling for workers implemented with the last commit. For now I added profiling for publish:enqueue and publish:transmit tasks but other can be added easily.

ioanpocol:
:+1:

petrjasek:
:+1:

PR #367, opened 2016-05-19
chore(apps: prepopulate: app_initialize): make stages, desks and content_templates updateable

actionless:


mdhaman:
This will overwrite setting in desk, stages every time the command is executed. Can we have switch in the command for this?

petrjasek:
what about adding etags handling, and only update if etag matches, so if an entity gets modified by users it won't be affected anymore? this could be generated if not provided in data itself, so something like `_etag` = `generated` and if that value doesn't match in db - leave it alone

actionless:
yup, etags could be a good part of this workaround while we are still missing an actual schema migration

petrjasek:
not sure how it's related to schema migration, this is about data

actionless:
@petrjasek  because without schema migration such data updates could cause unpredictable behaviour if schema changes

given what schema changed:
if etag won't change item will be updated to the one from prepopulation with the right schema
if it was (ie item was edited by user) when its schema will remain the same, ie app would function not correctly

petrjasek:
yep the app has to handle it - multiple versions of schema at the same time, if that's the case. we're not going to do any migrations when we would modify all items in collection afaik

actionless:
@petrjasek  because without schema migration such data updates could cause unpredictable behaviour if schema changes

given what schema changed:
if etag won't change item will be updated to the one from prepopulation with the right schema
if it was (ie item was edited by user) when its schema will remain the same, ie app would function not correctly

actionless:
here is also the related discussion:
https://docs.google.com/document/d/1SfANbF1SXDDGZas7o8cwBXHkQRKwxP5YYVMGONJrIIo/edit

PR #368, opened 2016-05-19
Published Renditions

vied12:
Publish only renditions which are defined by user in "crop_sizes" vocabulary.
And 'original'.

SDPA-452


mugurrus:
:+1:

petrjasek:
:+1:

PR #374, opened 2016-05-23
[SD-4490] - Validate Sms length before publishing

akintolga:
- Validating for publish_text and correct_text validators for the time being


marwoodandrew:
:+1:

petrjasek:
:+1:

PR #379, opened 2016-05-24
fix(takes): use mongodb when validating takes

petrjasek:
it was using mongodb to fetch takes during publishing so make
sure it gets consistent data


akintolga:
:+1:

mdhaman:
:+1:

PR #383, opened 2016-05-25
fix(ingest): add extra info to fetch notification

petrjasek:
add there item, desk and stage id

SD-4102


mdhaman:
:+1:

akintolga:
:+1:

petrjasek:
updated to have same format like `content:update` notification - will make client logic a bit simpler

PR #384, opened 2016-05-25
Refactored NITF feed parser so we can modify data mapping in settings

jerome-poisson:
Data mapping is now done in dictionnary, which can be overriden in
settings with NITF_MAPPING parameter.

Mapping dictionnary operating is explained in _generate_mapping docstring.

This has been done to implement specific behaviour requested by the NTB
project.

SDNTB-201


petrjasek:
:+1: 
just wondering what would be better approach in general, if we should implement such settings, or rather be able to register custom nitf parser, any ideas?
ping @superdesk/developers

jerome-poisson:
I'm resolving the conflict locally, I'll update this PR after.

@petrjasek if we register a custom parser we would have either to rewrite the generic fields or to do some kind of inheritance which can result in more complexe code IMO.
With the current approach, the specific settings are relatively short and easy to read/modify (as you can see on https://github.com/superdesk/superdesk/pull/2054/files ).
The only bad side I see is that it bring some minor logic into settings, and it require some technical skills (but I expect this to be done once per project, and either by developer or by instance's admin), not really a big deal.

Most of data can be mapped with a single xpath, with a pretty straightforward syntax.

actionless:
i don't like idea of using env var for this, for me it makes more sense to configure in ingest provider settings in UI

petrjasek:
@actionless it's not an env var, it's in settings config file

jerome-poisson:
I've fixed the conflict, I've moved the item[FORMAT] setting in get_format (before it was set in 2 places: in parse and could be modified in parse_meta).
Same thing for body_html which was and possibly modified in a different location

petrjasek:
as long as it's simple this could work, but once you do something more complex it would be imo hard to maintain (like [this](https://github.com/superdesk/superdesk-core/blob/master/superdesk/io/feed_parsers/nitf.py#L160-L170)).
also subclasses can be used in general for any case when extending something, while it might be hard to define how to configure it.

jerome-poisson:
in complex case you can use a callback, this is actually done for the example you're linking, get_task do the trick. Furthermore, a dictionary is far more easy to read than searching the code to find which part set the xxx key in the item.

It is also possible to have an hybrid approach: having a subclass which keep the initial dictionary and override it. In NTB case it's not necessary as the mapping are easy to implement.

Actually I would not mind to have subclasses, as long as the dictionary mapping approach is kept.

actionless:
> it's not an env var, it's in settings config file

so to change that option user have to change settings.py? or that's not something trivial which usually being changed?

jerome-poisson:
@actionless that's not trivial to change (i.e. it can't be done in the UI easily) and it need to be done once. It's actually mapping different element to item keys, as a specifity of customer (here NTB).

The change is done in settings at the moment, @petrjasek said that is could be done in a subclass which is true, but that can be done later IMO, as the current changes for NTB are trivial.

To have an idea of the change to be done, check https://github.com/superdesk/superdesk/pull/2054/files

for instance, there the abstract must be the content of body/body.content/p[@class='lead'], which is a specific request from NTB (and must be done once for the project).

jerome-poisson:
P.-S.: the request is available on the issue: https://dev.sourcefabric.org/browse/SDNTB-201, for abstract it's “ABSTRACT is in the first paragraph of the body text having class="lead""

superdesk/io/feed_parsers/nitf.py
```diff
@@ -240,5 +388,42 @@ def get_byline(self, tree):
...
+        elem = tree.find("head/meta/[@name='aap-version-creator']")
+        if elem is not None:
+            query = {'username': re.compile('^{}$'.format(elem.get('content')), re.IGNORECASE)}
+            user = superdesk.get_resource_service('users').find_one(req=None, **query)
+            if user:
+                return user.get('_id')
+        raise SkipValue()
+
+    def get_task(self, tree):
+        desk_name = tree.find('head/meta[@name="aap-desk"]')
```
actionless: i see here and in other places `aap` hardcoded, but i don't know enough details about that NITF format so i don't is that acceptable for implementation

jerome-poisson: yes it was hardcoded before the refactoring, I have let it there for now to avoid breaking the current implementation. Actually the idea of my refactoring is to avoid this kind of hardcoding in the future

jerome-poisson: Btw I have missed the :+1: from @vied12 on the top. But I haven't commit right on upstream yet, so I can't merge myself. @actionless are you OK with it?

superdesk/io/feed_parsers/nitf.py
```diff
@@ -21,163 +22,245 @@
...
+            'original_source': "head/meta/[@name='aap-source']/@content",
+            'source': "head/meta/[@name='aap-source']/@content",
+            'task': self.get_task,
+        }
+        self.metadata_mapping = None
+
+    def _parse_xpath(self, xpath):
+        """parse xpath and handle final attribute"""
+        last_idx = xpath.rfind('/')
+        last = xpath[last_idx + 1:].rstrip()
```
mugurrus: This could be moved after the first if

superdesk/io/feed_parsers/nitf.py
```diff
@@ -21,163 +22,245 @@
...
+            # an attribute is requested
+            return {'xpath': xpath[:last_idx], 'attribute': last[1:]}
+        else:
+            return {'xpath': xpath}
+
+    def _parse_mapping(self, value):
+        if isinstance(value, dict):
+            if not value:
+                return {}
+            try:
```
mugurrus: A shorter version would be:
`if 'xpath' not in value:
    return value`
So you have 2 lines instead of 5

superdesk/io/feed_parsers/nitf.py
```diff
@@ -207,28 +290,93 @@ def get_subjects(self, tree):
...
-            elements.append(etree.tostring(elem, encoding='unicode'))
+        soup = BeautifulSoup(element, 'html.parser')
+        for elem in soup.findAll(True):
+            elements.append(elem.get_text() + '\r\n')
         return ''.join(elements)
 
-    def get_norm_datetime(self, tree):
-        if tree is None:
-            return
+    def get_content(self, xml):
```
mugurrus: Should have documetation

superdesk/io/feed_parsers/nitf.py
```diff
@@ -207,28 +290,93 @@ def get_subjects(self, tree):
...
+        content = ''.join(elements)
+        if self.get_anpa_format(xml) == 't':
+            if not content.startswith('<pre>'):
+                # convert content to text in a pre tag
+                content = '<pre>{}</pre>'.format(self.parse_to_preformatted(content))
+            else:
+                content = self.parse_to_preformatted(content)
+        return content
+
+    def get_format(self, xml):
```
mugurrus: Should have documetation

superdesk/io/feed_parsers/nitf.py
```diff
@@ -207,28 +290,93 @@ def get_subjects(self, tree):
...
+            return FORMATS.PRESERVED if anpa_format == 't' else FORMATS.HTML
+
+        body_elem = xml.find('body/body.content')
+        # if the body contains only a single pre tag we mark the format as preserved
+        if len(body_elem) == 1 and body_elem[0].tag == 'pre':
+            return FORMATS.PRESERVED
+        else:
+            return FORMATS.HTML
+
+    def get_place(self, tree):
```
mugurrus: Should have documetation

mugurrus:
:+1: (aside from some minor comments)

superdesk/io/feed_parsers/nitf.py
jerome-poisson: this part is inherited from former parser, I have not written this.

superdesk/io/feed_parsers/nitf.py
jerome-poisson: indeed, done, thanks

superdesk/io/feed_parsers/nitf.py
jerome-poisson: I use exceptions because of the "It is Easier to Ask for Forgiveness than Permission" pattern which is common in Python. Also here 'xpath' is present most of time, so the EAFP version has better performances.

superdesk/io/feed_parsers/nitf.py
jerome-poisson: same thing, this part was not written by me (I have just moved it here to fix the merge conflict)

superdesk/io/feed_parsers/nitf.py
jerome-poisson: same thing as both doc issues above

jerome-poisson:
I have updated the PR according to feedback, and added comments (mainly: the undocumented parts are not from me but were there before).

PR #386, opened 2016-05-26
feat(locators): Add the option to set custom locators

ioanpocol:


superdesk/locators/__init__.py
```diff
@@ -22,7 +22,10 @@ def _load_json(file_path):
...
         return json.load(f)
 
+locators_list = []
 
-_dir_name = os.path.dirname(os.path.realpath(__file__))
-_locators_file_path = os.path.join(_dir_name, 'data', 'locators.json')
-locators = _load_json(_locators_file_path)
+
+def init_app(app):
+    _dir_name = os.path.dirname(os.path.realpath(__file__))
```
petrjasek: maybe this could read it from app config and use this as default value, so that we could change those files for deployments

PR #392, opened 2016-05-30
feat(locators): Add the option to specify the locators json file

ioanpocol:


superdesk/locators/__init__.py
```diff
@@ -24,5 +25,6 @@ def _load_json(file_path):
 
 
 _dir_name = os.path.dirname(os.path.realpath(__file__))
-_locators_file_path = os.path.join(_dir_name, 'data', 'locators.json')
+_locators_file = app.config.get('SUPERDESK_LOCATORS', 'locators.json')
```
petrjasek: imo would be better to use the folder we have for init data, so you don't have to modify core when updating it

akintolga:
:+1:

superdesk/locators/__init__.py
```diff
@@ -23,6 +25,9 @@ def _load_json(file_path):
...
 
 
-_dir_name = os.path.dirname(os.path.realpath(__file__))
-_locators_file_path = os.path.join(_dir_name, 'data', 'locators.json')
-locators = _load_json(_locators_file_path)
+def init_app(app):
+    app.locators = LocatorIndex()
+    superdesk.blueprint(bp)
+    _dir_name = os.path.dirname(os.path.realpath(__file__))
+    _locators_file_path = os.path.join(_dir_name, 'data', 'locators.json')
```
petrjasek: refactoring looks good, but reading the title imo this is the part that should be configurable

ioanpocol: This registering/init_app will be located on superdesk along with new data file, so it is not needed here. Anyway it will be difficult and error prone to load from core a file from superdesk.

petrjasek: I don't see any issue with that (we do it with app init data too). This way we are duplicating some code, and that's more error prone imo.

ioanpocol: I fixed the issue.

petrjasek:
:+1:

PR #396, opened 2016-05-31
Data updates

vied12:
It adds 3 new commands:

- `./manage.py data:generate_update  --resource {} [--global]`  
Creates a file where to define a new data update in `data_updates/`.
If `--global` is added, the data update will be created in Superdesk core folder.
- `./manage.py data:upgrade [--id {}] [--fake-init] [--dry-run]`  
Runs all the new data updates available. If `id` is given, runs new data updates until the given one.
- `./manage.py data:downgrade [--id {}] [--fake-init] [--dry-run]`  
Runs the latest data update backward. If `id` is given, runs all the data updates backward until the given one.

This is what is generated by `./manage.py data:generate_update --resource vocabularies` 

```python
# -*- coding: utf-8; -*-
# This file is part of Superdesk.
# For the full copyright and license information, please see the
# AUTHORS and LICENSE files distributed with this source code, or
#
# Author  : edouard
# Creation: 2016-05-31 16:28

from superdesk.commands.data_updates import DataUpdate


class DataUpdate(DataUpdate):

    resource = 'vocabularies'

    def forwards(self, mongodb_collection, mongodb_database):
        raise NotImplementedError()

    def backwards(self, mongodb_collection, mongodb_database):
        raise NotImplementedError()
```


superdesk/data_updates.py
```diff
@@ -0,0 +1,44 @@
...
+# at https://www.sourcefabric.org/superdesk/license
+
+from superdesk.resource import Resource
+from superdesk.services import BaseService
+import superdesk.metadata.utils
+import datetime
+import superdesk
+
+
+class DataUpdatesResource(Resource):
```
petrjasek: you could set this as `internal_resource` so there are no http endpoints created, it doesn't need it right?

petrjasek:
looks good :+1:

superdesk/data_updates.py
vied12: Good to know ! Done

ahilles107:
instead `--fake` i would call it `--dry-run` ;)

vied12:
@ahilles107 can you explain why `--dry-run` ? [On wikipedia](https://en.wikipedia.org/wiki/Dry_run_%28testing%29) it is associated to tests.
Here, this parameter is not related to tests. It is made to set the existing data updates as done, in order to do not run previously existing data updates next time you will do `./manage.py data:upgrade`.
Basically, you will add this parameter just after a fresh install, when you don't have data yet.

ahilles107:
Ahh i thought that it's used to fake system and actually just test upgrades (without persisting any data).

`Basically, you will add this parameter just after a fresh install, when you don't have data yet.` - then maybe it's better to call it `--fresh-install` or `--skip-prevoius-updates` ?

petrjasek:
what about `--init` ?

petrjasek:
:+1:

akintolga:
:+1:  Nice job.

superdesk/commands/data_updates.py
```diff
@@ -0,0 +1,255 @@
...
+                         help='Data update id to run last'),
+        superdesk.Option('--fake-init', dest='fake', required=False, action='store_true',
+                         help='Mark data updates as run without actually running them'),
+    ]
+
+    def run(self, data_update_id=None, fake=False):
+        self.data_updates_service = superdesk.get_resource_service('data_updates')
+        self.data_updates_files = get_data_updates_files(strip_file_extension=True)
+        # retrieve existing data updates in database
+        self.data_updates_applied = tuple(self.data_updates_service.get(req=None, lookup={}))
```
mugurrus: How can you be sure the docs will be sorted in the correct order?

superdesk/commands/data_updates.py
```diff
@@ -0,0 +1,259 @@
...
+    ]
+
+    def run(self, resource_name, global_update=False):
+        # create a data update file
+        timestamp = '%s%s' % (time.strftime('%Y%m%d-%H%M%S'), datetime.now().microsecond)
+        if global_update:
+            update_dir = MAIN_DATA_UPDATES_DIR
+        else:
+            update_dir = get_dirs(only_relative_folder=True)[0]
+        data_update_filename = os.path.join(update_dir, '{}_{}.py'.format(timestamp, resource_name))
```
mugurrus: Theoretically a new update can have an older timestamp than the last one if they are generated on different systems (different time zones). I think an incremented version number is better in this case. If two PRs want to merge updates with the same version the devs can figure out the order of the updates and change their PRs. This version can also be used for retrieving applied updates in the proper order.

vied12:
@mugurrus I use now an interger.
And use the same sequence between core and client's data updates.

mugurrus:
@vied12 Using a single integer means everything has to be centralized, you need to be able to read all directories containing update files, from all client repos. This is not practical. If 2 separate clients generate update files on their own repos they could end up with the same identifier for different updates.

vied12:
@mugurrus I'm not sure to see the problem here.
- the clients do not share their data update files. **global** data updates are made for that
- **This interger is used to sort the data updates**. Having two data updates with the same number can work. We just don't know which one will run first.
- **I continue to put a timestamp in the file name in order to make it unique**
- we can assume that developers have an up to date repository when they create a data update file (in order to know the last number)

Could you give me a scenario that could be a problem?

mugurrus:
There will be dependencies between update scripts, that's what the order is used for. An upgrade script expects a certain data schema. This data schema a script upgrades was generated by previous upgrade scripts from the initial data. Now, if you can't control the order of the upgrade scripts that were created by different users that may cause conflicts. It's not an urgent issue but a potential problem.

mugurrus:
:+1:

PR #401, opened 2016-06-01
keep scheduled date/time and timezone when descheduling an item

jerome-poisson:
SDNTB-127

### DIFF
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -443,8 +443,8 @@ class ArchiveService(BaseService):
         :param original: original is document.
         """
         updates[ITEM_STATE] = CONTENT_STATE.PROGRESS
-        updates[PUBLISH_SCHEDULE] = None
-        updates[SCHEDULE_SETTINGS] = {}
+        updates[PUBLISH_SCHEDULE] = original[PUBLISH_SCHEDULE]
+        updates[SCHEDULE_SETTINGS] = original[SCHEDULE_SETTINGS]
         updates[ITEM_OPERATION] = ITEM_DESCHEDULE
         # delete entry from published repo
         get_resource_service('published').delete_by_article_id(original['_id'])
--- a/features/embargo.feature
+++ b/features/embargo.feature
@@ -616,6 +616,16 @@ Feature: Embargo Date and Time on an Article (User Story: https://dev.sourcefabr
     When we get "/archive/123"
     Then we get existing resource
     """
+    {"state": "in_progress", "_id": "123", "publish_schedule": "#LAST_DATE_VALUE#"}
+    """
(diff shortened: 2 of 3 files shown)


apps/archive/archive.py
```diff
@@ -443,8 +444,10 @@ def deschedule_item(self, updates, original):
         :param original: original is document.
         """
         updates[ITEM_STATE] = CONTENT_STATE.PROGRESS
-        updates[PUBLISH_SCHEDULE] = None
-        updates[SCHEDULE_SETTINGS] = {}
+        # we need to convert schedule time to UTC
+        tz = pytz.timezone(original[SCHEDULE_SETTINGS]['time_zone'] or 'UTC')
+        localized = tz.localize(original['publish_schedule'].replace(tzinfo=None))
+        updates[PUBLISH_SCHEDULE] = localized.astimezone(pytz.utc)
```
akintolga: My opinion is that the changes on lines 448, 449, 450 may be unnecessary. We just need to keep whatever is in PUBLISH_SCHEDULE and SCHEDULE_SETTINGS fields, don't need to re-assign them.

akintolga:
A behave test to cover scenario of "descheduling a story keeps its schedule information" would be useful.

apps/archive/archive.py
jerome-poisson: @akintolga that's what I've done initially, but after doing some tests, I realized that hour was changing if I just keep original settings. I think the frontend modifying time according to time_zone (because the shift is equal to the time difference). So if I don't modify the time here, scheduled time will be changed on each deschedule.

An other option would be to not change time in frontend in case of descheduling. I did it here because I was feeling it like more safe, but I can do the other way if better.

jerome-poisson:
@akintolga indeed, there is also an embargo scenario failing that I need to fix.

apps/archive/archive.py
akintolga: @jerome-poisson I see your point. I think the issue might be on the client then.

Afaik, when a schedule saved by user, client sends date, time and timezone to server. Server uses PUBLISH_SCHEDULE field for the raw datetime value (as given by client) and uses SCHEDULE_SETTINGS to keep the UTC values.

You can test this by saving a story with schedule (without publishing) and if you re-open the story to see how the client sets date, time and timezone values; if the problem in the client you might see the same problem there as well.

jerome-poisson:
the fixed embargo test actually check that publish schedule date is still there now.

jerome-poisson:
I've removed the code converting to UTC, and filled an issue about this at the following link:
https://dev.sourcefabric.org/browse/SD-4759

petrjasek:
:+1:

PR #402, opened 2016-06-01
[SD-4565] - Filtering subscribers as 'targetable by users'

akintolga:
- New field is added in subscribers with default value True.
- It will be used in the client side to filter out which subscribers will be in the target subscribers and resend selections.


mdhaman:
:+1:

marwoodandrew:
:+1:

petrjasek:
:+1:

PR #404, opened 2016-06-02
refact(macros) - Moving macros to AAP branch

akintolga:
- Most of the AAP specific macros are moved to AAP branch
- Other clients can copy from AAP branch and modify them in their own repo/branch


petrjasek:
:+1:

mdhaman:
:+1:

PR #405, opened 2016-06-02
fix(packages): verify there is no more than 1 takes package

petrjasek:
asserting there is only 1 takes package, will monitor sentry if that raises an error


akintolga:
:+1:

ioanpocol:
:+1:

PR #410, opened 2016-06-05
SD-3487 As a user I want to be able to use a template in several desks

mugurrus:


apps/templates/content_templates.py
```diff
@@ -294,9 +311,9 @@ def render_content_template(item, template):
         elif not (isinstance(value, dict) or isinstance(value, list)):
             updates[key] = value
 
-    if template.get('template_desk'):
+    if template.get('template_desks') and len(template['template_desks']) == 1:
         updates['task'] = {}
-        updates['task']['desk'] = template.get('template_desk')
+        updates['task']['desk'] = template['template_desks'][0]
```
akintolga: Hi @mugurrus, does this line mean that when a story is created from a template, it will always be created in the first desk of the list? If so what happens:
- If the selected desk at the top dropdown is different?
- If user is not a member of the first desk?

akintolga: Sorry my mistake, I notice the line `len(template['template_desks']) == 1` later on so looks good.
:+1:

apps/templates/content_templates.py
```diff
@@ -335,7 +352,8 @@ def get_item_from_template(template):
     """
     item = template.get('data', {})
     item[ITEM_STATE] = CONTENT_STATE.SUBMITTED
-    item['task'] = {'desk': template.get('template_desk'), 'stage': template.get('template_stage')}
+    if len(template.get('template_desks', [])) == 1:
```
mdhaman: if multiple desks are assigned to a template and then if the template is used to schedule content creation then the item won't be located on any desk. We don't have a user assigned `task.user` for the scheduled content and `SUBMITTED` state is only allowed on desk. imo, if the item is created using scheduling then it would be hard to locate this item.

mugurrus:
Created another PR: https://github.com/superdesk/superdesk-core/pull/418
@mdhaman Please review the new PR.

PR #413, opened 2016-06-07
bug(ftp): Fix the ftp publish

ioanpocol:


ioanpocol:
@akintolga Please review this change

ioanpocol:
Looks like if we set the path as ./dir it works well so I will close it.

PR #414, opened 2016-06-07
NITF parser: improved mapping handling

jerome-poisson:
These 2 keys are now handled by the parser:
 - update: bool set when we want to keep default or inherited mapping
   but slighly modify the gotten value. Usefull for inheritance
 - key_hook: callback used to set the item, instead of automatically
   setting on the item with the same key as on the mapping.
   This is used in exceptional cases when a key need to be used several
   times (e.g.: subject used both for subject and category as for NTB NITF)

 Also the parser class or instance can now have a MAPPING attribute,
 which can be used for subclasses. The mapping order is default =>
 class => settings where the later override the values of the former.

SDNTB-219
SDNTB-213

### DIFF
--- a/superdesk/io/feed_parsers/nitf.py
+++ b/superdesk/io/feed_parsers/nitf.py
@@ -155,12 +155,14 @@ class NITFFeedParser(XMLFeedParser):
             return {}
 
     def _generate_mapping(self):
-        """generate self.metadata_mapping according to settings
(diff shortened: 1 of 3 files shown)


petrjasek:
would be good to add some tests

jerome-poisson:
@petrjasek test added, is it OK now ?

petrjasek:
:+1:

PR #418, opened 2016-06-08
SD-3487 As a user I want to be able to use a template in several desks

mugurrus:


vied12:
Looks good. Does it require a data update ?

mugurrus:
Theoretically yes, I don't think though that PA instances are affected by this db change.

ioanpocol:
:+1:

vied12:
@mugurrus How can we check that ? And apart from that, don't you think a migration is worth to be written?

mugurrus:
@vied12 If they had templates with stage set then they need migration, otherwise not. I don't think it's worth creating a migration script. We would need to check with AAP though if they had templates with the stage set.

apps/templates/content_templates_test.py
```diff
@@ -76,8 +76,7 @@ def test_render_content_template(self):
         template = {
             '_id': 'foo',
             'template_name': 'test',
-            'template_desk': 'sports',
-            'template_stage': 'schedule',
+            'template_desks': ['sports'],
```
akintolga: Hi @mugurrus, In this test could we assert that the created story will be in Sports desk?

apps/templates/content_templates.py
```diff
@@ -335,7 +349,8 @@ def get_item_from_template(template):
     """
     item = template.get('data', {})
     item[ITEM_STATE] = CONTENT_STATE.SUBMITTED
-    item['task'] = {'desk': template.get('template_desk'), 'stage': template.get('template_stage')}
+    if template.get('schedule_desk'):
```
mdhaman: imo, we have to raise error over here if the `schedule_desk` is not available otherwise we have item context (desk/user) problem.

mdhaman:
@mugurrus  we have templates on desk hence we would need a migration script.

apps/templates/content_templates_test.py
akintolga: Or alternatively a behave test case for "creating story out of a template that has multiple desks land on ... desk" would be great.

vied12:
@mdhaman @mugurrus I will try to make this migration.
I will push it in another PR

apps/templates/content_templates.py
mugurrus: The problem here is that the code runs in a celery task so no notification will be sent to the user if we raise an error.

apps/templates/content_templates_test.py
mugurrus: Sorry @akintolga, I don't understand what you mean.

akintolga: Sorry @mugurrus my previous comment was not clear. I wanted to say that a new behave test would be useful to cover a scenario where:
- A template with multiple desks created
- A story created from that template
- Assert that story is created on the desired desk.

apps/templates/content_templates.py
mdhaman: Correct user won;t be notified. 

My concern is that the users won't be able to search an item without user and desk. Can we exit the process from creating content and log the problem? Also change the validation for template so that `desk` information is mandatory for scheduled templates.

petrjasek:
:+1:

PR #420, opened 2016-06-09
[SD-4683] - Remove mark-up from fields before validating the length

akintolga:
- Client side character count ignores html markup
- Server side validation now will remove markup if length is checked


marwoodandrew:
:+1:

ioanpocol:
:+1:

PR #423, opened 2016-06-10
fix(publish): Fix the error on publish validation

ioanpocol:


apps/validate/validate.py
```diff
@@ -95,7 +95,11 @@ def _sanitize_fields(self, doc, validator):
         schema = validator.get('schema', {})
         for field in schema:
             if doc.get(field) and any(k in schema[field] for k in fields_to_check):
-                doc[field] = BeautifulSoup(doc[field], 'html.parser').get_text()
+                try:
+                    doc[field] = BeautifulSoup(doc[field], "html.parser").get_text()
+                except Exception:
```
vied12: A more precise exception would be safer

apps/validate/validate.py
```diff
@@ -95,7 +95,11 @@ def _sanitize_fields(self, doc, validator):
         schema = validator.get('schema', {})
         for field in schema:
             if doc.get(field) and any(k in schema[field] for k in fields_to_check):
-                doc[field] = BeautifulSoup(doc[field], 'html.parser').get_text()
+                try:
+                    doc[field] = BeautifulSoup(doc[field], "html.parser").get_text()
```
vied12: I think we use single quotes usually

akintolga:
:+1:

PR #425, opened 2016-06-14
fix(auth): fix is_authorized for missing session

petrjasek:
it fails sometimes when user is removing his current session
when it expired meanwhile. happened on travis


akintolga:
:+1:

mdhaman:
:+1:

PR #426, opened 2016-06-14
[ingest] fixed ingestion post processing when qcode is not a number

jerome-poisson:
In some cases, qcode may not be a number. This happen for NTB NITF
parser because the requested behaviour imply that category is stored in
item['subject'] with its name as qcode.
If the name is 8 characters long (e.g. "Utenriks") the post processing
is crashing, trying to find a unexisting key.

This commit check that the qcode is actually a number to avoid this
crash.

SDNTB-219


petrjasek:
:+1:

PR #427, opened 2016-06-15
fix(publish): Create a task for every publish transmit and set a smal…

ioanpocol:
…l timeout for those tasks

### DIFF
--- a/superdesk/publish/publish_content.py
+++ b/superdesk/publish/publish_content.py
@@ -96,52 +96,74 @@ def get_queue_items(retries=False):
     return get_resource_service(PUBLISH_QUEUE).get(req=request, lookup=lookup)
 
 
-@celery.task(soft_time_limit=600, bind=True)
+@celery.task(soft_time_limit=60, bind=True)
 def transmit_subscriber_items(self, queue_items, subscriber):
     # Attempt to obtain a lock for transmissions to the subscriber
     lock_name = get_lock_id("Subscriber", "Transmit", subscriber)
     if not lock(lock_name, expire=610):
         return
 
+    publish_queue_service = get_resource_service(PUBLISH_QUEUE)
     for queue_item in queue_items:
-        publish_queue_service = get_resource_service(PUBLISH_QUEUE)
-        log_msg = '_id: {_id}  item_id: {item_id}  state: {state} ' \
-                  'item_version: {item_version} headline: {headline}'.format(**queue_item)
-        try:
-            # update the status of the item to in-progress
-            logger.info('Transmitting queue item {}'.format(log_msg))
-            queue_update = {'state': 'in-progress', 'transmit_started_at': utcnow()}
-            publish_queue_service.patch(queue_item.get(config.ID_FIELD), queue_update)
-            destination = queue_item['destination']
-            transmitter = superdesk.publish.registered_transmitters[destination.get('delivery_type')]
-            transmitter.transmit(queue_item)
-            logger.info('Transmitted queue item {}'.format(log_msg))
-        except Exception as e:
-            logger.exception('Failed to transmit queue item {}'.format(log_msg))
+        lookup = {
+            '$and': [
+                {'$or': [
+                    {'state': QueueState.PENDING.value},
+                    {'state': QueueState.IN_PROGRESS.value},
+                    {'state': QueueState.RETRYING.value},
+                ]},
+                {'item_id': queue_item.get('item_id')},
+                {'published_seq_num': {'$lt': queue_item.get('published_seq_num')}}
+            ]
+        }
+        predecesors = list(publish_queue_service.get(req=None, lookup=lookup))
+        if len(predecesors):
+            continue
 
-            max_retry_attempt = app.config.get('MAX_TRANSMIT_RETRY_ATTEMPT')
-            retry_attempt_delay = app.config.get('TRANSMIT_RETRY_ATTEMPT_DELAY_MINUTES')
-            try:
-                orig_item = publish_queue_service.find_one(req=None, _id=queue_item['_id'])
-                updates = {config.LAST_UPDATED: utcnow()}
-
-                if orig_item.get('retry_attempt', 0) < max_retry_attempt and \
-                        not isinstance(e, PublishHTTPPushClientError):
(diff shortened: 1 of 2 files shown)


akintolga:
This fix will change the order of the transmissions. So if there's a correction and the story itself the correction can get to the client before the original story.

marwoodandrew:
So no guarantee that thing will actually get transmitted in the published_seq_num order?
Maybe a problem!

ioanpocol:
I fixed the issue, so all items will be sent in the right order.

superdesk/publish/publish_content.py
```diff
@@ -96,54 +96,76 @@ def get_queue_items(retries=False):
     return get_resource_service(PUBLISH_QUEUE).get(req=request, lookup=lookup)
 
 
-@celery.task(soft_time_limit=1800, bind=True)
+@celery.task(soft_time_limit=60, bind=True)
 def transmit_subscriber_items(self, queue_items, subscriber):
     # Attempt to obtain a lock for transmissions to the subscriber
     lock_name = get_lock_id("Subscriber", "Transmit", subscriber)
     if not lock(lock_name, expire=300):
```
petrjasek: imo the original problem could be this expire value which is lower than soft_time_limit, so if it gets stuck for 30m it will be able to start this task again every 5m and thus block more workers

ioanpocol: Yes, one worker becomes blocked and after 300s the lock has been released and other worker becomes blocked and so on until all workers are blocked for 30 minutes. This PR solves the block issue with one individual worker(current timeout is 10s for every transmission) and also solves the lock issue because now the lock release is 300s and the timeout for worker that request lock is 60.

superdesk/publish/publish_content.py
```diff
@@ -96,54 +96,76 @@ def get_queue_items(retries=False):
...
+            '$and': [
+                {'$or': [
+                    {'state': QueueState.PENDING.value},
+                    {'state': QueueState.IN_PROGRESS.value},
+                    {'state': QueueState.RETRYING.value},
+                ]},
+                {'item_id': queue_item.get('item_id')},
+                {'published_seq_num': {'$lt': queue_item.get('published_seq_num')}}
+            ]
+        }
```
akintolga: This query solves the problem of "Corrected story is transmitted before the original" issue. But the transmission order of the independent stories is also important. They need to be transmitted in the order of users publishing them.

akintolga:
@ioanpocol Is it possible for us to get more information regarding nature of the problem? We are publishing continuously in our UAT and we don't see any blocked workers.

ioanpocol:
@akintolga 
1. On pa and ntb there were some ftp transmission that blocked for indefinitely time. Because of this the subscriber transmission worker is blocked for 30 minutes (1800s).
2. Because the expire lock is 5 minutes (300s), after 5 minutes another worker is blocked for 30 minutes for the same items

In order to avoid worker block for long periods of time I created a task for every transmission and I set a timeout for every transmission.
Also the task that acquire the lock has a timeout less than the lock expire so we will avoid the situation that multiple workers are used for the same items.

I think that the order of individual published items should not be so important because every item contain the publish time, and a client can use this info for ordering.

If this order is mandatory we have to serialize the transmission and we can set dependencies between celery task, one will be started only after the preview one is finished (use celery subtasks/chords).

petrjasek:
maybe we can for start fix those lock expiry times and change the soft time limit and see how it goes?
it's getting complicated

ioanpocol:
We can try, it will be a small improvement, but I'm sure that in the case of multiple subscribers set will still not work well, because on subscriber transmit task we still have to set some large timeouts, change from 30 minutes to 10 probably.

petrjasek:
right, but it will block single worker instead of many, so it can be significant

superdesk/publish/publish_content.py
```diff
@@ -96,54 +96,76 @@ def get_queue_items(retries=False):
...
+        queue_item.update(queue_update)
+        publish_queue_service.patch(queue_item.get(config.ID_FIELD), queue_update)
+        transmit_subscriber_item.apply_async([queue_item])
 
     # Release the lock for the subscriber
     lock_name = get_lock_id("Subscriber", "Transmit", subscriber)
     unlock(lock_name)
 
 
+@celery.task(soft_time_limit=10, bind=True)
```
mugurrus: Soft time limit is too low. When you have items with pictures, video files, audio files this limit will certainly not permit a successful transmission.

akintolga:
@ioanpocol Thanks for the explanation know the problem is clearer. We discussed today and thought although we haven't had the issue, we can experience the same problem if ftp transmission blocked for some reason. I saw the other change to change the expiry values so let's see if that resolves the problem.

mugurrus:
:+1:

mdhaman:
We had a discussion today and also we are going live in 2 weeks time, it would be best to have a switch for `async transmision` and also move the `soft_timeout_limit` to `settings.py`.

ioanpocol:
OK, I will do the required changes.

superdesk/publish/publish_content.py
```diff
@@ -96,54 +96,76 @@ def get_queue_items(retries=False):
     return get_resource_service(PUBLISH_QUEUE).get(req=request, lookup=lookup)
 
 
-@celery.task(soft_time_limit=600, bind=True)
+@celery.task(soft_time_limit=60, bind=True)
 def transmit_subscriber_items(self, queue_items, subscriber):
     # Attempt to obtain a lock for transmissions to the subscriber
     lock_name = get_lock_id("Subscriber", "Transmit", subscriber)
     if not lock(lock_name, expire=610):
```
mdhaman: Can you also modify this line to `lock(lock_name, '', expire=610):` so that locking works on single host?

ioanpocol: I think lock expiry should be greater than soft_time_limit because otherwise if a transmission is blocked the after the lock will expire the same item will be selected to be transmitted again.

petrjasek: imo lock expiry should equal to soft_time_limit after which it will raise an exception and lock is not needed anymore

PR #430, opened 2016-06-16
fix(profiles): add default editor config when creating profile

petrjasek:
SD-4750


petrjasek:
will be part of https://github.com/superdesk/superdesk-core/pull/431

PR #431, opened 2016-06-16
fix(profiles): fix publishing of ingested items

petrjasek:
those could have data which is not set in schema,
so make validator ignore those.

SD-4688


akintolga:
:+1:

mdhaman:
:+1:

PR #433, opened 2016-06-17
(fix) Adding mongo lock for lock synchronization.

mdhaman:
- Adding  a `lock` for the `locking` the item. if multiple users try to lock an item then more than one user is getting `201 OK response`. 

@petrjasek I will share the jmeter test details on slack.


petrjasek:
:+1: great

mugurrus:
:+1:

PR #438, opened 2016-06-22
fix(profiles): use desk profile on fetch even if set

petrjasek:
it can be set already on ingest item when using
DEFAULT_CONTENT_TYPE config.

SD-4780


mdhaman:
:+1:

akintolga:
:+1:

PR #439, opened 2016-06-23
[SD-4784] Remove take key validator macro

marwoodandrew:
Make it a backend macro so it doesn't appear in the UI, but it is still available for tests


petrjasek:
:+1:

mdhaman:
:+1:

PR #440, opened 2016-06-23
Fix expired in ingest

petrjasek:
there are items in elastic without an item in mongo,
so rather remove it first from elastic and if that works
remove it from mongo, otherwise there is something showing up
that shouldn't be there and user actions fail on it.

plus remove expired ingest items from elastic which shouldn't be there
to clean up the mess.

SD-4793

### DIFF
--- a/superdesk/eve_backend.py
+++ b/superdesk/eve_backend.py
@@ -11,13 +11,13 @@
 
 import eve.io.base
 
-from flask import current_app as app
+from flask import current_app as app, json
 from eve.utils import document_etag, config, ParsedRequest
 from eve.io.mongo import MongoJSONEncoder
 from superdesk.utc import utcnow
 from superdesk.logging import logger, item_msg
 from eve.methods.common import resolve_document_etag
-from elasticsearch.exceptions import RequestError
+from elasticsearch.exceptions import RequestError, NotFoundError
 from superdesk.errors import SuperdeskApiError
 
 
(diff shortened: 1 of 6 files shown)


superdesk/eve_backend.py
```diff
@@ -189,26 +203,29 @@ def delete(self, endpoint_name, lookup):
...
         if search_backend and ids:
-            self._remove_documents_from_search_backend(endpoint_name, ids)
+            removed_ids = []
+            # first remove it from search backend, so it won't show up. when this is done - remove it from mongo
+            for _id in ids:
+                try:
+                    self.remove_from_search(endpoint_name, _id)
+                    removed_ids.append(_id)
+                except:
+                    continue
```
akintolga: It would be good to log the remove_from_search error to see which items cannot be removed. There are few other places this function is called so maybe it would be better within the function.

akintolga:
:+1:

superdesk/io/commands/remove_expired_content.py
```diff
@@ -36,25 +33,31 @@ def run(self, provider_name=None):
                 self.remove_expired(provider)
 
     def remove_expired(self, provider):
+        lock_name = 'ingest:gc'
+
+        if not lock(lock_name, expire=300):
```
mdhaman: To avoid running multiple `ingest:gc` task at the same time it would be better to use `lock(lock_name, '', expire=300)` with no host parameters.

petrjasek: if you set it to empty string it works the same like with `None`

superdesk/eve_backend.py
petrjasek: good point, will update

superdesk/io/commands/remove_expired_content.py
mdhaman: My bad, I thought it was `if host is None` then you can have two task running on two host (in HA). Negative side is more concurrency issues. 
We are running multiple celery beat so it creates multiple task and locks are host-aware as we are either passing blank or none. In this case, we are seeing two task trying to process same item twice. 
We are going to pass a dummy value for host so that there is only one lock across host. see ticket SD-4827.

petrjasek: looking at the library code, the host doesn't matter much. it's only used when you would like to extend the lock, but there can be only 1 lock with given name

mdhaman:
:+1:

PR #443, opened 2016-06-27
[SD-4827] - Concurrency issues

akintolga:
We have been observing occasional duplicate queue entries for the published stories and duplicate transmissions of the same queue item. Changed locks for enqueue and transmit workers to have the same lock across the hosts

`transmit_subscriber_items` function requires a refactoring. The list of same queue items could be retrieved by multiple workers before the lock is created. That's why we see multiple transmissions.
 
As a quick fix we have created a second check for the status of published and queue items before processing them which looks like solved the immediate problem..

### DIFF
--- a/apps/archive/commands.py
+++ b/apps/archive/commands.py
@@ -22,7 +22,6 @@ from superdesk.metadata.item import ITEM_STATE, CONTENT_STATE, ITEM_TYPE, CONTEN
 from superdesk.metadata.packages import PACKAGE_TYPE, TAKES_PACKAGE
 from superdesk.lock import lock, unlock
 from superdesk import get_resource_service
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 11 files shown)


mdhaman:
:+1:

apps/legal_archive/commands.py
```diff
@@ -332,12 +332,12 @@ class ImportLegalPublishQueueCommand(superdesk.Command):
...
         logger.info('Import to Legal Publish Queue')
         lock_name = get_lock_id('legal_archive', 'import_legal_publish_queue')
-        if not lock(lock_name, '', expire=600):
+        if not lock(lock_name, 'superdesk', expire=600):
             return
         try:
             LegalArchiveImport().import_legal_publish_queue()
         finally:
-            unlock(lock_name, '')
+            unlock(lock_name, 'superdesk')
```
petrjasek: imo this can make the issue worse, because every worker can unlock it for every other worker. so if worker A starts and it doesn't finish before expiry time, worker B will get the lock, and then worker A finishes and will unlock it. so any other worker after can get lock even if worker B is still working. and B will unlock it when finished, no matter if it got picked by other workes. etc..

akintolga: @petrjasek That's true. We were thinking to set the lock expiry equal to the soft time limit to handle this problem. I think for the en-queue and transmit tasks this is the case (we can make it for the legal_archive_import as well). I'll go through other lock cases as well.

akintolga:
@petrjasek After another discussion today we wanted to keep the lock expiry slightly greater than the soft time limit to prevent any race conditions. I have revisited all the lock case to make sure the lock expiry is properly set.

petrjasek:
sounds good, but setting the host name for locks to same value is wrong imo

mdhaman:
:+1:

apps/archive/commands.py
```diff
@@ -34,7 +33,8 @@ def run(self):
         self.log_msg = 'Expiry Time: {}.'.format(now)
         logger.info('{} Starting to remove expired content at.'.format(self.log_msg))
         lock_name = get_lock_id('archive', 'remove_expired')
-        if not lock(lock_name, '', expire=600):
+
+        if not lock(lock_name, '', expire=610):
```
mdhaman: remove empty string for hostname.

akintolga:
@petrjasek I removed the hostnames from locks but added the hostname along with the process_id in  lock.py.

petrjasek:
:+1: great

PR #445, opened 2016-06-27
fix(validate): set required to not accept empty values

petrjasek:
it can be empty string by default, so override it to be not empty

SD-4676


akintolga:
:+1:

mdhaman:
:+1:

PR #448, opened 2016-06-28
feat(profiles): implement removing of profiles

petrjasek:
- if profile is used it can't be removed
- if profile is removed also all templates which are using it are
  removed

SD-3895

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -1,6 +1,7 @@
 
 import superdesk
 
+from superdesk.errors import SuperdeskApiError
 from apps.auth import get_user_id
 from superdesk import get_resource_service
 from copy import deepcopy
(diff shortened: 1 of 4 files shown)


akintolga:
Instead of creating an in_use field, wouldn't it be better to query archive in elastic if any story with content_profile_id exists? Imo with this way it would be harder to maintain the cases like expired stories where in_use needs to revert to false if there's no other story.

apps/content_types/content_types.py
```diff
@@ -114,5 +119,21 @@ def on_create(self, docs):
             self._set_updated_by(doc)
             self._set_created_by(doc)
 
+    def on_delete(self, doc):
+        if doc.get('is_used'):
+            raise SuperdeskApiError(status_code=202, payload={"is_used": True})
+
+    def on_deleted(self, doc):
+        superdesk.get_resource_service('content_templates').delete({'data.profile': doc.get('_id')})
```
mdhaman: Why do we have to delete all templates if profile is deleted?

petrjasek: good question, ping @hlmnrmr

petrjasek:
@akintolga item could be in legal archive or it could be used by one version of an item, and I didn't want to query everything. maybe it should let admin remove it anyhow but show an extra warning if this flag is on

akintolga:
:+1:  Sounds good.

mdhaman:
:+1:

petrjasek:
removed that remove all templates using it part

mugurrus:
:+1:

PR #450, opened 2016-06-30
fix(ftp): use passive connection by default

petrjasek:
plus set timeout on ftp connections so it won't block forever.

SDNTB-17


mdhaman:
:+1:

akintolga:
:+1:

ioanpocol:
:+1:

PR #451, opened 2016-07-01
fix(templates) - Concurrency in scheduled template creation

akintolga:
- Multiple stories getting created from scheduled templates when more than one worker/host is active


mdhaman:
:+1:

PR #452, opened 2016-07-04
some personal information need to be  hidden from users

ancafarcas:
The changes below are made in order to be able to solve https://dev.sourcefabric.org/browse/LP-11

In Liveblog `client_modules` (the modules that don't need authentication), we need to hide some user personal information.

The related PR: https://github.com/liveblog/liveblog/pull/418

### DIFF
--- a/superdesk/users/users.py
+++ b/superdesk/users/users.py
@@ -15,112 +15,113 @@ from superdesk.resource import Resource
 
 
 class UsersResource(Resource):
+    readonly = True
 
-    def __init__(self, endpoint_name, app, service, endpoint_schema=None):
-        self.readonly = True if app.config.get('LDAP_SERVER', None) else False
+    additional_lookup = {
+        'url': 'regex("[\w]+")',
+        'field': 'username'
+    }
 
-        self.additional_lookup = {
-            'url': 'regex("[\w]+")',
-            'field': 'username'
+    schema = {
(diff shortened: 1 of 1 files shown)


superdesk/users/users.py
```diff
@@ -15,112 +15,113 @@
 
 
 class UsersResource(Resource):
+    readonly = True
```
mdhaman: This will make first_name,  email, last_name, etc as read_only for db authentication mode as well.

superdesk/users/users.py
```diff
@@ -15,112 +15,113 @@
...
 
-    def __init__(self, endpoint_name, app, service, endpoint_schema=None):
-        self.readonly = True if app.config.get('LDAP_SERVER', None) else False
```
ancafarcas: @mdhaman I would have kept the above verification but I'm getting "working outside context" error, since I let go the `__init__` method. Any ideas of how I could keep the `if`?

mdhaman: What exactly are you trying to do? 
Can you use projection to hide fields?

ancafarcas: @mdhaman Apparently the `__init__` method was not the problem and I managed to solve the issue by modifying the liveblog files only (fortunately). I'll close this PR since it is not useful anymore.

PR #457, opened 2016-07-07
[SDNTB-218] Ninjs: filters renditions only for pictures

vied12:
Because now we can have video

SDNTB-218


mugurrus:
:+1:

ioanpocol:
:+1:

vied12:
I didn't know that NTB does not use Ninjs, but anyway, now it's there ^^

PR #462, opened 2016-07-12
[FORMATTER][NITF] reorganized NITF formatter code to make it more eas…

jerome-poisson:
…y to subclass

this commit mainly replaces double underscore prefixes by single ones to
avoid name mangling, so subclassing is more doable.

Some code parts have been removed to specific methods, again to make
subclassing more easy.

There is also a slight behaviour modification, the former code was:
    if 'priority' in article:
	SubElement(head, 'meta', {'name': 'aap-priority', 'content': str(article.get('priority', '3'))})
This may the '3' default value could never be used as the "if" forbid the
code to be executed if article['priority'] is not set. So the "if" has
been removed.

commit needed for SDNTB-229

### DIFF
--- a/superdesk/publish/formatters/nitf_formatter.py
+++ b/superdesk/publish/formatters/nitf_formatter.py
@@ -49,28 +49,34 @@ class NITFFormatter(Formatter):
         body_head = SubElement(body, "body.head")
         body_content = SubElement(body, "body.content")
 
-        if article.get(FORMAT) == FORMATS.PRESERVED:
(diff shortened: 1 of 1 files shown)


marwoodandrew:
:+1:  AAP  should remove the AAP specific meta tags from the core class and create an AAP specific class in the AAP repo.

jerome-poisson:
@marwoodandrew I've created https://dev.sourcefabric.org/browse/AAP-100 ticket to move AAP specific elements in a new class.

PR #465, opened 2016-07-13
[SD-4715] - Add ability to test what products a content item will match

akintolga:


mdhaman:
:+1:

marwoodandrew:
:+1:  Should it be a POST?

akintolga:
This end point only for testing the products against a given article so POST or GET doesn't make much difference imo.

PR #466, opened 2016-07-13
small changes needed by liveblog

ancafarcas:


apps/packages/package_service.py
```diff
@@ -159,13 +159,14 @@ def check_not_in_personal_space(self, docs):
...
         """
         for doc in docs:
-            if not doc.get('task') or not doc['task'].get('desk'):
-                message = 'Packages can not be created in the personal space.'
-                logger.error(message)
-                raise SuperdeskApiError.forbiddenError(message=message)
-            if not doc['task'].get('stage'):
-                desk = get_resource_service('desks').find_one(req=None, _id=doc['task']['desk'])
-                doc['task']['stage'] = desk['working_stage']
+            if not doc.get('particular_type'):
```
mdhaman: where is the field `particular_type` defined?

ancafarcas: In liveblog. With this check-up I prevent the liveblog type items to go through the task/stage/desks verification because it is not needed.

PR #469, opened 2016-07-14
fix(content types): Fix the issue on prepopulation of content type re…

ioanpocol:
…: custom fields

On prepopulation if we have a custom_subject defined in editor, because
of default values it will be added also subject.
The default values should be applied only if editor and schema are not
set at all.

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -70,12 +70,10 @@ class ContentTypesResource(superdesk.Resource):
             'type': 'string',
         },
         'schema': {
-            'type': 'dict',
-            'default': DEFAULT_SCHEMA,
+            'type': 'dict'
         },
         'editor': {
-            'type': 'dict',
-            'default': DEFAULT_EDITOR,
+            'type': 'dict'
         },
         'priority': {
             'type': 'integer',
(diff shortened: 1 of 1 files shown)


apps/content_types/content_types.py
```diff
@@ -111,6 +109,9 @@ def on_create(self, docs):
         for doc in docs:
             self._set_updated_by(doc)
             self._set_created_by(doc)
+            if 'schema' not in doc and 'editor' not in doc:
+                doc['schema'] = DEFAULT_SCHEMA
+                doc['editor'] = DEFAULT_EDITOR
```
jerome-poisson: if there is "schema" and no "editor" or vice/versa there will be no default value, is it expected ?

jerome-poisson:
:+1:

apps/content_types/content_types.py
ioanpocol: This is used to make distinction between prepopulation and creation of new content item from UI.
For prepopulation the definition of content type is complete so default is not needed
For add new content type from GUI is created a content type by specify only name and description.
Also editor and schema are correlated so if one is set also the other should be set.

PR #470, opened 2016-07-15
Refactor NITF Formatter, Split off AAP specific meta data

marwoodandrew:
@jerome-poisson are you OK with removing the meta tags from the head! I doubt that they would be of much use to NTB.

Since we've been in production we've had a lot of trouble with the body html from the editor, I don't think the map_html_to_xml function is good enough.

Also in production we should not be injecting the schema definitions into the root.

### DIFF
--- a/superdesk/publish/formatters/nitf_formatter.py
+++ b/superdesk/publish/formatters/nitf_formatter.py
@@ -10,7 +10,7 @@
 
 import xml.etree.ElementTree as etree
 from xml.etree.ElementTree import SubElement
-
+from flask import current_app as app
 from superdesk.publish.formatters import Formatter
 import superdesk
 from superdesk.errors import FormatterError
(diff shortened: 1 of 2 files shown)


mdhaman:
:+1:

mdhaman:
@jerome-poisson  we need to be merged. Any issues?

jerome-poisson:
we don't use meta tags from AAP but we need to keep  the call to _append_meta_priority which is removed in this patch (you can check what we use in https://github.com/superdesk/superdesk/blob/ntb/server/ntb/publish/ntb_nitf.py )

We also use anpa_category, so this one should not be removed neither.

We don't use anymore map_html_to_xml as we had troubles with that, now we just convert the HTML to XML.

I don't see other keys that we are using in the ones removed in this patch. But please let anpa_category and the call to _append_meta_priority.

thanks

akintolga:
:+1:

jerome-poisson:
@mdhaman I made a confusion actually, we use anpa_category but in the item json, not in the produced NITF, so it need to be removed from core too, as you did initially. Can you remove it again? 

Sorry for that

Thanks

jerome-poisson:
@mdhaman @akintolga @marwoodandrew sorry to bother you with that, but NTB ask to remove the anpa_category element, and I can't do it myself because you use it in production. Sorry my previous request to keep it was a mistake (we need in in json data, not in the NITF output).

Can you remove it please?

thanks :)

marwoodandrew:
@jerome-poisson Will do, No problem!

jerome-poisson:
@marwoodandrew cool, thanks :)

PR #471, opened 2016-07-15
SD-4121 Items created in personal workspace should not be reported in the activity stream

mugurrus:
Items created in personal workspace should not be reported in the
activity stream


akintolga:
:+1:

mdhaman:
:+1:

PR #473, opened 2016-07-18
feature(content types): Add support to edit content types if cuustom …

ioanpocol:
…vocabularies are defined

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -2,6 +2,8 @@
 import superdesk
 
 from apps.auth import get_user_id
+from superdesk import get_resource_service
+from copy import deepcopy
 
 CONTENT_TYPE_PRIVILEGE = 'content_type'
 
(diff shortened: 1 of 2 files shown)


petrjasek:
is there any related ui PR ?

ioanpocol:
Not yet

apps/content_types/content_types.py
```diff
@@ -105,13 +129,67 @@ def _set_updated_by(self, doc):
...
+                    schema[field] = deepcopy(DEFAULT_SCHEMA[field])
+        else:
+            editor = doc['editor'] = deepcopy(DEFAULT_EDITOR)
+            schema = doc['schema'] = deepcopy(DEFAULT_SCHEMA)
+            default = True
+
+        # process custom fields defined on vocabularies
+        vocabularies = get_resource_service('vocabularies').find({})
+        for vocabulary in vocabularies:
+            if 'service' not in vocabulary:
```
petrjasek: this filtering could be done via `find`

apps/content_types/content_types.py
```diff
@@ -105,13 +129,67 @@ def _set_updated_by(self, doc):
     def _set_created_by(self, doc):
         doc['created_by'] = get_user_id()
 
+    def get_allowed_list(self, schema):
```
petrjasek: does it always return `['subject']` or something else too?

ioanpocol: This is to access some parts of the schema definition for subject and the 'subject' name is not changed (if changed also the mongodb field will change).

petrjasek:
:+1:

PR #474, opened 2016-07-18
fix(filemeta): store filemeta json encoded

petrjasek:
filemeta could have inconsistent types for certain exif keys from
different cameras etc, which can confuse elastic auto mapping and
lead to conflicts there. so storing it json encoded instead will
avoid such conflicts in filemeta, no matter what is there.

SD-4964


akintolga:
:+1:

ioanpocol:
:+1:

PR #476, opened 2016-07-19
changes needed for liveblog

ancafarcas:
This PR is a copy of https://github.com/superdesk/superdesk-core/pull/466 but I removed some commits that are unnecessary for the moment and they were causing some failures.

### DIFF
--- a/apps/packages/package_service.py
+++ b/apps/packages/package_service.py
@@ -159,6 +159,9 @@ class PackageService():
         items if it can't remove it from the packages it belongs to.
         """
         for doc in docs:
+            # the next 'particular_type' containing line is liveblog related and it is usefull since
+            # liveblog users don't have a personal space. Therefore, if the the task/desk condition
+            # is in place, it prevents the liveblog posts from beeing published.
             if not doc.get('particular_type'):
                 if not doc.get('task') or not doc['task'].get('desk'):
                     message = 'Packages can not be created in the personal space.'
--- a/apps/packages/package_service.py
+++ b/apps/packages/package_service.py
@@ -159,13 +159,14 @@ class PackageService():
         items if it can't remove it from the packages it belongs to.
         """
         for doc in docs:
-            if not doc.get('task') or not doc['task'].get('desk'):
-                message = 'Packages can not be created in the personal space.'
-                logger.error(message)
-                raise SuperdeskApiError.forbiddenError(message=message)
-            if not doc['task'].get('stage'):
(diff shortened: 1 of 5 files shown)


apps/packages/package_service.py
```diff
@@ -159,13 +159,14 @@ def check_not_in_personal_space(self, docs):
...
         """
         for doc in docs:
-            if not doc.get('task') or not doc['task'].get('desk'):
-                message = 'Packages can not be created in the personal space.'
-                logger.error(message)
-                raise SuperdeskApiError.forbiddenError(message=message)
-            if not doc['task'].get('stage'):
-                desk = get_resource_service('desks').find_one(req=None, _id=doc['task']['desk'])
-                doc['task']['stage'] = desk['working_stage']
+            if not doc.get('particular_type'):
```
akintolga: How can a document with this field saved into database? Is it from ingest?

ioanpocol: This is a liveblog field, so better to add this condition in liveblog by extending package_service and initialize the extended class instead of original one.

ancafarcas: I tried to move the condition in liveblog but it seems to be very complicated to extend package_service in a place where I've already extended ArchiveService which comprises package_service. So, this condition is finally not accepted in superdesk?

petrjasek: I think it could be fine, as long as you add some comment why it is working like that, not sure what `particular_type` means

ancafarcas: :+1:  I'll add a comment similar to the following
`particular_type` is a liveblog specific field telling us weather the material that we are posting is "item"  or "post". For us, a post is  a package and consists in one or more items.
Since we don't have tasks and desks in liveblog, posting in the personal space verification is not needed so this `particular_type` check-up is preventing the personal space verification to be executed.

akintolga:
:+1:

petrjasek:
I think you might need to rebase it to get rid of all those other commits

superdesk/__init__.py
```diff
@@ -13,7 +13,7 @@
 import blinker
 import logging as logging_lib
 from flask import abort, json, Blueprint, current_app as app  # noqa
-from flask.ext.script import Command as BaseCommand, Option  # noqa @UnresolvedImport
```
petrjasek: I think here you should keep the `Option`, but remove `@UnresolvedImport` comment
and that would be the case for the other lines too

petrjasek:
:+1:

PR #481, opened 2016-07-22
(fix) [SD-5077] Moving invisible stages to user collection to reduce queries.

mdhaman:
- This reduces the no. of calls to get the invisible stages for each search operation.
- Downside is Desk/Stage CRUD takes bit longer.
- It also includes schema migration for users collection.

### DIFF
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -74,7 +74,11 @@ def private_content_filter():
             {'term': {'original_creator': str(user['_id'])}},
         ]}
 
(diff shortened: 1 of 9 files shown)


akintolga:
:+1:

marwoodandrew:
:+1:

petrjasek:
:+1: but maybe we should implement some generic cache if there are more usecases like this, otherwise maintaining could get quite complex

mdhaman:
Agree that we need a generic cache.

PR #485, opened 2016-07-22
[Publishers] better file extension handling

jerome-poisson:
file output now use get_file_extension for better consistency with
other publishers.

get_file_extension now returns 'txt' as default instead of None.
Extensions can be re-used if a format use the same name (e.g. NTB NITF
will use NITF extension by default).

register_file_extension can be used to change the extension for a
specific format

needed to fix SDNTB-252

### DIFF
--- a/superdesk/publish/publish_service.py
+++ b/superdesk/publish/publish_service.py
@@ -19,6 +19,10 @@ from superdesk.utc import utcnow
 from superdesk.errors import SubscriberError, SuperdeskPublishError, PublishQueueError
 
 logger = logging.getLogger(__name__)
+extensions = {
+    'NITF': 'ntf',
+    'XML': 'xml',
+    'NINJS': 'json'}
 
 
 class PublishService():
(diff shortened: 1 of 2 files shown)


akintolga:
:+1:

petrjasek:
:+1:

akintolga:
@jerome-poisson We noticed that this PR has removed the wiring of the file extension from the subscriber settings.

image

As you can see we used to set the file extension from the destination configuration of the subscriber for file-output delivery before but now it looks like we need to hard code the extension to a value in our NITF formatter.

I understand the change is necessary for NTB and I presume the problem was in FTP delivery settings as originally there's no file extension setting. I wanted to check with you if it is possible to add the file extension to the subscriber settings for FTP and make it configurable from there?

Pinging @marwoodandrew @mdhaman @petrjasek

petrjasek:
so the current handling would be used as default and it would be possible to change it via config?
sounds good to me

PR #488, opened 2016-07-25
fix(kill): On kill ignore any validation on slugline and body_html

ioanpocol:


akintolga:
Validators collection should have a `kill_text` validator. I think we can remove body_html and slugline from the schema on that validator (_in validator.json_) to have the same affect without a code change.

mugurrus:
:+1:

akintolga:
Sorry @ioanpocol probably you missed my earlier comment but I don't think hard-coding those fields for kill were necessary. I thought we can change the validator.json to tell which fields to check during which publish action. What do you think?

ioanpocol:
@akintolga If the content type is defined the validations defined on validation.json is skip.
There is a proposal that only body + headline + abstract will be editable on kill template (https://dev.sourcefabric.org/browse/SD-5107) so it will be safe to hardcode those values.
If this is not working for you I can change the code to apply this kind of validation only if a content profile is used.

akintolga:
As far as I can see, with the current validators we have in production, it doesn't work for us as we validate headline and body_html in validators. 

As a quick fix you can add an `if content_profile:` section but in a month time _(when we start using content_profiles)_ this also won't work for us again. I think the validation fields need to be dynamic, either by content_profiles or validators.

ioanpocol:
On content profile we have only publish validation compared with validators where there are also kill type of validation.
Because kill operation changes only body + headline + abstract, I think it is not needed dynamic validation, just ignore any error on those fields if profile is defined.
If you think that it is still needed dynamic validation on kill we have to decide how we add that setting on content profile definition and also to decide if correct type of validation is needed (currently on validation are defined publish, correct and kill validators).

ioanpocol:
https://github.com/superdesk/superdesk-core/pull/499

akintolga:
@ioanpocol thanks for the other fix.

PR #493, opened 2016-07-26
company_codes field added.

eozan:
SD-4968

### DIFF
--- a/superdesk/metadata/item.py
+++ b/superdesk/metadata/item.py
@@ -155,6 +155,17 @@ metadata_schema = {
             }
         }
     },
+    'company_codes': {
+        'type': 'list',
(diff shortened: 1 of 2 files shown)


superdesk/metadata/item.py
```diff
@@ -155,6 +155,17 @@
             }
         }
     },
+    'company_codes': {
```
mdhaman: Can you remove `company_codes` definition from https://github.com/superdesk/superdesk-core/blob/master/apps/archive/common.py#L138?

eozan: done.

ioanpocol:
:+1:

petrjasek:
:+1:

PR #499, opened 2016-07-27
fix(kill): On kill ignore any validation on slugline, abstract and bo…

ioanpocol:
…dy_htm if profile is set


petrjasek:
:+1:

akintolga:
:+1:

PR #502, opened 2016-07-28
(fix)[SD-5150][SD-5140] Modify kill for NZN content.

mdhaman:
- Making template folder configurable per customer repo. 
- Allowing multiple emails for broadcast kills.

@petrjasek


apps/templates/content_templates.py
```diff
@@ -273,6 +273,9 @@ def create(self, docs, **kwargs):
         doc = docs[0] if len(docs) > 0 else {}
         template_name = doc.get('template_name')
         item = doc.get('item') or {}
+        desk = get_resource_service('desks').find_one(req=None, _id=item.get('task', {}).get('desk'))
```
petrjasek: this could probably use `get_desk_name` method

superdesk/factory/app.py
```diff
@@ -66,13 +66,11 @@ def get_app(config=None, media_storage=None, config_object=None):
         settings=app_config,
         json_encoder=MongoJSONEncoder,
         validator=SuperdeskValidator,
-        template_folder=os.path.join(abs_path, 'templates'))
+        template_folder=app_config.get('TEMPLATE_PATH') or os.path.join(abs_path, 'templates'))
```
petrjasek: imo this should be kept as it was to have the fallback

superdesk/factory/app.py
```diff
@@ -66,13 +66,11 @@ def get_app(config=None, media_storage=None, config_object=None):
...
 
     superdesk.app = app
 
-    template_paths = [abs_path + '/../templates']
-    for template_path in app_config.get('CUSTOM_TEMPLATE_PATH'):
-        template_paths.append(template_path)
+    template_paths = [app_config.get('TEMPLATE_PATH') or os.path.join(abs_path, 'templates')]
 
     custom_loader = jinja2.ChoiceLoader([
         app.jinja_loader,
```
petrjasek: and if we define custom template path, it should add that `FileSystemLoader(template_paths)` before `app.jinja_loader`, so that we could override the default (core/templates) via settings

petrjasek:
actually, not sure if we need to have it configurable, we could just set it to `templates` and `core/templates` via that choice loader, so what you would have in client repo templates folder would have higher priority, but you woudn't have to change everything

superdesk/factory/app.py
petrjasek: what about:
```
app.jinja_loader = jinja2.ChoiceLoader([
    FileSystemLoader('templates'),  # using cwd, this should be in superdesk repo `server/templates` folder
    FileSystemLoader(os.path.join(
        os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
        'templates'
    ),
])
```
this should take it from superdesk repo `server/templates`, if not there from core repo `superdesk/templates`

apps/templates/content_templates.py
mdhaman: forgot to change that line. will do that.

superdesk/factory/app.py
mdhaman: will try this.

mdhaman:
@petrjasek I tried to override the template from the client repo folder but it didn't work. Hence, I took the approach of separating templates and it has the side effects as you have mentioned.  I will try what you have suggested.

petrjasek:
:+1:

akintolga:
:+1:

PR #505, opened 2016-07-29
feat(superdesk): add allowed_values resource

petrjasek:
you can set allowed values for certain key and use it on client
to filter certain options.

SD-5118


mdhaman:
:+1:

akintolga:
:+1:

PR #508, opened 2016-08-02
Adds a vidible endpoint to fetch metadata

vied12:
Moved from `pa-poc` to https://github.com/superdesk/superdesk-core after the front end part has moved with https://github.com/superdesk/superdesk-client-core/pull/768/


petrjasek:
:+1:

PR #509, opened 2016-08-02
[publishing] encoding handling in formatter and transmitter

jerome-poisson:
formatter can now specify the final encoding using "transmitter_encoding" encoding, to do this they can return a list of dictionnaries instead of a list of tuples.
Only file_output transmitter was testing if "formatted_item" is a bytes
or str, so this part was removed, and it is now assumed that
"formatted_item" is always str, while encoded_item will be bytes.

needed for SDNTB-253

### DIFF
--- a/apps/publish/enqueue/enqueue_service.py
+++ b/apps/publish/enqueue/enqueue_service.py
@@ -10,10 +10,12 @@
 
 from functools import partial
 import logging
+import io
 import json
 
 from bson import ObjectId
 
+from flask import current_app as app
 from superdesk import get_resource_service
 from superdesk.errors import SuperdeskApiError, SuperdeskPublishError
 from superdesk.metadata.item import CONTENT_TYPE, ITEM_TYPE, ITEM_STATE, PUBLISH_SCHEDULE
@@ -320,15 +322,21 @@ class EnqueueService:
                                                           subscriber,
                                                           subscriber_codes.get(subscriber[config.ID_FIELD]))
 
-                        for pub_seq_num, formatted_doc in formatted_docs:
-                            publish_queue_item = dict()
+                        for idx, publish_data in enumerate(formatted_docs):
+                            if not isinstance(publish_data, dict):
+                                pub_seq_num, formatted_doc = publish_data
+                                formatted_docs[idx] = {'published_seq_num': pub_seq_num,
+                                                       'formatted_item': formatted_doc}
+                            else:
+                                assert 'published_seq_num' in publish_data and 'formatted_item' in publish_data,\
+                                    "missing keys in publish_data"
+
+                        for publish_queue_item in formatted_docs:
                             publish_queue_item['item_id'] = doc['item_id']
                             publish_queue_item['item_version'] = doc[config.VERSION]
-                            publish_queue_item['formatted_item'] = formatted_doc
                             publish_queue_item['subscriber_id'] = subscriber[config.ID_FIELD]
                             publish_queue_item['codes'] = subscriber_codes.get(subscriber[config.ID_FIELD])
                             publish_queue_item['destination'] = destination
-                            publish_queue_item['published_seq_num'] = pub_seq_num
                             # publish_schedule is just to indicate in the queue item is create via scheduled item
                             publish_queue_item[PUBLISH_SCHEDULE] = get_utc_schedule(doc, PUBLISH_SCHEDULE) or None
                             publish_queue_item['unique_name'] = doc.get('unique_name', None)
@@ -339,6 +347,13 @@ class EnqueueService:
                                 ObjectId(doc.get('ingest_provider')) if doc.get('ingest_provider') else None
                             if doc.get(PUBLISHED_IN_PACKAGE):
                                 publish_queue_item[PUBLISHED_IN_PACKAGE] = doc[PUBLISHED_IN_PACKAGE]
+                            try:
+                                encoded_item = publish_queue_item.pop('encoded_item')
+                            except KeyError:
+                                pass
+                            else:
+                                binary = io.BytesIO(encoded_item)
+                                publish_queue_item['encoded_item_id'] = app.storage.put(binary)
                             publish_queue_item.pop(ITEM_STATE, None)
                             get_resource_service('publish_queue').post([publish_queue_item])
                             queued = True
--- a/superdesk/publish/publish_queue.py
+++ b/superdesk/publish/publish_queue.py
@@ -15,6 +15,7 @@ from superdesk.notification import push_notification
 from superdesk.resource import Resource
 from superdesk.services import BaseService
 from superdesk.utils import SuperdeskBaseEnum
+from flask import current_app as app
 
 logger = logging.getLogger(__name__)
 
@@ -38,6 +39,8 @@ class PublishQueueResource(Resource):
         'item_version': {'type': 'integer', 'nullable': False},
 
         'formatted_item': {'type': 'string', 'nullable': False},
+        'item_encoding': {'type': 'string', 'nullable': True},
+        'encoded_item_id': {'type': 'objectid', 'nullable': True},
         'subscriber_id': Resource.rel('subscribers'),
         'codes': {'type': 'list', 'nullable': True},
         'destination': {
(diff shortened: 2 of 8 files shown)


akintolga:
@marwoodandrew With this change the default encoding will be utf-8. Would you mind please pulling this change and checking if our ascii encoded publish get affected?

marwoodandrew:
Does this work for saving files as bytes?

jerome-poisson:
@akintolga beside file_output, all transmitters were expecting "formatted_item" to be str (so unicode), I was wondering why the file_output was reacting differently, but this should not be the case.

I was more in favor of doing the encoding directly in the formatter, but after talking with @petrjasek we have decided to keep unicode for previewing, and encode only before publishing.

Let me know if this cause trouble.

@marwoodandrew to save files as byte you need to specify the encoding (e.g. for NTB I use 'transmitter_encoding': 'iso-8859-1'). What is your use case exactly ?

marwoodandrew:
The ANPA formater produces bytes, It's used in AAP production to supply contents to a legacy Lotus Notes based service.

jerome-poisson:
@marwoodandrew how do you store bytes in database ? formatted_item is flagged as string, so I guess binary data should cause troubles. In any case, I think it's a bad thing to have different behavior with transmitters (i.e. file_output handling bytes or str, and not the others).

@petrjasek wanted to keep unicode, because binary would make preview harder

An option to handle this may be to have 2 values: formatted_item for str, or encoded_item for bytes.
If the formatter wishes to produce byte it create encoded_item itself, in addition to formatted_item.
If not it can give formatted_item with on optional encoding (which default to "utf-8").

This way I think we can keep best of 2 options: we have unicode for previewing, and bytes for binary data.

@petrjasek @marwoodandrew @akintolga what do you think ?

jerome-poisson:
I've updated the P.R.
Now you can send directly bytes in "encoded_item" if you need some binary format output. In this case you still need to provide a "formatted_item" (which can be used for previewing). If it's not possible, just use an empty str or whatever text which would make sense.

Encoding can also be specified in "item_encoding". In this case, and if encoded_item doesn't already exist, formatted_item will be encoded just before being transmitted. If "item_encoding" doesn't exists, utf-8 will be used for encoding.

This way bytes version is only stored when it is needed, and formatted_item (str version) is still available for preview or anything else. It should be flexible enough for all use cases.

apps/publish/enqueue/enqueue_service.py
```diff
@@ -320,15 +320,20 @@ def queue_transmission(self, doc, subscribers, subscriber_codes={}):
                                                           subscriber,
                                                           subscriber_codes.get(subscriber[config.ID_FIELD]))
 
-                        for pub_seq_num, formatted_doc in formatted_docs:
-                            publish_queue_item = dict()
+                        for idx, publish_data in enumerate(formatted_docs):
```
petrjasek: why this change?

jerome-poisson: to be able to return a dictionary, like here: https://github.com/superdesk/superdesk/pull/2207/files#diff-5b639099f2d4d0f4d8acad9623b95628R44

petrjasek: I see, would be good to have some test for it in core

jerome-poisson: right, it's used in ntb branch test but not in core, I'll add a test

marwoodandrew:
Need to be sure that the resend feature in the publish queue feature will still work.

jerome-poisson:
I've added a test to check that dictionary with correct key pass while it fails with missing keys, and that tuple (like before) pass, but fails with bad number of items

jerome-poisson:
@marwoodandrew resend should take the same code path as publish, so it should not be affected. But as said before you have to do a slight modification in your code (use encoded_item(bytes) instead of formatted_item(str), and put either a unicode friendly version in formatted_item or an empty string). As you were already using formatted_item, I guess you can use the same value for encoded_item and formatted_item in your particular case.

If you use the ANPA formatter in production, it's probably a good idea to pull this P.R. to test with it before we do the merge.

I'm waiting for your ack.

@superdesk/superdesk-core waiting also for code review by the way.

thanks

marwoodandrew:
@jerome-poisson sorry I didn't get a chance to try your changes to the core today, but will try first thing tomorrow.

jerome-poisson:
@marwoodandrew ok thanks

marwoodandrew:
@jerome-poisson I pulled your PR and had a look!

I made the required changes to the formatters in the AAP repo and discovered the following 

- The publish_queue API end point returns a 500 when there are items in the publish queue that have the encoded_item populated.


  __File "/home/marwood/superdesk/server/src/superdesk-core/superdesk/storage/desk_media_storage.py", line 27, in get
TypeError: id must be an instance of (bytes, str, ObjectId), not <class 'bytes'>
    _id = bson.ObjectId(_id)
  File "/home/marwood/miniconda3/envs/superdesk/lib/python3.4/site-packages/bson/objectid.py", line 110, in __init__
    self.__validate(oid)
HTTP Exception 500 has been raised: None
  File "/home/marwood/miniconda3/envs/superdesk/lib/python3.4/site-packages/bson/objectid.py", line 209, in __validate
    "not %s" % (text_type.__name__, type(oid)))__ 

- When you use the resend button from the "publish queue" page for an item the that has been published, the client does not pass the item_encoding in the publish_queue POST request. So by default it gets encoded as utf-8. I have no idea what happens for an item that has the encoded_item populated!

marwoodandrew:
I've put a PR with the required changes to the AAP repo see superdesk/superdesk-aap#80

jerome-poisson:
Seeing that encoded_item is not used in @marwoodandrew PR and it's causing trouble, I've removed it (it is still available in the transmitter, but it can't be specified anymore in the formatter). If in the future a formatter need to produce binary data, we can handle this case at the moment.

the second point is not handled yet.

marwoodandrew:
In server/aap/publish/formatters/anpa_formatter.py the bytes version of the story is put into the encode_item! I thought that was the point?

         docs.append({'published_seq_num': pub_seq_num, 'encoded_item': b''.join(anpa),
 +                             'formatted_item': b''.join(anpa).decode('ascii')})

jerome-poisson:
@marwoodandrew or right, it was indeed the point, but I've missed it because I've seen  server/aap/publish/formatters/aap_nitf_formatter.py, I've missed the second formatter.

OK, so encoded_item need to be settable in formatter, I'll update the P.R. again.

jerome-poisson:
Ok I have updated the P.R., "encoded_item" can be filled again by formatter, an objectid type is now used instead of media. 

thanks @marwoodandrew for your feedback.

marwoodandrew:
@jerome-poisson Pulled your PR again, publish_queue get is OK with binary now! 

When you track down the missing item_encoding on resend you might want to make sure the encoded_item_id is there as well!!

Will the binary items in the grid fs storage be properly deleted when items expire from the queue?

jerome-poisson:
PR updated, item_encoding and encoded_item_id are restored from db if missing (checked with resend, it works as expected now).

@marwoodandrew good point about the storage deletion, I do it on on_delete, but I'm not sure about the exact internal workflow there, so it's still missing a test. I'll need some input on monday to check if it's done correctly.

superdesk/publish/publish_queue.py
```diff
@@ -126,6 +129,15 @@ def on_create(self, docs):
                 subscriber = subscriber_service.find_one(req=None, _id=doc['subscriber_id'])
                 doc['published_seq_num'] = subscriber_service.generate_sequence_number(subscriber)
 
+            # if keys are not returned by client (can happen on resend), we need to get them for database
+            keys = {'item_encoding', 'encoded_item_id'}
+            if not keys.issubset(doc):
+                item = get_resource_service("publish_queue").find_one(req=None, item_id=doc['item_id'])
+                if item is not None:
```
marwoodandrew: There may be multiple entries in the publish queue with the same "item_id", If there are multiple format for example.

jerome-poisson: I didn't knew we could have several item_id, I've tried with pick_fields but it doesn't seems to work. Unfortunately I spend most of my days in meetings, I could really find the origin of the problem, will continue on this tomorrow.

jerome-poisson:
PR updated

marwoodandrew:
:+1:

petrjasek:
:+1:

PR #510, opened 2016-08-03
fix(highlight): Prepopulate fields from highlight package and highlig…

ioanpocol:
…ht text generated content item


mugurrus:
:+1:

petrjasek:
:+1:

PR #512, opened 2016-08-03
fix(dictionaries): store big dictionaries into gridfs

petrjasek:
when it gets bigger, store it in gridfs, otherwise keep it in doc

SD-4986


mugurrus:
:+1:

akintolga:
:+1:

mdhaman:
:+1:

petrjasek:
@akintolga updated that store method, so it will handle those cases properly.

akintolga:
:+1:

PR #514, opened 2016-08-08
Feature/s3 version enhancement

nistormihai:


nistormihai:
I need review

petrjasek:
:+1: so this creates a new folder on amazon every 15m right?

nistormihai:
yes, but not only that. I added also the option to opt out for versioning or to give one as you want.

petrjasek:
cool

mugurrus:
:+1:

PR #515, opened 2016-08-09
SD-5181 : Provide some basic curation functions for Text Archive

naomiaro:


superdesk/commands/delete_text_doc.py
```diff
@@ -0,0 +1,58 @@
...
+                    '_index': _index,
+                    '_type': _type,
+                    '_id': guid,
+                }
+
+                bulk_commands.append(data)
+
+        try:
+            es = get_es(superdesk.app.config['ELASTICSEARCH_URL'])
+            elasticsearch.helpers.bulk(es, bulk_commands)
```
akintolga: I think `superdesk.get_resource_service('..').delete` deletes from both elastic and mongo so this section may not be needed..

superdesk/commands/delete_text_doc.py
```diff
@@ -0,0 +1,58 @@
...
+        except elasticsearch.helpers.BulkIndexError as BulkIndexError:
+            print(BulkIndexError)
+        except elasticsearch.exceptions.NotFoundError as nfe:
+            print(nfe)
+
+        try:
+            superdesk.get_resource_service('archive').delete(lookup={'_id': {'$in': guids}})
```
akintolga: The ticket asks for a delete from Text Archive so the collection should be `archived`. It doesn't say the deletion from the `archive` which is the production.

superdesk/commands/delete_text_doc.py
```diff
@@ -0,0 +1,58 @@
...
+        except Exception as e:
+            print(e)
+
+        try:
+            superdesk.get_resource_service('legal_archive').delete(lookup={'_id': {'$in': guids}})
```
akintolga: Again ticket doesn't ask for a deletion from Legal Archive. Theoretically nothing should be deleted from legal archive so this needs to be removed.

superdesk/commands/delete_text_doc.py
```diff
@@ -0,0 +1,58 @@
...
+        except Exception as e:
+            print(e)
+
+        try:
+            superdesk.get_resource_service('legal_archive_versions').delete(lookup={'_id_document': {'$in': guids}})
```
akintolga: The same as above.

akintolga:
@naomiaro I know this PR is still `wip` but added few comment so let me know if you have any questions. Thanks.

naomiaro:
aha, @akintolga thanks. Yeah I mostly posted this for comments. I thought this was meant for totally deleting some article from the system, not just the archived section. I'll check out if ES is included in that delete command.

ioanpocol:
@naomiaro Please rebase this PR

naomiaro:
manage.py app:deletedoc -i urn:newsml:localhost:2016-09-05T11:54:49.869091:6bdbd02b-f6c5-46d3-bc53-4c1fc74252af

superdesk/commands/reassign_metadata.py
```diff
@@ -0,0 +1,56 @@
...
+        category_list = superdesk.get_resource_service('vocabularies').find_one(req=None, _id='categories')
+        category_dict = {entry['qcode']: entry for entry in category_list['items']}
+        patched = value
+
+        try:
+            if key not in ['anpa_category', 'source']:
+                raise Exception('Invalid meta value to update.')
+
+            # http://blog.labix.org/2008/06/27/watch-out-for-listdictkeys-in-python-3
+            if key == 'anpa_category' and value not in list(category_dict.keys()):
```
naspeh: You need just:
```
- value not in list(category_dict.keys())
+ value not in category_dict
```
it checks key existence in dict

superdesk/commands/reassign_metadata.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def run(self, guids, key, value):
+        guids = guids.split(",")
+        category_list = superdesk.get_resource_service('vocabularies').find_one(req=None, _id='categories')
+        category_dict = {entry['qcode']: entry for entry in category_list['items']}
+        patched = value
+
+        try:
+            if key not in ['anpa_category', 'source']:
+                raise Exception('Invalid meta value to update.')
```
naspeh: You don't need to wrap this `if` to `try...except` and then raise Exception...

It is good practice to put in `try...except` as less code as possible.

superdesk/commands/reassign_metadata.py
```diff
@@ -0,0 +1,56 @@
...
+                    'subject': entry.get('subject', None),
+                    'scheme': entry.get('scheme', None)
+                }]
+
+            service = superdesk.get_resource_service('archive')
+
+            for guid in guids:
+                print(service.patch(guid, {key: patched}))
+        except Exception as e:
+            print(e)
```
naspeh: You need there:
```
raise SystemExit(e)
```
Instead of just `print` because command fails and you want to see code 1 instead of 0.

ref: https://docs.python.org/3.5/library/exceptions.html#SystemExit

naspeh:
It is good to have some tests for this, like here: https://github.com/superdesk/superdesk-core/blob/master/tests/rebuild_elastic_index_tests.py

naomiaro:
@naspeh thanks for the input! I'll work on these points tomorrow.

superdesk/commands/delete_text_doc.py
```diff
@@ -0,0 +1,27 @@
...
+        -i urn:newsml:localhost:2016-09-05T11:54:49.869091:6bdbd02b-f6c5-46d3-bc53-4c1fc74252af
+
+    '''
+    option_list = [
+        superdesk.Option('--guids', '-i', dest='guids', required=True,
+                         help='pass guids for articles to delete')
+    ]
+
+    def run(self, guids):
+        guids = guids.split(",")
```
petrjasek: guid can contain `,` eg reuters: `tag:reuters.com,0000:newsml_TRE7AA2CP:472094323`

superdesk/commands/reassign_metadata.py
petrjasek: or you might remove that try/except altogether and let it raise an error

superdesk/commands/reassign_metadata.py
```diff
@@ -0,0 +1,56 @@
...
+            if key == 'anpa_category':
+                entry = category_dict[value]
+                patched = [{
+                    'qcode': entry.get('qcode'),
+                    'name': entry.get('name'),
+                    'subject': entry.get('subject', None),
+                    'scheme': entry.get('scheme', None)
+                }]
+
+            service = superdesk.get_resource_service('archive')
```
mdhaman: This should be `archived` resource not `archive`

superdesk/commands/delete_text_doc.py
naomiaro: hmm that's tricky if they want to comma separate guids, maybe they can use another character.

petrjasek: some providers use [taguri](http://www.faqs.org/rfcs/rfc4151.html)
what about reading list of guids from input, one per line?

naspeh: Nope, just use `superdesk.Option(action='appned'...)`

and then you can put multiple `-i` options

mdhaman: We might want to delete 100s of documents due to contractual agreement.  A `file` options could  be handy for batch delete. We can do `mongoexport` of guids and feed it to the `deletecommand`. 

Moreover, it would be nice to have an option for `batch size` for deleting documents.

PR #519, opened 2016-08-12
fix(metadata): set item fields nullable to match default values config on client

petrjasek:
to be able to reset it


akintolga:
:+1:

marwoodandrew:
:+1:

PR #524, opened 2016-08-16
fix(content-filters) - Keywords field can't operate with string operators

akintolga:
- Keywords field is a list so only `in` and `not-in` operators can be used


mdhaman:
:+1:

petrjasek:
:+1:

PR #526, opened 2016-08-17
fix(core): add picture_crop to core apps

petrjasek:
and make sure every app is inited only once

SD-5124


thnkloud9:
:thumbsup:

mdhaman:
:+1:

PR #527, opened 2016-08-17
fix(dictionary): remove BOM from file input if present

petrjasek:
this would make first word from a file not usable,
because it would be read with bom prefix, so better filter it out.

SD-5227


akintolga:
:+1:

mdhaman:
:+1:

PR #529, opened 2016-08-18
fix(rewrite) - Rewritten item is deleted if it is spiked

akintolga:
- This PR is to handle the deleted rewrite but there'll be one later to prevent the orphans. 
- Publishing of an item fails rewrite is deleted.


mdhaman:
:+1:

marwoodandrew:
:+1:

PR #531, opened 2016-08-19
fix(publish): copy associations info to takes package

petrjasek:
it contains info about embedded items in body_html,
which is required futher when published.

SD-5254


mdhaman:
:+1:

ahilles107:
👍

PR #532, opened 2016-08-19
[NITF] [Formatter] separate title from _format_head

jerome-poisson:
title is now set in a separate _append_title, so it can be put before
meta, and be the first element in <head>

requested in SDNTB-254 feedback


petrjasek:
:+1:

marwoodandrew:
:+1:

mugurrus:
:+1:

PR #537, opened 2016-08-24
[publish] PublishService and filename generation refactoring

jerome-poisson:
PublishService base class can now be changed using set_publish_service,
this allow overriding in specific superdesk flavours without modifying current code.

filename in file_output and ftp now use a common get_filename method,
which is part of PublishService so it can be overridden

ftp filename pattern has changed, it is now common with file_output and
use by default version and sequence number.

needed for SDNTB-264

### DIFF
--- a/superdesk/publish/transmitters/file_output.py
+++ b/superdesk/publish/transmitters/file_output.py
@@ -8,24 +8,22 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-from superdesk.publish.publish_service import PublishService, get_file_extension
+from superdesk.publish.publish_service import PublishService
 from superdesk.publish import register_transmitter
 from superdesk.errors import PublishFileError
+from os import path
 
 errors = [PublishFileError.fileSaveError().get_error_description()]
 
 
 class FilePublishService(PublishService):
     def _transmit(self, queue_item, subscriber):
-        config = queue_item.get('destination', {}).get('config', {})
         try:
-            # use the file extension from config if it is set otherwise use extension for the format
-            extension = config.get('file_extension') or get_file_extension(queue_item)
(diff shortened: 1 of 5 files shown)


superdesk/publish/transmitters/file_output.py
```diff
@@ -8,24 +8,22 @@
...
-            with open('{}/{}-{}-{}.{}'.format(config['file_path'],
-                                              queue_item['item_id'].replace(':', '-'),
-                                              str(queue_item.get('item_version', '')),
-                                              str(queue_item.get('published_seq_num', '')),
-                                              extension), 'wb') as f:
+            config = queue_item['destination']['config']
+            file_path = config['file_path']
+            if not path.isabs(file_path):
+                file_path = "/" + file_path
+            with open(path.join(file_path, '{}'.format(PublishService.get_filename(queue_item))), 'wb') as f:
```
petrjasek: here you can remove that `'{}'.format` part imo

petrjasek:
:+1: but can use some unittests for those `get_file*` methods

superdesk/publish/transmitters/file_output.py
jerome-poisson: oh right, I've changed this one and forgot to remove this format, it's useless here indeed, thanks.

akintolga:
Not related directly with this PR but I have comment regarding if the Publish Service is the best place to have file extension functionality. It looks to me that file extension doesn't concern the Publish Service. It is more of Transmission concern where if the transmission is a file-based one then we worry about the file extension. So to me this function can be a candidate for refactoring to move towards transmission domain.

jerome-poisson:
@akintolga the get_filename is only used by file based transmitters (ftp and file_output), PublishService only give suggested filename on demand. It doesn't seems wrong to me to put it there.

PR #538, opened 2016-08-24
fix(authoring): make version 0 items invisible

petrjasek:
in order not to be able to see those in different session and remove
it via unlocking

SD-4737


akintolga:
I think we may also need to change the filter in [/apps/search/__init__.py](https://github.com/superdesk/superdesk-core/blob/master/apps/search/__init__.py#L43) to prevent them appearing from search results as well.

mdhaman:
:+1:

PR #541, opened 2016-08-26
fix(publish): fix validation when using default values

petrjasek:
it didn't know any rule for default, thus it was failing for
content profiles with default values.


jerome-poisson:
:+1:

naspeh:
👍

mugurrus:
:+1:

PR #542, opened 2016-08-26
fix(publish): ignore schema fields set to None

petrjasek:
this would fail in validator in some cases (when it expects integer),
so ignore all fields set to Null.


akintolga:
:+1:

PR #544, opened 2016-08-30
fix(highlight) Fixing the query to return items for generating highlight.

mdhaman:
User are getting confused as the current query return only 10 records.


akintolga:
:+1:

petrjasek:
:+1:

petrjasek:
released as 1.1.1

PR #547, opened 2016-08-30
[publishing] fixed associated items validation

jerome-poisson:
associated items was failing when an ITEM_STATE was not set (happening
with NTB Scanpix search provider for instance). This commit fix it by
using a default value of PUBLISHED in this case.

SD-5409


petrjasek:
:+1:

PR #548, opened 2016-08-30
fix(publish): update associated items before publishing

petrjasek:
in case those embedded items fail validation, there is no way to fix
it because it keeps a copy of the item. now it will try to fetch that
item from archive and update basic metadata, in case this got fixed
there it will use it for validation.

SD-5408


mugurrus:
This should fix some validation and editing issues. I implemented this functionality some time ago but it was rejected for some reason I don't remember. Hope this will be merged though.
:100:

mdhaman:
:+1:

PR #549, opened 2016-08-30
fix(fetch): set profile to None for media type items

petrjasek:
instead of using default profiles, so it will use default
validation which shouldn't require body_html for picture items

SD-5300


akintolga:
:+1:

mdhaman:
:+1:

PR #550, opened 2016-08-30
[NITF] [formatter] moved some fields creation to a separated method

jerome-poisson:
date.expire, doc-id and dateline has been moved to separated methods so they can
be overriden in subclasses
_append_xxx method have been renamed to _format_xxx for consistency

needed for SDNTB-254


petrjasek:
:+1:

PR #551, opened 2016-09-01
[SD-5387] - Remove more_coming field

akintolga:

PR #553, opened 2016-09-01
fix(item): add more_coming field back to schema

petrjasek:
it won't be used anymore, but there will be items already having it
and any further action on them would fail due this unknown field.

so partially reverting https://github.com/superdesk/superdesk-core/pull/551

/cc @akintolga


akintolga:
I see. Thanks for this.
:+1:

PR #554, opened 2016-09-01
fix(updates): avoid creating of directories in get_data_update_files

petrjasek:
there is race condition when multiple workers are starting they can
try to create those folders, but only one will succed and others will
get OSError.

/cc @vied12

### DIFF
--- a/superdesk/commands/data_updates.py
+++ b/superdesk/commands/data_updates.py
@@ -76,7 +76,7 @@ def get_data_updates_files(strip_file_extension=False):
     # create folder if doens't exist
     for folder in get_dirs():
         if not os.path.exists(folder):
-            os.makedirs(folder)
+            continue
         # list all files from data updates directory
         if os.path.exists(folder):
             files += [f for f in os.listdir(folder) if os.path.isfile(os.path.join(folder, f))]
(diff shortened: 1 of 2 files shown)


superdesk/commands/data_updates.py
```diff
@@ -76,7 +76,7 @@ def get_data_updates_files(strip_file_extension=False):
     # create folder if doens't exist
     for folder in get_dirs():
         if not os.path.exists(folder):
-            os.makedirs(folder)
+            continue
```
ioanpocol: Now the folder is not created at all. Is this ok?

petrjasek: I'm not sure, @vied12 ?

vied12: I think we need this folder to be created somewhere. Maybe putting a `try...except OSError: pass` around `os.makedirs(...)` could solve the initial problem ?

petrjasek: it has even `exists_ok` param that would ignore that error, but I would rather not create anything in `get_` method, so will update this to create those when writing the file

ioanpocol:
:+1:

vied12:
:+1:

PR #556, opened 2016-09-02
[SD-5402] Category mapping for PA ingest

marwoodandrew:


petrjasek:
:+1:

ioanpocol:
I think this parser should be located on superdesk/pa branch since it is custom for PA

marwoodandrew:
This is for AAP (or anyone else) to parse PA sourced NITF! There could be an argument that it belongs in the AAP branch. The problem was that we were routing the content based in the IPTC code, but it seems not all articles has IPTC codes present. So mapped the PA specific category to ANPA categories in Superdesk.

ioanpocol:
OK, now I understood.

ioanpocol:
:+1:

PR #557, opened 2016-09-02
fix(ingest): sync ingest items in elastic when using routing

petrjasek:
while routing it updates the item with archived info, which also
changes etag, but those changes were later overriden in elastic
via bulk index update, so client would get wrong etag and wouldn't
be able to update/remove such item.

SD-5416


mdhaman:
:+1:

ioanpocol:
:+1:

PR #558, opened 2016-09-02
[NITF] [FORMATTER] fixed doc-id

jerome-poisson:
method generating doc-id element was not called, this commit fix it

SDNTB-280


marwoodandrew:
Ah! We don't have a test that makes sure we have a doc-id?

PR #559, opened 2016-09-05
[NITF] [FORMATTER] added distributor method

jerome-poisson:
added a distributor method which actually do nothing, but can be
overriden in subclass to add <distributor/> element in the right
location.

needed for SDNTB-280


petrjasek:
:+1:

PR #561, opened 2016-09-05
[SETTINGS] use local timezone in DEFAULT_TIMEZONE when it is not expl…

jerome-poisson:
…icitly set

This commit add a new dependency: tzlocal

DEFAULT_TIMEZONE was defaulting on Europe/Prague. This commit fix it by using the timezone actually set
on the running machine.

SDNTB-285

### DIFF
--- a/superdesk/factory/default_settings.py
+++ b/superdesk/factory/default_settings.py
@@ -15,6 +15,7 @@ import json
 from datetime import timedelta
 from celery.schedules import crontab
 from kombu import Queue, Exchange
+import tzlocal
 
 try:
     from urllib.parse import urlparse
@@ -359,7 +360,13 @@ ADMINS = [MAIL_FROM]
 SUPERDESK_TESTING = (env('SUPERDESK_TESTING', 'false').lower() == 'true')
 
 # Default TimeZone
-DEFAULT_TIMEZONE = env('DEFAULT_TIMEZONE', 'Europe/Prague')
+DEFAULT_TIMEZONE = env('DEFAULT_TIMEZONE')
+
+if DEFAULT_TIMEZONE is None:
+    DEFAULT_TIMEZONE = tzlocal.get_localzone().zone
+
+if not DEFAULT_TIMEZONE:
+    raise ValueError("DEFAULT_TIMEZONE is empty")
 
 # The number of minutes since the last update of the Mongo auth object after which it will be deleted
 SESSION_EXPIRY_MINUTES = int(env('SESSION_EXPIRY_MINUTES', 240))
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ install_requires = [
     'python-magic>=0.4',
     'python3-ldap>=0.9.8',
     'pytz>=2015.4',
+    'tzlocal>=1.2.2',
     'raven[flask]>=5.10.0,<5.11',
     'requests>=2.7.0',
     'statsd>=3.1',


naspeh:
Maybe UTC is enough for default timezone?

For example django is using it:
https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-TIME_ZONE
> New project templates default to 'UTC'.

ioanpocol:
:+1:

jerome-poisson:
@naspeh UTC can be set in settings, the idea is to use local timezone by default, and if something else is needed it can be set explicitly.

superdesk/factory/default_settings.py
```diff
@@ -359,7 +360,15 @@ def env(variable, fallback_value=None):
 SUPERDESK_TESTING = (env('SUPERDESK_TESTING', 'false').lower() == 'true')
 
 # Default TimeZone
-DEFAULT_TIMEZONE = env('DEFAULT_TIMEZONE', 'Europe/Prague')
+DEFAULT_TIMEZONE = env('DEFAULT_TIMEZONE')
+
+if DEFAULT_TIMEZONE is None:
+    try:
```
naspeh: Do we really need this `try ... except`?

You still raise `ValueError`, so if another Exception appears let's just keep it...

jerome-poisson: I raise ValueError if the value is None or Empty, but this doesn't catch any other error, and this part of code must not fail. As I can't be sure of what can be raised here (and it can change with future versions), the except Exception is needed. So yes this try/except is needed.

jerome-poisson: what I can add is a log message so the exception is not totally hidden, but settings should not fail for a timezone IMHO.

naspeh:
👍 yep, thanks for answers.

superdesk/factory/default_settings.py
naspeh: yep, warning will be good.

jerome-poisson:
just added logging message, it effectively better to not hide the exception.

jerome-poisson:
@naspeh I've just talked with @petrjasek and he agree with you about the Exception, so I remove the try/except

superdesk/factory/default_settings.py
```diff
@@ -359,7 +360,12 @@ def env(variable, fallback_value=None):
...
-DEFAULT_TIMEZONE = env('DEFAULT_TIMEZONE', 'Europe/Prague')
+DEFAULT_TIMEZONE = env('DEFAULT_TIMEZONE')
+
+if DEFAULT_TIMEZONE is None:
+    DEFAULT_TIMEZONE = tzlocal.get_localzone().zone
+    if not DEFAULT_TIMEZONE:
+        raise ValueError
 
 # The number of minutes since the last update of the Mongo auth object after which it will be deleted
 SESSION_EXPIRY_MINUTES = int(env('SESSION_EXPIRY_MINUTES', 240))
```
naspeh: How about description message for `ValueError`
```
ValueError('DEFAULT_TIMEZONE is empty after tzlocal')
```

jerome-poisson:
thanks for the feedbacks, I'm waiting for the travis green light then I'll merge this.

superdesk/factory/default_settings.py
jerome-poisson: yes it makes sense now. I add it.

naspeh:
BTW: when I'm talking about a warning, I mean [warnings module from python](https://docs.python.org/3.5/library/warnings.html) it is also just logging message, but with a different meaning.

naspeh:
Thank you for changes, I like it 👍

PR #564, opened 2016-09-06
feat(publish): add `USE_TAKES` config option

petrjasek:
it will be on by default, but in case it's set to off, superdesk
won't create any takes packages when publishing.

SD-5456


ioanpocol:
:+1:

mdhaman:
:+1:

PR #565, opened 2016-09-07
fix(publish): avoid sending individual package items

petrjasek:
when using embed items config, it won't transmit these,
only the package. only enabled this when using `NO_TAKES`
config, it's kind of experimental atm.

also renamed this `USE_TAKES` to `NO_TAKES` to match client config

SD-5458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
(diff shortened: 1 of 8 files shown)


mdhaman:
:+1:

ioanpocol:
:+1:

PR #566, opened 2016-09-07
Keep only one base TestCase

naspeh:
I'll implement on this base `TestCase` some logic for improving time of test runs.

There is introduction of two new methods `setUpForChildren` and `tearDownForChildren` to rid `super().setUp()` for each test case with such methods.

Also, I sorted imports in touched files according to [Google Python Style Guide][1].

[1]: https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#Imports_formatting


petrjasek:
:+1: not sure I like this parent `setUp` magic, didn't mind this explicit invocation, what others think?

naspeh:
It really useful don't have `setUp` methods in base `TestCase`, but you usually need some general `setUp` logic there,  this is hard to introduce such logic with `unittest.TestCase`.

Django introduces `_pre_setup` in his base `TestCase` and override `__call__` method with bunch of logic, original `unittest.TestCase.__call___` has even more logic.

Also, I saw that other do like here: https://gist.github.com/twolfson/13f5f5784f67fd49b245, this was my inspiration.

But I think my solution is better than last one, and I don't want to override `__call__` method.

mdhaman:
:+1: Changes are good. @naspeh These changes will break most of the tests in superdesk-aap repo.  Would it be possible to refactor changes to superdesk-aap repo?

naspeh:
@mdhaman, yes, I'll refactor superdesk-aap repo

mdhaman:
@naspeh thanks..

naspeh:
The error was regarding Elasticsearch In Travis build:
> HTTPConnectionPool(host='localhost', port=9200): Read timed out.

So I'm merging this.

petrjasek:
just when merging - pls squash all those commits

naspeh:
yep, I'll squash them next time

PR #567, opened 2016-09-07
[NITF] [FORMATTER] added pubdata generation method

jerome-poisson:
this commit add a method to add pubdata element at the right location,
but the method doesn't actually do anything. The method can be used by
subclasses

SDNTB-286


mdhaman:
:+1:

ioanpocol:
:+1:

PR #570, opened 2016-09-08
fix(ingest): make routed items deletable

petrjasek:
previous fix worked for routed items, but not for items in packages,
so extended syncing to all items ingested

SD-5416


naspeh:
👍 looks good

ioanpocol:
:+1:

PR #571, opened 2016-09-08
Try to modify settings in Travis

naspeh:
I just test different things with Travis in this PR.


petrjasek:
fyi you can enable travis on your fork, then it's not using superdesk quota which is only 5 concurrent builds for the whole org

naspeh:
@petrjasek: thanks, I'll do this. At the night builds were working better without delays as now :)

PR #572, opened 2016-09-08
Tests: use the same app for all test cases

naspeh:
According to https://dev.sourcefabric.org/browse/SD-5460

Some changes from #571


naspeh:
Brr, nosetests works better, but I've troubles with behave tests 😠

naspeh:
There is replacement #577

PR #573, opened 2016-09-09
feat(cache): add superdesk.cache module

### LINKED ISSUE #487: Superdesk Caching
we need caching in superdesk to keep performance under load. Lets gather requirements and do some research before implementing:

### Requirements
- cache db/elastic queries
- cache file reading/parsing
- share cache between requests/celery workers
- invalidating cache by tag
- per request/session cache

### Required Backends
- Redis
- Memcached

### Libraries
- Flask-Cache
- Werkzeug cache module
- Hermes
- Beaker

## API

there should be a decorator for caching, something like:

```
from superdesk.cache import cache

@cache(ttl=600)
def do_something(foo):
    ...

class CachedService(Service):

    @cache(ttl=30, tags=['activity'])
    def get_activity(self, user_id):
        ...
```

this way we can wrap db/elastic/etc calls and cache it.

Pls let me know what other requirements/libs/etc you can think of and I will look into it and evaluate.
/cc @mugurrus @jerome-poisson @ioanpocol @mdhaman @akintolga @marwoodandrew @hlmnrmr

petrjasek:
it is a wrapper for hermes cache library. handles cache configuration using `app.config`.

fix #487 

/cc @nistormihai


tests/cache_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+foo = Foo()
+
+
+class CacheTestCase(TestCase):
+
+    def test_cache(self):
+        self.assertEqual(1, foo.count_calls())
+        self.assertEqual(1, foo.count_calls(), 'not using cache')
+        sleep(4)
```
naspeh: Please, use here smaller time 1-2sec, because of synchronous running this test case adds ~5sec and it is too much for one test case.

superdesk/cache.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @property
+    def _backend(self):
+        if not app:
+            raise RuntimeError('You can only use cache within app context.')
+
+        if not app.cache:
+            cache_url = app.config.get('CACHE_URL', '')
+            if 'redis' in cache_url or 'unix' in cache_url:
+                print('using redis cache', cache_url)
```
naspeh: Seems `logging` is better here than `print`

naspeh:
@petrjasek: could you write why do you choose `hermes` over other to #487 or here.

👍 otherwise looks good, except some minor comments

petrjasek:
so I picked hermes because:
- works on both memcached and redis
- supports tags
- seemed simple to integrate

beaker was next on my list if this wouldn't work. the others (werkzeug, flask-cache) don't support tags

naspeh:
Thanks for the clarification!

mdhaman:
:+1:

PR #574, opened 2016-09-12
fix(saved_search): Fixing the saved search endpoint to return correct next page and previous page links correctly.

mdhaman:
With `ListCursor` we don't get the next page link if count of documents returned is same as count of documents requested.


petrjasek:
:+1:

petrjasek:
:+1:

PR #577, opened 2016-09-14
Improve tests time

naspeh:
I did it, I did it!

```
## Before
 superdesk/superdesk-core  1544
 Duration: 44 min 23 sec
 Finished: about 9 hours ago

## After my changes
 naspeh-sf/superdesk-core  29
 Duration: 18 min 44 sec
 Finished: 6 minutes ago
```
**44 vs 19**

Regarding: https://dev.sourcefabric.org/browse/SD-5460

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 /data
 static/*
 upload/*
+.noseids
 .coverage
 cover
 celerybeatschedule.db*
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ before_script: >
     && echo "index.store.type: memory" | sudo tee -a /etc/elasticsearch/elasticsearch.yml
     && tail -n10 /etc/elasticsearch/elasticsearch.yml
     && sudo service elasticsearch restart
+    && sleep 1
     && curl -XPUT 'http://localhost:9200/_snapshot/backups' -d '{"type": "fs", "settings": {"location": "/tmp/es-backups"}}'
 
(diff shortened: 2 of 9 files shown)


ioanpocol:
:100: 
The new time is very good. In the end seams that the problem was that app.data.init_elastic(app) took too much time to run.
I think the same solution for both mongo and es applied to e2e tests will get the same or better results.

petrjasek:
@naspeh can you pls describe those changes a bit?

naspeh:
@ioanpocol: there are two things in general which improve the time, `app.data.init_elastic(app)` is the second one. For mongo we don't have initialisation step, we just drop it, so there probably nothing regarding the time improvement.

@petrjasek here is description:

Things which improve the time:
1. it calls `superdesk.factory.app.get_app` every test case and this function is too complicated to run it so much time, usually app initialisation is a one-time operation.
2. elasticsearch drop & init indices, so there is snapshot/restore feature in use.

### One-time app initialisation
Nosetests runs good, I fixed few tests regarding the right `setUp` and `tearDown` actions.

Behave tests had an issue, so I fully initialize app for each feature. And it also resets for apply config options.

### Next step
I want to merge it as soon as possible and then I'll work for cleaning up some moments and probably will add some comments in the code.

petrjasek:
:+1: looks good, would be good to squash some commits

naspeh:
@petrjasek: sure, I'll squash them during the merging.

PR #580, opened 2016-09-16
[NITF] [FORMATTER] HTML 2 NITF conversion

jerome-poisson:
this is a HTML 2 NITF conversion first step. It currently remove
unknown elements and attributes, and map HTML one to NITF.

This method is *NOT USED* in core at the moment, to avoid changing the
formatting all of a sudden, it will be used in NTB branch first.

A second step may be needed for a more comprehensive conversion.

SD-5475


jerome-poisson:
@petrjasek @ioanpocol sorry I have to ask you to review again, I have changed the behaviour so unknown elements are removed but their children and texts are kept and parsed.

This make the code a bit more complicated, but that's the way it is done in XHTML browsers, and it should be done like this in our case too.

PR #587, opened 2016-09-26
WIP: feature(docs): added rst files from sphinx-apidoc

thnkloud9:
first attempt to extend server documentation using sphinx-apidoc and sphinx-autodoc.  Still kind of messy but better than nothing in my opinion.  From here we should begin to update doc blocks to proper pep8 format and add missing doc blocks where necessary.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -4,6 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
+SPHINXAPIDOC  = sphinx-apidoc
 PAPER         =
 BUILDDIR      = _build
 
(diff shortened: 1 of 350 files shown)


vied12:
Do we want the documentation files inside this repository ?
Another option would be to put them on a `gh-pages` branch, then the documentation would be accessible online on http://superdesk.github.com/superdesk
There is probably a way to generate and push the doc there automatically when master changes.

I also have a question in mind. How to handle documentation for different versions (the last release and the development one) ? But do we need to support that if we start to release more often ?

petrjasek:
it's published to http://superdesk.readthedocs.io/ on update, it has some versioning support, so there will be stable one for last tag and latest for master branch.

petrjasek:
close in favour of https://github.com/superdesk/superdesk-core/pull/590

PR #590, opened 2016-09-27
Docs refactoring: add rst files from sphinx-apidoc

petrjasek:
first attempt to extend server documentation using sphinx-apidoc and sphinx-autodoc. Still kind of messy but better than nothing in my opinion. From here we should begin to update doc blocks to proper pep8 format and add missing doc blocks where necessary.

copy of https://github.com/superdesk/superdesk-core/pull/587 using core branch

idea is to add flake8-docstrings to travis so every commit will be checked for docstrings, but before we enable it, existing code must pass..

- [x] generated rst docs for all server code
- ~~[ ] make common names private - eg rename `on_create` to `_on_create` for resources to avoid documenting it all~~
- [x] fix flake8-docstrings issues for existing code
- ~~[ ] add missing doc strings~~


petrjasek:
after all relaxed the goals, to document the code properly it would also need some refactoring as we go, renaming methods and making some private, combine files etc. so we rather keep this to add links to docs and fix existing docstrings style, and we add documentation for one component at a time with its own PR.

PR #591, opened 2016-09-29
fix(crop): cast image crop vocab width/height values to int

petrjasek:
it is validated later when doing the crop, and if updated via
ui those values are stored as strings.


thnkloud9:
:thumbsup:

PR #592, opened 2016-09-29
Restore from snapshot in prepopulation step

naspeh:
- restore from snapshot in prepopulation step;
- some refactoring;


thnkloud9:
:thumbsup:

PR #595, opened 2016-10-03
docs(publish): update publishing docs

petrjasek:
- add page for each formatter/transmitter with info
- add some initial publish info to main page

SDESK-122


petrjasek:
let me know what you think, it's very early stage, but would need some feedback about the level of information you are expecting from such docs, should there be more code references or rather more general info about how things work?
ping @superdesk/superdesk-core @superdesk/superdesk-developers

petrjasek:
obsolete now, see https://github.com/superdesk/superdesk-core/pull/598

PR #597, opened 2016-10-04
fix(rewrite): Preserv the set language on update/rewrite action

ioanpocol:
Languages are defined on settings.py like this:
LANGUAGES = [{'language': 'en', 'label': 'English', 'source': True, 'destination': False},
             {'language': 'de', 'label': 'German', 'source': True, 'destination': False},
             {'language': 'nb-NO', 'label': 'Bokmål', 'source': False, 'destination': True},
             {'language': 'nn-NO', 'label': 'Nynorsk', 'source': False, 'destination': True}]

### DIFF
--- a/apps/archive/archive_rewrite.py
+++ b/apps/archive/archive_rewrite.py
@@ -125,7 +125,7 @@ class ArchiveRewriteService(Service):
         """
         rewrite = dict()
 
-        fields = ['family_id', 'event_id', 'flags']
+        fields = ['family_id', 'event_id', 'flags', 'language']
 
         if existing_item:
             # for associate an existing file as update merge subjects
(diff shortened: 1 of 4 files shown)


apps/archive/archive_rewrite.py
```diff
@@ -125,7 +125,7 @@ def _create_rewrite_article(self, original, digital, existing_item=None, desk_id
         """
         rewrite = dict()
 
-        fields = ['family_id', 'event_id', 'flags']
+        fields = ['family_id', 'event_id', 'flags', 'language']
```
petrjasek: seems like only this part is related to PR, right?

ioanpocol: Yes

PR #598, opened 2016-10-04
docs: add missing docs

petrjasek:
- [x] publishing
- [x] ingest
- [x] settings
- [x] metadata
- [ ] ...

SDESK-122

replace https://github.com/superdesk/superdesk-core/pull/595
this should be available now on http://superdesk.readthedocs.io/en/docs/ so you can read it there


petrjasek:
good point, but imo there will be more info for each of them eventually to have them them all on single page

naspeh:
Maybe we could put them to one for now and then change in future when file get bigger

petrjasek:
updated the structure to make it flat, what do you think @naspeh ?

naspeh:
Great, I like it.

Would be good to see full `toctree` on main page, that I could go fast to [transmitters.HTTPPushService](http://superdesk.readthedocs.io/en/docs/publish.html#superdesk.publish.transmitters.HTTPPushService), etc.

petrjasek:
these are not entries now, only classes, so you don't get it in toc, just like in werkzeug docs

naspeh:
Yep, it's my fault. Good! Thanks.

petrjasek:
added some info about ingest, what do you think?

naspeh:
Pretty good, I really like one topic on one page (even they'll be long)!

naspeh:
Maybe you could merge it already and create another PR for next things?

Or, does it keep in one PR for some reasons?

BTW: name of PR include only *publishing* :)

petrjasek:
yeah, will merge it at some point :wink:

superdesk/default_settings.py
```diff
@@ -369,18 +419,15 @@ def env(variable, fallback_value=None):
 if not DEFAULT_TIMEZONE:
     raise ValueError("DEFAULT_TIMEZONE is empty")
 
-# The number of minutes since the last update of the Mongo auth object after which it will be deleted
+#: The number of minutes since the last update of the Mongo auth object after which it will be deleted
 SESSION_EXPIRY_MINUTES = int(env('SESSION_EXPIRY_MINUTES', 240))
 
-# The number of minutes before content items purged
-# akin.tolga 06/01/2014: using a large value (30 days) for the time being
-CONTENT_EXPIRY_MINUTES = int(env('CONTENT_EXPIRY_MINUTES', 4320))
```
petrjasek: hey @akintolga this seems to be 3 days and not 30, so I've fixed the number, so if you want to keep it 3 days you should update settings in superdesk-aap to be sure after update

akintolga: Thanks for this @petrjasek I'll make the necessary changes if needed.

PR #600, opened 2016-10-04
test(templates): make template schedule tests use timezone

petrjasek:
it was comparing utc time with local, but utc offset would change
when daylight saving is on for timezone, so rather use local time
together with timezone and convert to utc in steps.

### DIFF
--- a/features/steps/template_steps.py
+++ b/features/steps/template_steps.py
@@ -1,4 +1,5 @@
 
+import pytz
 from datetime import datetime, timedelta
 from superdesk.tests import set_placeholder
 from steps import when, then, get_json_data, parse_date  # @UnresolvedImport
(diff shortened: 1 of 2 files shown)


features/steps/template_steps.py
```diff
@@ -19,6 +20,21 @@ def when_we_run_create_content_task(context):
 def then_next_run_is_on_monday(context, time):
     data = get_json_data(context.response)
     next_run = parse_date(data.get('next_run'))
+    fmt = '%H:%M:%S'
+
+    try:
+        # assume time is set in given time zone
+        tz = pytz.timezone(data.get('schedule').get('time_zone'))
+    except:
```
jerome-poisson: catch all except may hide errors, would be better a data['schedule']['time_zone'] and except(KeyError) IMHO

petrjasek: right, fixing

naspeh:
Build is green

PR #601, opened 2016-10-04
[SD-5508] Flag Auto Published items so that they can be identified by…

marwoodandrew:
… the formatters

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -383,6 +383,8 @@ class BasePublishService(BaseService):
 
             metadata_tobe_copied = self.takes_package_service.fields_for_creating_take.copy()
             metadata_tobe_copied.extend([PUBLISH_SCHEDULE, SCHEDULE_SETTINGS, 'byline', EMBARGO])
+            if 'auto_publish' in updates_of_take_to_be_published:
+                metadata_tobe_copied.extend(['auto_publish'])
             updated_take = original_of_take_to_be_published.copy()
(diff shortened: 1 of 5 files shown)


superdesk/publish/formatters/__init__.py
```diff
@@ -112,6 +112,23 @@ def map_html_to_xml(self, element, html):
         if len(list(element)) == 0:
             SubElement(element, 'p').text = soup.get_text()
 
+    def pass_through(self, article):
```
petrjasek: just wondering why it's on formatter, is it supposed to be used by specific formatter implementation?

marwoodandrew: It will be used by a number of formatter implimentations :- IPNews, Newscentre, ANPA and Bulletin builder. All AAP specific. I could move it to the AAP repo if you like!

The Medianet, Racing system, pagemasters and AP ingests will be set to pass through.

marwoodandrew: That's where I had it originally! I'll move it back!

marwoodandrew:
We're having a re-think of how this is implemented.

apps/publish/content/common.py
```diff
@@ -123,7 +123,7 @@ def update(self, id, updates, original):
         """
         try:
             user = get_user()
-            auto_publish = updates.pop('auto_publish', False)
+            auto_publish = updates.get('auto_publish', False)
```
mdhaman: Needs to be added to the `digital_item`.

PR #602, opened 2016-10-05
[SD-5512] - Embargo privilege

akintolga:


petrjasek:
those reported errors are not related, will fix it in another PR

PR #605, opened 2016-10-06
feat(translate): Add support for translate operation

ioanpocol:


apps/duplication/archive_translate.py
```diff
@@ -0,0 +1,76 @@
...
+from superdesk.resource import Resource
+from superdesk.services import BaseService
+from superdesk.workflow import is_workflow_state_transition_valid
+
+
+class TranslateResource(Resource):
+    endpoint_name = 'translate'
+    resource_title = endpoint_name
+
+    schema = {
```
petrjasek: imo would be better to have `archive_translate` url and put item `guid` into payload, this way we can in theory translate multiple items with single request

ioanpocol: Done

PR #606, opened 2016-10-06
Activityreports

naomiaro:


apps/activityreports/resource.py
```diff
@@ -0,0 +1,24 @@
...
+        'date': {
+            'type': 'datetime',
+            'required': True
+        },
+        'desk': Resource.rel('desks', nullable=True),
+    }
+
+    item_methods = ['GET', 'PATCH', 'PUT', 'DELETE']
+    resource_methods = ['GET', 'POST', 'DELETE']
+    privileges = {'POST': 'activityreports', 'PATCH': 'activityreports', 'DELETE': 'activityreports'}
```
mugurrus: PUT is missing from privileges

apps/activityreports/service.py
```diff
@@ -0,0 +1,5 @@
+from superdesk.services import BaseService
+
+
+class ActivityReportService(BaseService):
```
mugurrus: There's no need to create a specific service if no customization is needed

mugurrus:
You should also add tests for report edit/delete operations

mdhaman:
@naomiaro @mugurrus it looks like we are recording `publish` and `correct` activity by desk. Is this correct? 
Can you provide some documentation on this i.e. jira ticket, problem you are trying to solve, etc?

mugurrus:
Closed because the feature was reimplemented with changes to the functionality

PR #607, opened 2016-10-06
[NITF] [FORMATTER] fixed first level children parsing in html2nitf

jerome-poisson:
first level children parsing (i.e. parsing of rool element children)
were not handling unknown elements. This commit fix this.

SDNTB-307


jerome-poisson:
@petrjasek indeed, the method is not called so I have put tests were it is used (in nitf), but I'll add one, it should be there anyway

jerome-poisson:
@petrjasek test added :)

petrjasek:
:+1:

PR #609, opened 2016-10-07
live suggestions implementation

mugurrus:


apps/suggestions/service.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def _transform(self, item):
+        """
+        Transforms an item in dictionary form to plain text.
+        :param item: dict
+        :return: str
+        """
+        text = ''
+        fields = ['slugline', 'headline', 'body_html', 'body_text', 'description_text']
+        for field in fields:
```
petrjasek: imo a bit more pythonic would be: `return '\n\n'.join([item[field] for field in fields])`

ioanpocol: return '\n\n'.join([item[field] for field in fields if field in item])

apps/suggestions/service.py
```diff
@@ -0,0 +1,72 @@
...
+        keywords = self.provider.get_keywords(self._transform(item))
+        if not keywords:
+            return ElasticCursor([])
+
+        query = {
+            'query': {
+                'filtered': {
+                    'query': {
+                        'query_string': {
+                            'query': ' '.join(kwd['text'] for kwd in keywords)
```
ioanpocol: The keywords comes with a relevance factor between 0 and 1. You can use this factor on elastic search in order to improve the relevance of the returned content items.

ioanpocol:
@mugurrus  Please rebase this PR

apps/suggestions/service.py
mugurrus: In another PR

PR #611, opened 2016-10-11
fix(ingest): Fix ftp ingest

ioanpocol:
Calculate the last updated value based on last modified file timestamp
from ftp.

### DIFF
--- a/superdesk/io/commands/update_ingest.py
+++ b/superdesk/io/commands/update_ingest.py
@@ -220,7 +220,7 @@ def update_provider(self, provider, rule_set=None, routing_scheme=None):
 
         update = {LAST_UPDATED: utcnow()}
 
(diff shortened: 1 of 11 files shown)


superdesk/io/commands/update_ingest.py
```diff
@@ -226,6 +227,10 @@ def update_provider(self, provider, rule_set=None, routing_scheme=None):
             if items:
                 update[LAST_ITEM_UPDATE] = utcnow()
 
+        if CRT_LAST_UPDATED in provider:
```
petrjasek: seems more like a hack, wouldn't it be better to pass this update dict to `feeding_service.update` call and let it do any changes there if needed?

ioanpocol: Done

PR #615, opened 2016-10-12
[NITF] [FORMATTER] added an attribute to reset HTML attributes in htm…

jerome-poisson:
…l2nitf

attr_reset can take a list of HTML attributes to reset before
converting to NITF

SDNTB-311

### DIFF
--- a/superdesk/publish/formatters/nitf_formatter.py
+++ b/superdesk/publish/formatters/nitf_formatter.py
@@ -149,13 +149,13 @@ class NITFFormatter(Formatter):
         else:
             parent.text = (parent.text or '') + text
 
-    def _parse_children(self, html_elem, root_elem):
+    def _parse_children(self, html_elem, root_elem, attr_remove):
         children = list(html_elem)
         idx = 0
         while idx < len(children):
             child = children[idx]
             try:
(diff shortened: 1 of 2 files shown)


superdesk/publish/formatters/nitf_formatter.py
```diff
@@ -190,18 +190,19 @@ def _parse_children(self, html_elem, root_elem):
 
         return html_elem
 
-    def html2nitf(self, html_elem, root_elem=None):
+    def html2nitf(self, html_elem, root_elem=None, attr_reset=None):
         """Convert HTML elements to NITF compatible elements
 
         :param ET.Element: HTML to clean/transform
         :param ET.Element: None if its the root element (must be a <div>)
+        :param list attr_reset: attributes to remove if present
```
petrjasek: shouldn't it be named `attr_remove` or `attr_blacklist` then? seems like it doesn't reset it

jerome-poisson: style can be put afterwards by the NITF mapping (e.g. if there where "style='fontstyle:italic'" when <em> is found), that's why I put reset (it's set to nothing before being updated), so I think "blacklist" is not OK, "remove" may do it.

PR #616, opened 2016-10-12
[SD-5551] Changes to duplicate an item to a different desk

nrvikas:


mdhaman:
@petrjasek merged this PR but client-core PR requires more changes. It should not affect as payload from the client is always desk where the item is located.

petrjasek:
ok

PR #619, opened 2016-10-14
feat(translate): Automatically call translate macros on translate oper…

ioanpocol:
…ation

The translate macros had defined the translate related settings
from_language(optional) and to_language(mandatory).
Example:
from_languages = ['en']
to_languages = ['nn-NO', 'nb-NO']
If from_language is not defined, the from language is not checked.


ioanpocol:
done

petrjasek:
:+1:

PR #622, opened 2016-10-17
docs update

petrjasek:
- remove autogenerated docs for now - it made live preview slooow
- add initial architecture docs describing main components
- add docs contributing info - how to write docs for superdesk


petrjasek:
will make this PR from docs branch, so the changes will be visible on readthedocs

PR #625, opened 2016-10-18
added get_text and get_char_count methods to superdesk.etree

jerome-poisson:
needed for SDNTB-310

### DIFF
--- a/superdesk/etree.py
+++ b/superdesk/etree.py
@@ -21,14 +21,31 @@ def get_text_word_count(text):
     return len(text.split())
 
 
+def get_text(html):
+    """Get plain text version of HTML element
+
+    if the HTML string can't be parsed, it will be returned unchanged
+    :param html: html string to convert to plain text
+    """
+    try:
+        root = etree.fromstringlist('<doc>{0}</doc>'.format(html))
+        text = etree.tostring(root, encoding='unicode', method='text')
+        return text
+    except ParseError:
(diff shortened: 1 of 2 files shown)


superdesk/etree.py
```diff
@@ -21,14 +21,26 @@ def get_text_word_count(text):
     return len(text.split())
 
 
+def get_text(html):
+    try:
```
mdhaman: missing docstring.

marwoodandrew:
Whilst not directly related to this pull request, I noticed the other day that the following test would fail, 

        self.assertEqual(2, get_word_count('<p>one</p><p>two</p>'))

I couldn't think of an elegant solution, other than perhaps injecting spaces into the string before parsing it, so I courageously ignored it.

superdesk/etree.py
jerome-poisson: added, thanks for the review

jerome-poisson:
@marwoodandrew thanks for the review.

The issue you're mentioning would need a separate ticket in my opinion.

I think it could be fixable with a parser with a list of block elements, adding extra space when it find one, that's doable without too much complications.

jerome-poisson:
@marwoodandrew I've opened https://dev.sourcefabric.org/browse/SDESK-240 for the issue you mentioned. I'm merging the current P.R., thanks you all for review.

PR #626, opened 2016-10-18
[SD-5553] Modified saved search end-point.

nrvikas:


apps/saved_searches/saved_searches.py
```diff
@@ -199,6 +201,19 @@ def on_fetched(self, docs):
...
+    def on_delete(self, doc):
+        self._validate_user(doc['user'])
+
+    def _validate_user(self, doc_user_id, create=False):
+        session_user = get_user(required=True)
+        if create and str(session_user['_id']) != doc_user_id:
+            raise SuperdeskApiError.forbiddenError(
+                'Unauthorized to create another user\'s search.')
+
+        if str(session_user['_id']) != doc_user_id and not current_user_has_privilege('global_saved_searches') and \
```
akintolga: `current_user_has_privilege` function checks if the user is admin so we don't need to check it here again...

apps/saved_searches/saved_searches.py
```diff
@@ -102,8 +102,9 @@ def on_fetched(self, docs):
 class SavedSearchesService(BaseService):
     def on_create(self, docs):
         for doc in docs:
-            if 'user' not in doc and request:
-                doc['user'] = request.view_args.get('user')
+            if 'user' not in doc:
```
akintolga: Maybe 'user' could be removed from the resource as we can infer the user when creating a saved search for the request. Also when update and delete we wouldn't need the user. In this case we can save the user field before committing the changes to database.

features/saved_searches.feature
```diff
@@ -232,29 +253,30 @@ Feature: Saved Searches
...
         {
         "name": "basket ball",
         "filter": {"query": {"q": "basket ball", "repo": "ingest"}},
-        "description": "abc"
+        "description": "abc",
+        "user": "#CONTEXT_USER_ID#"
         }
         """
-        When we patch "/users/#users._id#/saved_searches/#saved_searches._id#"
+        When we patch "/saved_searches/#saved_searches._id#"
```
akintolga: Before this check the other user needs to be switched...

akintolga: Also this test should pass if the updated search is global and fail if not.

nrvikas: Updated. And also added other relevant test cases

apps/saved_searches/saved_searches.py
```diff
@@ -102,8 +107,7 @@ def on_fetched(self, docs):
 class SavedSearchesService(BaseService):
     def on_create(self, docs):
         for doc in docs:
-            if 'user' not in doc and request:
-                doc['user'] = request.view_args.get('user')
+            doc['user'] = get_user(required=True)['_id']
```
petrjasek: here you could use `get_user_id`

nrvikas: Yup, I have modified the file. Thanks.

PR #628, opened 2016-10-19
update bcrypt and run tests on both python 3.4 and 3.5

petrjasek:


petrjasek:
to make sure it works on both, there are those differences like in `backend_meta`

naspeh:
I just think that we support only 3.5 for now

petrjasek:
we use 3.4 in docker too, it's ubuntu 14.04

PR #630, opened 2016-10-19
fix(validate): Ignore dateline if item is corrected because dateline …

ioanpocol:
…can't be changed after the item is published


petrjasek:
would be good to add some note in the code, maybe same thing like in the PR title

ioanpocol:
done

PR #631, opened 2016-10-19
Update minimal data for clean setups [SDESK-199]

naspeh:
- we can't use empty vocabularies, so these is one fake item at least
- we can't add kill template via UI, so it also here
- we need validators for publishing, etc.


naspeh:
I keep this sample data in repo, so you can run `./manage.py app:initialize_data --sample-data` now to get desks, stages, etc.

petrjasek:
would be good to add some info into docs

naspeh:
I'll document this during "Unify deployment" task, there I need to write proper docs for sure

petrjasek:
:+1:

PR #636, opened 2016-10-25
SDESK-162 activity reports

ancafarcas:


apps/analytics/activityreports.py
```diff
@@ -0,0 +1,58 @@
...
+                                {"term": {"task.desk": str(desk)}},
+                            ]
+                        }
+                    }
+                }
+            }
+        }
+        request = ParsedRequest
+        request.args = {'source': json.dumps(query), 'repo': 'archive,published,archived,ingest'}
+        items_list = list(get_resource_service('search').get(req=request, lookup=None))
```
ancafarcas: Does anyone know what I've been missing here? When using the `search` endpoint I'm getting 0 items in the report but when I explicitly use `archive` endpoint, the report is being populated with the correct number of items.

PR #640, opened 2016-11-03
Content API Refactoring

petrjasek:
this is an initial step
- [x] moving code to superdesk-core
- [x] publishing all items to content api
- [x] generating subscribers tokens
- [x] use tokens in content api auth and filter content by subscriber
- [x] disable publish endpoints and user/client resources in content api

atm it's using jwt for auth, this might change based on further specs (like if we need to revoke tokens).
this `content_api/tokens` is the only new module, rest is more or less copy from `superdesk-content-api` repo.


apps/publish/enqueue/enqueue_service.py
```diff
@@ -212,6 +212,9 @@ def publish(self, doc, target_media_type=None):
             logger.exception('Nothing is saved to publish queue for story: {} for action: {}'.
                              format(doc[config.ID_FIELD], self.publish_type))
 
+        # publish to content api
+        get_resource_service('content_api').publish(doc, subscribers)
```
akintolga: If we don't load the content-api module in our settings we can turn this off initially I guess. Is that right?

superdesk/publish/subscribers.py
```diff
@@ -226,3 +230,8 @@ def generate_sequence_number(self, subscriber):
             max_seq_number=max_seq_number,
             min_seq_number=min_seq_number
         )
+
+    def on_fetched(self, docs):
+        for subscriber in docs['_items']:
+            if not subscriber.get('content_api_token'):
+                subscriber['content_api_token'] = generate_subscriber_token(subscriber)
```
akintolga: I think this would also be optional until content-api is enabled.

apps/publish/enqueue/enqueue_service.py
petrjasek: I've added this config `PUBLISH_TO_CONTENT_API` which you can use to turn this off

superdesk/publish/subscribers.py
petrjasek: will use the same config here

petrjasek:
I've made the setting `CONTENTAPI_ENABLED` which is `True` by default so you can turn this off with your settings.

superdesk/eve_backend.py
```diff
@@ -150,6 +150,7 @@ def create_in_search(self, endpoint_name, docs, **kwargs):
         """
         search_backend = self._lookup_backend(endpoint_name)
         if search_backend:
+            print('insert search', endpoint_name, search_backend)
```
naspeh: Is this needed?

petrjasek: not really, I was chasing that print for a while :wink:

petrjasek:
@akintolga also it requires `SECRET_KEY` now to be set so it wouldn't be enabled anyhow

PR #642, opened 2016-11-07
SDESK-311 Server tech reference structure facelift

hlmnrmr:
Adds new general structure
Adds new localtoc-based navigation for internal pages
Modifies configuration page by adding tables to list of available settings

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -5,7 +5,6 @@
  * :copyright: Copyright 2016 Sourcefabric
  * :license: See LICENSE for details.
  */
-img.floatinglogo {
-    padding: 0 0 20px 20px;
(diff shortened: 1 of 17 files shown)


petrjasek:
the only thing I'm not sure about is this settings formatting, I was testing that but it's hard to maintain within that table, also you can't link to specific config.
that's why I did it like in django https://docs.djangoproject.com/en/1.10/ref/settings/#settings and not like python eve or flask http://python-eve.org/config.html#global-configuration

hlmnrmr:
Makes sense, maintainability is not such a big deal but linking to config is. The problem for me was readability, but now that you pointed out the Django page it is more clear what was intended in the first place :-)

I will update the PR.

PR #644, opened 2016-11-07
fix(celery): set celery version <4

petrjasek:
there is a new celery version, which we need to test first before
using it in production


akintolga:
👍

PR #650, opened 2016-11-10
Cleanup tests for new LXC containers

naspeh:
- get rid `superdesk.tests.test_settings`, it outdated copy of default_settings with small changes, so I moved these settings to `superdesk.tests.update_config` (previously named `get_test_settings`)
- fixed `default_settings` a little (for example previous order for `CORE_APPS` broke the behave tests)
- update some test cases


petrjasek:
this probably broke another test

naspeh:
@petrjasek: yes, that test is written in a strange manner.

I updated description. I wanted to remove `superdesk.tests.test_settings` some time ago, I've decided to do it now. Seems all tests are passed!

So please review again.

tests/io/feeding_services/http_tests.py
```diff
@@ -70,12 +69,19 @@ def test_get_expired_token(self):
         self.assertEquals('', TestFeedingService()._get_auth_token(provider))
 
     def test_fetch_token(self):
+        # TODO: need some rewriting
```
petrjasek: those credentials we were using don't work anymore, so you can probably get rid of it

naspeh: but the test name is `test_fetch_token`, so we need some replacement for that scenario, this why I'm keeping it there

PR #656, opened 2016-11-15
fix(ci): fix code style issues after flake8 update

petrjasek:


petrjasek:
@akintolga after all I've set those upper versions too, not super strict like you have in superdesk-aap, but just to get some initial stability

akintolga:
@petrjasek Thanks for notification petr, going forward to the new version we can try this way to keep them consistent with superdesk repo.

PR #670, opened 2016-11-23
feat(schema): add schema utility classes

petrjasek:
it allows you to define schema as class rather than dict
which can be documented via autodoc.

SDESK-363

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -1,53 +1,95 @@
 
 import superdesk
+import superdesk.schema as schema
 
+from copy import deepcopy
 from superdesk.errors import SuperdeskApiError
 from apps.auth import get_user_id
 from superdesk import get_resource_service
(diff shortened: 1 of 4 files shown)


naspeh:
Maybe just use this
```
If dictionary value is not computed and human readable like this 
FRUITS = {
   "Apple": "Red and Delicious",
   # note: eating too much orange make your hands orange
   "Orange": "A lot of vitamin C"
}
say you have the above dict defined in fruit.py starting from line#15

then you can do: 
.. literalinclude:: ../path-to-file/fruit.py
   :language: python
   :lines: 15-
   :linenos:
and you will the Human readable value + comments etc right on doc
```
ref: http://stackoverflow.com/a/38286330

petrjasek:
@naspeh there will be more complex docs, with code samples etc, so that wouldn't work

PR #673, opened 2016-11-24
Use separate flag to skip "os.environ" usage in settings

naspeh:
@akintolga reported they have problems with running e2e tests locally, so let's use separate flag for such purpose.

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -29,7 +29,7 @@ except ImportError:
 
 
 def env(variable, fallback_value=None):
-    if os.environ.get('SUPERDESK_TESTING'):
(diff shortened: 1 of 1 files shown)


superdesk/default_settings.py
```diff
@@ -29,7 +29,7 @@
 
 
 def env(variable, fallback_value=None):
-    if os.environ.get('SUPERDESK_TESTING'):
+    if os.environ.get('SUPERDESK_USE_DEAFULTS'):
```
petrjasek: there is typo `s/DEAFULTS/DEFAULTS/`

naspeh:
Fixed.

PR #674, opened 2016-11-24
refactor(search): refactor search providers

petrjasek:
Change the way how these are registered, instead of creating an
endpoint and datalayer for each it only requires a single class.

There is a proxy resource that will forward the call to appropriate
instance.


petrjasek:
this is a proof of concept, while I was documenting how to create a search provider I've realized how complex it is so I've started this refactoring, might make some changes after working on the ui, will see how it goes. feedback welcome..

PR #677, opened 2016-11-30
[SDESK-404][SDESK-406] Set expiry for published items based on the PU…

mdhaman:
…BLISHED_CONTENT_EXPIRY_MINUTES setting.

### DIFF
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -8,6 +8,7 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
+import json
 import flask
 from superdesk.resource import Resource
 from superdesk.metadata.utils import extra_response_fields, item_url, aggregations, \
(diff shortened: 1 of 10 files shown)


apps/archive/archive.py
```diff
@@ -698,9 +699,30 @@ def get_expired_items(self, expiry_datetime, invalid_only=False):
...
+        count = cursor.count()
+        no_of_pages = 0
+
+        if count:
+            no_of_pages = len(range(0, count, config.MAX_EXPIRY_QUERY_LIMIT))
+            expiry = cursor[0]['expiry']
+
+        logger.info('Number of items to expire: {}, pages={}'.format(count, no_of_pages))
+
+        for page in range(0, no_of_pages):
```
petrjasek: imo it would be better to do `while` loop, it would avoid that initial `count` call that could be expensive, like:
```
while True:
    req = ParsedRequest()
    ...
    items = list(self.get_from_mongo(...))
    if not len(items):
        break
    ...
```
it iterates over all expired items anyhow

mdhaman: I have modified the logic to use `while` loop and iterating the archive collection in a batch.

PR #681, opened 2016-12-01
[SD-5472] - Marking stories for desks

akintolga:
- A new end point created to mark a story for the desks. Similar to the highlights, a story can be marked for 0..n desks. This won't change existing functionality about the actual desk the story is in.

### DIFF
--- a/apps/highlights/service.py
+++ b/apps/highlights/service.py
@@ -115,10 +115,10 @@ class MarkedForHighlightsService(BaseService):
                     publishedService.update(publishedItem['_id'], updates, publishedItem)
 
             push_notification(
-                'item:highlight',
+                'item:highlights',
                 marked=int(highlight_on),
                 item_id=item['_id'],
-                highlight_id=str(doc['highlights']))
+                mark_id=str(doc['highlights']))
 
         return ids
 
--- a/apps/marked_desks/service.py
+++ b/apps/marked_desks/service.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014, 2015 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+import json
+from superdesk import get_resource_service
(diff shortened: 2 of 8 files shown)


apps/highlights/service.py
```diff
@@ -115,10 +115,10 @@ def create(self, docs, **kwargs):
                     publishedService.update(publishedItem['_id'], updates, publishedItem)
 
             push_notification(
```
naspeh: the nose tests are failing because of this change

superdesk/metadata/item.py
```diff
@@ -374,6 +374,10 @@
         'type': 'list',
         'schema': Resource.rel('highlights', True)
     },
+    'marked_desks': {
```
petrjasek: I'm just wondering if this should be list of ids, or if it would make sense to add some more metadata to each desk (like if this was seen by someone etc), just thinking ahead so we wouldn't have to migrate such data

apps/marked_desks/service.py
```diff
@@ -0,0 +1,77 @@
...
+            else:
+                marked_desks = [h for h in marked_desks if h != doc['marked_desk']]
+                marked_desks_on = False  # highlight toggled off
+
+            updates = {
+                'marked_desks': marked_desks,
+                '_updated': item['_updated'],
+                '_etag': item['_etag']
+            }
+            service.update(item['_id'], updates, item)
```
mdhaman: Do we need to change the `ETAG`? imo, it should be `system_update`.

apps/marked_desks/service.py
```diff
@@ -0,0 +1,77 @@
...
+
+            publishedItems = published_service.find({'item_id': item['_id']})
+            for publishedItem in publishedItems:
+                if publishedItem['_current_version'] == item['_current_version'] or not marked_desks_on:
+                    updates = {
+                        'marked_desks': marked_desks,
+                        '_updated': publishedItem['_updated'],
+                        '_etag': publishedItem['_etag']
+                    }
+                    published_service.update(publishedItem['_id'], updates, publishedItem)
```
mdhaman: same as archive endpoint. imo, we need to use `system_update`.

apps/highlights/service.py
akintolga: Thanks @naspeh I have fixed those tests.

PR #688, opened 2016-12-09
[SDESK-461] Copy fields from the original article to rewritten article.

mdhaman:
@petrjasek @ioanpocol i was thinking of adding `'keywords', 'target_regions', 'target_types', 'target_subscribers'` fields to the default schema and provide then in the content profile editor. imo, `Keywords` does qualify but not sure about other fields. Pls let me know your thoughts?


petrjasek:
I agree only keywords should be in editor, and having it all in default schema makes sense

mdhaman:
Will a separate PR for schema changes.

PR #692, opened 2016-12-13
[SDESK-395] Purge the audit collection

marwoodandrew:


superdesk/audit/audit.py
```diff
@@ -0,0 +1,83 @@
...
+
+        user = getattr(g, 'user', None)
+        if not user:
+            return
+
+        audit = {
+            'user': user.get('_id'),
+            'resource': resource,
+            'action': 'updated',
+            'extra': doc
```
mdhaman: Can we add `_id` in the `doc` from `original`?

superdesk/audit/commands.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Scan the audit collection for items to delete
+        while True:
+            query = self.item_entry_query
+            query['$and'].append({'_updated': {'$lte': date_to_str(self.expiry)}})
+            if current_id:
+                query['$and'].append({'_id': {'$gt': current_id}})
+            req = ParsedRequest
+            req.sort = '[("_id", 1)]'
+            req.projection = '{"_id": 1, "extra.guid": 1, "extra._id": 1, "extra.item_id": 1, "extra.item": 1}'
```
mdhaman: We have more than 5 million documents in audit. Do we need to batch the query?

marwoodandrew: It should be possible to run it as a command, I was thinking to set the expiry to a value that would delete a sane number and run the command and repeat. 

We'll need some indexes on the audit collection. I guess we can import the production audit collection into UAT and tune it there, the fact that there would be no related items in the archive shouldn't matter.

superdesk/audit/audit.py
marwoodandrew: Done

superdesk/audit/audit.py
```diff
@@ -0,0 +1,85 @@
...
+    endpoint_name = 'audit'
+    resource_methods = ['GET']
+    item_methods = ['GET']
+    schema = {
+        'resource': {'type': 'string'},
+        'action': {'type': 'string'},
+        'extra': {'type': 'dict'},
+        'user': Resource.rel('users', False)
+    }
+    exclude = {endpoint_name, 'activity', 'dictionaries', 'macros'}
```
petrjasek: maybe there should be also `auth` for sessions, but this might be also filtered out via `g.user`

superdesk/default_settings.py
```diff
@@ -446,6 +455,9 @@ def env(variable, fallback_value=None):
 #: The number of minutes before published content items are purged
 PUBLISHED_CONTENT_EXPIRY_MINUTES = int(env('PUBLISHED_CONTENT_EXPIRY_MINUTES', 0))
 
+#: The number of minutes before audit content is purged
+AUDIT_EXPIRY_MINUTES = int(env('AUDIT_EXPIRY_MINUTES', 0))
```
petrjasek: maybe this could be enabled by default, 1 day or 7 days, not sure how big it gets?

marwoodandrew: In production we have 5 Million records consuming 20Gb after around 6 Months. It is becoming a concern as far as storage is concerned.

It has been quite useful for diagnosing production issues but the size is starting to be a problem for searching as well. 

In terms of a value for the parameter I was thinking along the lines of a couple of months. I'd deliberately defaulted it to zero so that an installation would need to opt in to have the collection purged.

superdesk/audit/audit.py
marwoodandrew: Yes session creates were filtered out, It should now keep them.

PR #695, opened 2016-12-14
feat(subscriber): add endpoint to generate tokens

petrjasek:
- it is using redis now to store auth tokens
- those can be revoked if needed

SDESK-437


content_api/tokens/__init__.py
```diff
@@ -45,35 +46,36 @@ def generate_subscriber_token(subscriber, ttl_days=365):
         payload = {'sub': str(subscriber['_id']), 'exp': _timestamp(exp)}
```
naspeh: You already `ex` in `redis.set`

> ex sets an expire flag on key name for ex seconds.

So probably you don't need this `exp` key in payload

superdesk/utils.py
```diff
@@ -74,6 +75,13 @@ def get_random_string(length=12):
     return str(uuid4())
 
 
+def get_random_token(length=40,
```
naspeh: Could we use just
```
>>> uuid.uuid4().hex
'7193d651965d499496c9e8f439e8a9ce'
```
it has 32 length, `uuid4` is usually the best way to get a random string (more secure as I know)

content_api/tokens/__init__.py
petrjasek: right

superdesk/utils.py
petrjasek: yep, updated it to even more secure version

naspeh:
BTW: Why have you decided to use Redis for this (except expiration feature)? Could we keep these tokens in Mongo, I just have been thinking more and seems it's better to keep important data in one storage and this is Mongo for us.

petrjasek:
I wanted to use different db there so that high load on content api won't affect production. But given that it can be different mongo instance this might not be an issue.
any feedback @mdhaman @marwoodandrew @akintolga ?

mdhaman:
@petrjasek it would be good to use the mongo so that we have one data store. Moreover, Redis HA would need redis-sentinel i.e. more nodes/processes to monitor. Our infrastructure would also be simple if we use mongo.

petrjasek:
so now it's using mongo. storing token as `_id` so there is no need for another index

naspeh:
Error in behave:
```
--------------------------------------------------------------------------------
FAILURE in step 'we get new resource':
  Feature:  Subscribers
  Scenario: Generate token for subscriber
Assertion Failed: 400 not found in (200, 201) : Status code not matches.Response body:"""{"_error": {"code": 400, "message": "Insertion failure: 1 document(s) contain(s) error(s)"}, "_issues": {"ttl": "unknown field"}, "_status": "ERR"}"""
Captured logging:
WARNING:superdesk:No documents for subscribers resource were deleted using lookup {}
exception: 400 not found in (200, 201) : Status code not matches.Response body:"""{"_error": {"code": 400, "message": "Insertion failure: 1 document(s) contain(s) error(s)"}, "_issues": {"ttl": "unknown field"}, "_status": "ERR"}"""
```

PR #698, opened 2016-12-16
fix(capi): store subscriber tokens in content api db

petrjasek:


superdesk/publish/subscriber_token.py
```diff
@@ -4,6 +4,7 @@
 from datetime import timedelta
 from superdesk.utc import utcnow
 from superdesk.utils import get_random_token
+from content_api import MONGO_PREFIX
```
naspeh: It's strange to see here import from `content_api`, seems we need to reorganize things regarding `superdesk`, `apps`, `content_api`

petrjasek: yep I wasn't sure how to do this, but imo this way it's kind of explicit that it manages a content api resource. any suggestions how to make this better?

naspeh: I think we could keep this for now.

But need to think about that superdesk shouldn't use imports from `conten_api` and `apps` I don't know how it could be done yet... I also think that `legal_archive` and `archived` also have these prefixes so maybe it should also exist an ability to run archive as separate WSGI application, etc. Just need to look over all this code and think.

naspeh: For example there is great ability to combine wsgi application in one application, but in production depends on requirements it could be run separately:
- http://flask.pocoo.org/docs/0.11/patterns/appdispatch/#combining-applications

for example, with this we could run `content_api` in one WSGI application with superdesk, but it needs more thinking for sure

PR #699, opened 2016-12-19
fix comments with dot in username

petrjasek:
- fix username and desk name parser to pick up these
- add a workaround for storage - it doesn't allow `.` in dict keys

SDESK-472

### DIFF
--- a/apps/comments/comments.py
+++ b/apps/comments/comments.py
@@ -42,6 +42,36 @@ class CommentsResource(Resource):
     privileges = {'POST': 'archive', 'DELETE': 'archive'}
 
 
+def encode_keys(doc, field):
+    """Encode field keys before storing.
+
+    Mongo doesn't allow `.` in dict key.
+
+    :param doc: doc
+    :param field: doc field
+    """
+    if not doc.get(field):
+        return
+    encoded = {}
+    for key, val in doc.get(field).items():
+        encoded[key.replace('.', ';')] = val
+    doc[field] = encoded
+
+
+def decode_keys(doc, field):
+    """Decode encoded field keys.
+
+    :param doc: doc
+    :param field: doc field
+    """
+    if not doc.get(field):
+        return
+    decoded = {}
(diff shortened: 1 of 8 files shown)


apps/comments/user_mentions.py
```diff
@@ -23,8 +23,8 @@ def get_mentions(text):
     Returns the names of desks and users from a given text
     User names starts with '@' and desk names starts with '#'
     """
-    user_pattern = re.compile("(^|\s)\@([a-zA-Z0-9-_.]\w+)")
-    desk_pattern = re.compile("(^|\s)\#([a-zA-Z0-9-_.]\w+)")
+    user_pattern = re.compile("(^|\s)\@([a-zA-Z0-9-_.\w]+)")
+    desk_pattern = re.compile("(^|\s)\#([a-zA-Z0-9-_.\w]+)")
```
naspeh: `\w` consists of `a-zA-Z0-9_` so pattern can be simplified like `[\w-.]+`

apps/comments/comments.py
```diff
@@ -42,6 +42,36 @@ class CommentsResource(Resource):
...
+    """
+    if not doc.get(field):
+        return
+    encoded = {}
+    for key, val in doc.get(field).items():
+        encoded[key.replace('.', ';')] = val
+    doc[field] = encoded
+
+
+def decode_keys(doc, field):
```
naspeh: So if we'll have original symbol `;` in name it'll be replaced with `.` too, should we think about this case or not?

naspeh: What do you think about to use `md5` for encoding/decoding?

petrjasek: base64 sounds good, but it makes those names very cryptic, will leave it with replace for now, we don't allow `;` in username

naspeh:
@petrjasek I've just checked, we can create desks with `;` symbols, so we need to add restrictions to desk's names or use `base64` there

PR #702, opened 2016-12-20
fix(ingest): try ingesting file multiple times

petrjasek:
it might fail when file is uploaded at the very moment,
so instead of moving it away on error it will keep it there
and try during next ingest cycles up to 10m.


naspeh:
This test failing time to time:
```
Failing scenarios:
  features/content_publish.feature:173  Publish a ingested content
```
https://travis-ci.org/superdesk/superdesk-core/jobs/185704907

PR #705, opened 2016-12-21
fix(errors): report to sentry actual error

petrjasek:
instead of superdesk errors so we get some idea what is the issue

### DIFF
--- a/superdesk/io/feed_parsers/rfc822.py
+++ b/superdesk/io/feed_parsers/rfc822.py
@@ -256,6 +256,7 @@ class EMailRFC822FeedParser(EmailFeedParser):
             soup = BeautifulSoup(html, "html.parser")
         except Exception as e:
             # special handling?
+            logger.exception(e)
             raise e
 
         # remove the doctype declaration if present
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,5 +43,5 @@ before_script: >
 
 script:
     - flake8
-    - time nosetests -v --with-timer --with-id
+    - time nosetests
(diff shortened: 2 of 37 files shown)


naspeh:
I think it's better to use `logging` module for that:
- https://docs.sentry.io/clients/python/integrations/logging/

in our codebase we should use `logger.exception()` instead of `app.sentry.captureException()`

with this I can configure logging to file, to sentry, to logstash, etc...

petrjasek:
good point, will update

superdesk/io/feed_parsers/rfc822.py
```diff
@@ -132,7 +133,8 @@ def parse(self, data, provider=None):
                                 html_body = self.safe_html(html_body)
                                 continue
                             except Exception as ex:
-                                logger.exception(
+                                logger.exception()
+                                logger.error(
```
naspeh: Maybe just one error logging with `exc_info=True`, like:
```
# If you're actually catching an exception, use `exc_info=True`
logger.error('There was an error, with a stacktrace!', exc_info=True)

```

naspeh:
Sorry, seems you need `logger.exception(ex)` everywhere

petrjasek:
yep makes sense, working on it

naspeh:
@petrjasek thanks! :+1:

mdhaman:
in most cases, we are doing `logger.exception(ex)` and  then raising some Superdesk defined errors, can we move logger.exception(ex) into Superdesk Errors.

petrjasek:
@mdhaman updated it to log exceptions for publish/ingest/internal errors

PR #709, opened 2016-12-29
[SDESK-530] fix(search_providers):Marking the search provider as used when item is fetched.

mdhaman:


mdhaman:
I have merged this PR. @petrjasek this PR needs to be cherry picked into 1.5 branch.

PR #713, opened 2016-12-30
[SDESK-119] password-less authentication

jerome-poisson:
implementation of secure login (aka password-less authentication).

please check https://wiki.sourcefabric.org/display/NR/How+to+use+secure+login

### DIFF
--- a/apps/auth/service.py
+++ b/apps/auth/service.py
@@ -26,7 +26,10 @@ class AuthService(BaseService):
     def on_create(self, docs):
         for doc in docs:
             user = self.authenticate(doc)
-
+            if 'is_enabled' in user and not user.get('is_enabled', False):
+                raise UserDisabledError()
+            if not user.get('is_active', False):
+                raise UserInactiveError()
             self.set_auth_default(doc, user['_id'])
 
     def on_created(self, docs):
@@ -46,14 +49,6 @@ class AuthService(BaseService):
         doc['user'] = user_id
         doc['token'] = utils.get_random_string(40)
         del doc['password']
-
-    def check_user(self, user):
-        """Check if the user is enabled and active"""
-        if 'is_enabled' in user and not user.get('is_enabled', False):
-            raise UserDisabledError()
-
-        if not user.get('is_active', False):
-            raise UserInactiveError()
 
 
 class UserSessionClearService(BaseService):
--- a/apps/auth/service.py
+++ b/apps/auth/service.py
@@ -14,6 +14,8 @@ from superdesk import utils as utils, get_resource_service
 from superdesk.services import BaseService
 from eve.utils import config
 from superdesk.errors import SuperdeskApiError
+from apps.auth.errors import UserDisabledError
+from superdesk.users.errors import UserInactiveError
 
 
 class AuthService(BaseService):
@@ -44,6 +46,14 @@ class AuthService(BaseService):
         doc['user'] = user_id
         doc['token'] = utils.get_random_string(40)
         del doc['password']
+
+    def check_user(self, user):
+        """Check if the user is enabled and active"""
+        if 'is_enabled' in user and not user.get('is_enabled', False):
+            raise UserDisabledError()
+
+        if not user.get('is_active', False):
+            raise UserInactiveError()
 
 
 class UserSessionClearService(BaseService):
--- a/apps/auth/xmpp/auth.py
+++ b/apps/auth/xmpp/auth.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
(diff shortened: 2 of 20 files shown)


petrjasek:
imo it would be better if there is single resource handling authentication with different types/methods, with some specific adapters to implement authentication. this might be more obvious once there are docs in place.

jerome-poisson:
@mdhaman I'll do testing, need to do some setup before (openldap etc).

@petrjasek I have considered using a single endpoint/resource, but I've chosen this option with 1 common endpoint + one per auth method for different reasons:

- minimal changes in backend, the most annoying one is the change of database endpoint (auth => auth_db), but beside that most of code is untouched
- we can have specific schema validation with different resource endpoints
- in client changes are also minimal, and we have easy way to call different auth (change endpoint and use specific API). For logout, we can still use the common endpoint (auth).
- for replacing auth method like LDAP, there is basically nothing to do (beside changing the module to import), and nothing in client neither. We can simulate this behaviour with a one API for all with some kind of default value though

globally a one endpoint for all is a good way, but it need to change a lot of things in code + tests, I'm not sure if it worth it. We can discuss it in the scrum.

Also there is a egg and chicken problem with merging: merge need to be done in backend to validate tests in client, but this will break all superdesk during the change. It's probably more secure to use a branch during the merge then move everything to master when tests are validated. An other point we can discuss in scrum.

mdhaman:
@jerome-poisson/ @petrjasek agree about having a separate branch for this feature. I can do some testing with our AD at our end.

apps/auth/xmpp/auth.py
```diff
@@ -0,0 +1,84 @@
...
+            raise SuperdeskApiError.notConfiguredError()
+        domain = app.config['XMPP_AUTH_DOMAIN']
+        jid = credentials.get('jid')
+        if not jid:
+            raise CredentialsAuthError(credentials)
+        user = get_resource_service('auth_users').find_one(req=None, jid=jid)
+        if not user:
+            raise CredentialsAuthError(credentials)
+
+        if 'is_enabled' in user and not user.get('is_enabled', False):
```
petrjasek: I think this enabled and active handling should be done in base service, and only after user authenticates

apps/auth/xmpp/auth.py
```diff
@@ -0,0 +1,84 @@
...
+            'type': 'string',
+            'required': True
+        },
+        'token': {
+            'type': 'string'
+        },
+        'user': Resource.rel('users', True)
+    }
+    resource_methods = ['POST']
+    item_methods = ['GET', 'DELETE']
```
petrjasek: imo there should be no item methods, only `POST` to create new tokens, tokens are fetched/removed via `auth` resource right?

jerome-poisson: they can be removed by both endpoints (auth and auth_db). I've kept DELETE because it's used in at least one test: in https://github.com/superdesk/superdesk-core/pull/713/files#diff-eea4291faf24535b36481f7c501fbb1eL202 the _remove latest_ reference auth_db. That's the main reason, but the test can maybe be rewritten and then we can delete imte_methods here.

apps/auth/xmpp/auth.py
jerome-poisson: after user authenticates? This means that user would enter credentials even if it can't work? Agree that this should be in BaseService as it is a common check. I've put it a AuthService method, do you think it should be somewhere else?

mdhaman:
@jerome-poisson I have made some change and done a PR against your repo https://github.com/jerome-poisson/superdesk-core/pull/1.

jerome-poisson:
@mdhaman thanks, the tests are failing on your P.R., and there is conflict. Can you rebase on the last version ?

jerome-poisson:
@mdhaman great thanks! So you have tested LDAP auth with your settings right? It is OK on this point or I need to do further testing? Just to be sure before merging.

apps/auth/xmpp/auth.py
petrjasek: I would rather remove it then, and also don't think DELETE should be public method, only POST

apps/auth/service.py
```diff
@@ -45,6 +47,14 @@ def set_auth_default(self, doc, user_id):
         doc['token'] = utils.get_random_string(40)
         del doc['password']
 
+    def check_user(self, user):
```
petrjasek: imo this should be part of `on_create` and not being called from every implementation,
authenticate should just check authentication and return a user, no matter if such
user is enabled etc. so it wouldn't do any authorization, just authentication

apps/auth/xmpp/auth.py
jerome-poisson: @petrjasek that's legacy code, I'll remove it and check if nothing is broken.

jerome-poisson:
@mdhaman yes, I was actually thinking about adding a check in this case, and more generally if the XMPP auth URL is not set.

apps/auth/service.py
jerome-poisson: OK, I've mostly kept legacy code, I'll move it.

jerome-poisson:
@petrjasek is it alright now? Can I merge?

@mdhaman the link is not shown anymore if secure login is not activated and configured.

PR #716, opened 2017-01-04
fix(sign_off): Update sign off if mapped field is updated on user

ioanpocol:


superdesk/users/services.py
```diff
@@ -107,6 +107,16 @@ def set_sign_off(user):
             user[SIGN_OFF] = '{first_name[0]}{last_name[0]}'.format(**user)
 
 
+def update_sign_off(updates):
+    """
+    Update sign_off property on user if the mapped field is changed.
+    """
+
+    signOffMapping = app.config.get('SIGN_OFF_MAPPING', None)
```
petrjasek: this var should be `sign_off_mapping`

ioanpocol:
fixed

PR #718, opened 2017-01-05
[doc] authentication documentation

jerome-poisson:
- authentication mechanism documentation
- added plantuml into Sphinx configuration and sphinxcontrib-plantuml in requirements, now ..uml syntax can be used (check http://plantuml.com/ and https://pypi.python.org/pypi/sphinxcontrib-plantuml)
- imports have been moved in search_providers/__init__.py to avoid error during Sphinx build

SDESK-551

### DIFF
--- a/apps/auth/__init__.py
+++ b/apps/auth/__init__.py
@@ -63,7 +63,7 @@ def get_user(required=False):
 def get_user_id(required=False):
     """Get authenticated user id.
 
-    :param boolean requred: if True and there is no user it will raise an error
+    :param boolean required: if True and there is no user it will raise an error
     """
     user = get_user(required)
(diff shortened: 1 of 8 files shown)


jerome-poisson:
This P.R. should not be merged before:

- sphinx is available on readthedocs (not really mandatory, but the diagram would not be visible without it)
- password-less authentication is merged, as this doc is based on the changes there: https://github.com/superdesk/superdesk-core/pull/713

jerome-poisson:
password-less authentication has been merged, but the P.R. integrating plantuml in docker image is not merged yet. I'm merging this P.R. anyway, schema will just not be visible on readthedocs until  rtfd/readthedocs-docker-images#17 is merged.

PR #721, opened 2017-01-09
fix(renditions): keep renditions until expiry

petrjasek:
instead of removing old renditions on update, it will remove
these for all versions when item expires.

SDESK-429

### DIFF
--- a/apps/archive/__init__.py
+++ b/apps/archive/__init__.py
@@ -16,8 +16,10 @@ import logging
 import superdesk
 from superdesk.celery_app import celery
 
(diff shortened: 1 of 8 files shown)


mdhaman:
@petrjasek but with this approach we would have picture items not expiring from production.

petrjasek:
@mdhaman exactly. still figuring out the workflow for those

petrjasek:
@mdhaman removed that embedded handling

mdhaman:
@thanks petr

PR #724, opened 2017-01-10
fix(rendition): Fixing the issue if rendition is published then rendition should not be deleted.

mdhaman:
@petrjasek: I have take a different approach using a separate collection to keep track of media ids that are published for following reasons:
- To move/delete the picture item from production once it expires.
- Avoid querying multiple collection to find if the media is published. 

**Things to do:**
- [x] Handle spike of the picture item
- [x] Behave tests

https://github.com/superdesk/superdesk-core/pull/721


mdhaman:
@petrjasek can you please review

superdesk/media/media_references.py
```diff
@@ -0,0 +1,44 @@
...
+        'published': {
+            'type': 'boolean'
+        }
+    }
+    endpoint_name = 'media_references'
+    internal_resource = True
+    resource_methods = []
+    item_methods = []
+    mongo_indexes = {
+        'item_id_1_media_1': ([('item_id', 1), ('media_id', 1)], {'unique': True}),
```
petrjasek: not sure if this should be unique, I guess it wouldn't be common use case to have single picture used more times in an article, but also not sure if we should prevent it. it works fine with only featured, but we can have more placeholders later

apps/archive/common.py
```diff
@@ -396,41 +399,32 @@ def remove_unwanted(doc):
...
+    for rendition in renditions.values():
+        media = rendition.get('media') if isinstance(rendition.get('media'), str) else str(rendition.get('media'))
+        try:
+            references = get_resource_service('media_references').get(req=None, lookup={
+                'media_id': media, 'published': True
+            })
+
+            if references.count() == 0:
+                logger.info('Deleting media:{}'.format(rendition.get('media')))
+                app.media.delete(media)
```
petrjasek: I think now it could also remove `media_references` for that media file with `published = False`

superdesk/media/media_references.py
mdhaman: agree. I have removed the unique_id as well as `featuremedia` references.

apps/archive/common.py
mdhaman: If media is not published via any of its item it will remove the delete the media file.

mdhaman:
@petrjasek can you please review again.

petrjasek:
:+1:

PR #730, opened 2017-01-12
fix(users): make activate user token live for 7 days

petrjasek:
- fix `store_reset_password_token` to use ttl from args not config

SDESK-509

### DIFF
--- a/apps/auth/db/reset_password.py
+++ b/apps/auth/db/reset_password.py
@@ -79,11 +79,10 @@ class ResetPasswordService(BaseService):
             raise SuperdeskApiError.badRequestError('Either key:password or email must be provided')
 
     def store_reset_password_token(self, doc, email, days_alive, user_id):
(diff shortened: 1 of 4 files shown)


apps/auth/db/reset_password.py
```diff
@@ -78,12 +78,12 @@ def create(self, docs, **kwargs):
 
             raise SuperdeskApiError.badRequestError('Either key:password or email must be provided')
 
-    def store_reset_password_token(self, doc, email, days_alive, user_id):
-        token_ttl = app.config['RESET_PASSWORD_TOKEN_TIME_TO_LIVE']
-        now = utcnow()
+    def store_reset_password_token(self, doc, email, days_alive, user_id, now=None):
+        if not now:
```
naspeh: If this parameter is not used in code, only for test, then probably it's better to just use `unittest.mock.patch` in test,
something like
```
now = utcnow()
with patch('apps.auth.db.reset_password.utcnow', return_value=now):
...
'_created': now,
...
```

PR #732, opened 2017-01-13
fix(content_type): Add feature media to the list of supported fields

ioanpocol:


ioanpocol:
No, On master I have to do a different PR

PR #734, opened 2017-01-17
fix(content types): Delete media field and keep only feature media

ioanpocol:


petrjasek:
@ioanpocol is this also on master?

ioanpocol:
@petrjasek yes

PR #736, opened 2017-01-17
fix(validate): ignore enabled field in schema

petrjasek:
SDESK-579


petrjasek:
@ioanpocol not sure if this is fixed on 1.4? can you publish using some newly created content type?

PR #742, opened 2017-01-18
feat(ingest): store response data that could not be parsed

petrjasek:
so far added this to rss, we should have a file for ftp and file
ingest.

SDESK-471


petrjasek:
plus it will send that data to sentry

PR #752, opened 2017-01-24
fix issues reported on sentry

petrjasek:


naspeh:
hm, there is no Travis build somehow...

superdesk/io/commands/remove_expired_content.py
```diff
@@ -58,9 +58,12 @@ def remove_expired_data(provider):
     logger.info('Removing expired content for provider: %s' % provider.get('_id', 'Detached items'))
 
     if provider.get('feeding_service', None):
```
naspeh: Probably it is not needed anymore if you are trying catch a `KeyError` exception

petrjasek: yep

PR #754, opened 2017-01-25
fix(content_profile): Fix mandatory_in_list on content profile save

ioanpocol:


petrjasek:
what's the severity here? something that should go to 1.6?

ioanpocol:
For NTB on publish request genre and place. Yes, add on 1.6. I will add also on 1.4.

PR #755, opened 2017-01-25
Analytics - item process duration

ancafarcas:


ancafarcas:
All the changes will be done in the `superdesk-analytics` repo from now on.

PR #760, opened 2017-01-26
fix(content types): Fix disabling subject on content type editing

ioanpocol:


ioanpocol:
I will add the test in a different PR.

ioanpocol:
and seams that a fix too ...

ioanpocol:
here is the PR:
https://github.com/superdesk/superdesk-core/pull/762

PR #767, opened 2017-01-27
fix(validators): set dateline as not required for picture kill

petrjasek:
it's not required for publish or correct, but if it's not provided
there is no way to add it when you're killing an item.

SDESK-650

### DIFF
--- a/apps/prepopulate/data_sample/validators.json
+++ b/apps/prepopulate/data_sample/validators.json
@@ -511,11 +511,7 @@
             },
             "renditions": {
                 "type": "dict",
-                "required": true,
-                "schema": {
-                    "4-3": {"type": "dict", "required": true},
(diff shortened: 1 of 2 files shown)


apps/prepopulate/data_sample/validators.json
```diff
@@ -601,11 +597,7 @@
             },
             "renditions": {
                 "type": "dict",
-                "required": false,
-                "schema": {
-                    "4-3": {"type": "dict", "required": false},
-                    "16-9": {"type": "dict", "required": false}
-                }
+                "required": false
```
naspeh: it also update `renditions`, is it ok or just occasionally?

petrjasek: yep that's by design to sync it with init data

PR #769, opened 2017-01-27
fix(content_types): Fix edit content types

ioanpocol:
SDESK-647
Subject field can't be saved in Content Profiles management
The changes from this master based PR were missing: https://github.com/superdesk/superdesk-core/pull/760

SDNTB-370
Cannot publish on ntb.dev due to validation error
The required for subject was not saved correctly

SDNTB-372
NTB: Content profiles eiditing: When Place is required, saving fails
The required for place was readed from subject schema not from place schema

SDESK-652
Service and Subject values saved in item created from content profile are not preserved in resulting package
Set the scheme for default value.

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -274,8 +274,9 @@ def set_enabled_for_custom(editor, allowed, fieldsMap):
 
 def set_required_for_custom(editor, schema, mandatory, fieldsMap):
     for field, value in mandatory.items():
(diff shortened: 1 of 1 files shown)


ioanpocol:
Tests will be added in the next PR when the set default issue will be fixed

petrjasek:
can you add more info or those tests? it's hard to see how this fixes eg "cannot publish" issue

ioanpocol:
I added more info. I will add also the tests in next PR.

PR #771, opened 2017-01-27
SDESK-588 Backend related changes

Anca21:


apps/analytics/processed_items_report.py
```diff
@@ -0,0 +1,79 @@
...
+    def count_items(self, query, state, starting, ending):
+        """
+        Return the number of items which was created in a given time interval and having a certain state.
+        If no state is give, it will return all items created in the given time interval.
+        """
+        my_list = self.get_items(query)
+        if state == '':
+            items = sum(1 for i in my_list
+                        if (i['state'] in ['published', 'spiked', 'corrected', 'killed', 'updated']) and
+                        str(i['firstcreated']) >= str(starting) and str(i['firstcreated']) <= str(ending))
```
mugurrus: You should use the _updated field instead of firstcreated. _updated is set for each item version, firstcreated doesn't change. We are only interested in the version timestamp.

Anca21: I changed that now, along with other aspects.

PR #772, opened 2017-01-27
fix(nose): reset config after no takes tests

petrjasek:


tests/enqueue_test.py
```diff
@@ -22,10 +22,14 @@ def setUp(self):
         ])
 
         self.service = EnqueueService()
-
-    def test_resend_no_takes(self):
         self.app.config['NO_TAKES'] = True
 
+    def tearDown(self):
+        self.app.config['NO_TAKES'] = False
```
naspeh: it's better to use `unittest.mock.patch.dict` to modify `app.config`

PR #775, opened 2017-01-30
[SDESK-258] - Archived collection curation functions

akintolga:
- [x] Delete item(s) from archived collection
- [x] Update item(s)


mdhaman:
@petrjasek can we add this to 1.6 milestone.

PR #776, opened 2017-01-30
SDESK-678 adding range for priority and urgency fields

ancafarcas:


ancafarcas:
The changes regarding data analytics will be made in the new repo `superdesk-analytics`

PR #780, opened 2017-02-01
support elastic2

petrjasek:
seems like we can support 1 and 2 with same codebase, but for version 5 we would have to get rid of
deprecated queries like `and`, `or` and `filtered`.

requires https://github.com/petrjasek/eve-elastic/pull/73

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,47 +1,61 @@
 language: python
 
 sudo: required
-
-python:
-    - "3.4"
-    - "3.5"
+dist: trusty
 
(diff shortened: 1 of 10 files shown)


jerome-poisson:
Hey @petrjasek I'm surprised to see it's working without changing requests as in superdesk/superdesk-core#326 . It's removed only in E.S. 5 ?

petrjasek:
@jerome-poisson yep those are deprecated but still working in 2.x

PR #784, opened 2017-02-02
[sdesk-664] Content Filter to block content from the API

marwoodandrew:


marwoodandrew:
Depends on superdesk/superdesk-client-core#1307

PR #793, opened 2017-02-08
[SDESK-706] - Adding missing history details to items

akintolga:
- Creating a new version in following scenarios:
  - when a story is rewritten
  - when there's a new take created
  - when an existing story is linked as a rewrite or take
  - when a story is unlinked (for takes and rewrites)


akintolga:
@mdhaman This increases the number of versions by 1 or 2 depending on the scenario. The changes in the behave tests will show those figures.

PR #798, opened 2017-02-10
[SDESK-494] Export feature and addition of Text Formatter

nrvikas:
Refactored Preview-Format and Export functionality. Hence, needs another round of review.

### DIFF
--- a/apps/export/service.py
+++ b/apps/export/service.py
@@ -1,0 +1,69 @@
+
+import logging
+from superdesk.services import BaseService
+from superdesk import get_resource_service
+from superdesk.errors import SuperdeskApiError
+from superdesk.publish.formatters import get_all_formatters
+from eve.validation import ValidationError
+from io import BytesIO
+from zipfile import ZipFile
+from flask import current_app as app
+
+logger = logging.getLogger(__name__)
+
+
+class ExportService(BaseService):
+    def create(self, docs, **kwargs):
+        doc = docs[0]
+        formatter = self._validate_and_get_formatter(doc)
+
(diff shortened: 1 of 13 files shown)


apps/export/service.py
```diff
@@ -0,0 +1,86 @@
...
+from superdesk.errors import SuperdeskApiError
+from superdesk.publish.formatters import get_all_formatters
+from eve.validation import ValidationError
+from io import BytesIO
+from zipfile import ZipFile
+from flask import current_app as app
+
+logger = logging.getLogger(__name__)
+
+FORMATTER_FORMAT_TYPES = {'AAPIpNewsFormatter': 'AAP IPNEWS', 'NINJSFormatter': 'ninjs'}
```
mdhaman: Do we need `site-specific` formatters over here?

nrvikas: Tks Mayur. I have added these attributes(format_type and output_field) to these specific formatters directly.

apps/export/service.py
```diff
@@ -0,0 +1,81 @@
...
+                        except:
+                            unsuccessful_exports += 1
+                    else:
+                        unsuccessful_exports += 1
+
+            url = None
+            # Store the zip file on media_storage
+            # only if at least one item is formatted successfully
+            if unsuccessful_exports < len(doc.get('item_ids')):
+                zip_id = app.media.put(in_memory_zip.getvalue(), filename='export.zip', content_type='application/zip')
```
ioanpocol: I think that we need to add a celery task to clean those files after a given period of time.

PR #799, opened 2017-02-10
added wufoo service provider and feed parser

jerome-poisson:
Wufoo is form building service, this commit add feeding service and
feed parser using its API.

this also fix a bug in superdesk/io/registry.py forbidding using a
feeding service and a feed parser with the same name.

It also add a last_ingested_id field for feeding services which can be
used to only ingest new items, this is more resilient than using a date.

Fixed documentation for register_feeding_service and register_feed_parser

SDNTB-377

### DIFF
--- a/apps/io/feeding_services/__init__.py
+++ b/apps/io/feeding_services/__init__.py
@@ -10,3 +10,4 @@
 
 
 import apps.io.feeding_services.reuters  # NOQA
+import apps.io.feeding_services.wufoo  # NOQA
--- a/apps/io/feeding_services/wufoo.py
+++ b/apps/io/feeding_services/wufoo.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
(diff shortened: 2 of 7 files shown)


ioanpocol:
There are a few NTB related hardcoded values and I think we have to move it on NTB repo or to add additional settings in order to map those values to current vocabulary.

petrjasek:
this should be added to docs, there is a list of parsers/services

petrjasek:
and tests pls

jerome-poisson:
@ioanpocol actually the whole parser is related to NTB, it make more sense to move it there indeed, I can keep the feeding service here.

@petrjasek didn't knew parser were in docs, will add it. And for tests the thing is that it's external service, but I can still add unit testing with mocks.

thanks for reviewing.

petrjasek:
sure, parser tests should be enough

jerome-poisson:
@ioanpocol I've moved the parser to https://github.com/superdesk/superdesk-ntb/pull/99
@petrjasek I've added (and fixed) doc for feedind_service. I've also added the test which is in the above-mentioned NTB P.R.

jerome-poisson:
@ioanpocol @petrjasek can you validate the review ? I need it to merge this P.R. (according to @karelpetrak I can merge now)

PR #804, opened 2017-02-15
[sdesk-742] keep audit of items/assets retrieved from the API

marwoodandrew:


petrjasek:
:+1:
index should work with eve 0.7.2 (https://github.com/nicolaiarocci/eve/pull/990)

marwoodandrew:
@petrjasek Can we merge this before the new version of eve, it still works, just doesn't create the index?

petrjasek:
yep

PR #805, opened 2017-02-15
[ingestion] webhooks handling

jerome-poisson:
this commit add a basic handling of webhooks.
For now settings is done on backend using FEEDING_WEBHOOKS where key is
provider name and value is endpoint suffix.

when adding a webhook, e.g. with FEEDING_WEBHOOKS = {'Personalia':
'wufoo'}, a webhook_[endpoint_suffix] will be used (the prefix help to
prevent conflicts), so here webhook_wufoo will be used.

When an GET or POST request is done on webhook_wufoo, this will trigger
the Personalia provider update.

This is a first step, improvments like client-side configuration and
new option in update schedule could follow.

This commit also add handling of transparent_schema_rules in Resource.

SDNTB-378

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -339,6 +339,7 @@ CORE_APPS.extend([
     'superdesk.io',
     'superdesk.io.feeding_services',
     'superdesk.io.feed_parsers',
+    'superdesk.io.webhooks',
     'superdesk.io.subjectcodes',
     'superdesk.io.iptc',
     'apps.io',
--- a/superdesk/io/webhooks/__init__.py
+++ b/superdesk/io/webhooks/__init__.py
@@ -1,0 +1,77 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013 - 2017 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+from superdesk.resource import Resource
+from superdesk.services import BaseService
+from superdesk.io.commands import update_ingest
+from flask import request, abort
+import os
+import superdesk
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class FeedinServiceWebhookAuth(object):
+
+    def authenticate(self):
+        abort(403, description='You are not authorized to access this resource')
+
+    def authorized(self, allowed_roles, resource, method):
+        try:
+            # provider's id may be used in the future
+            provider_name = request.args['provider_name']
+        except KeyError:
+            logger.warning('Got an invalid webhook request (missing provider name) ')
+            return False
+        # FIXME: there can be a conflict if 2 provider name are similar but with different case
+        #        this will be fixed when we'll use id instead of name
+        env_name = 'WEBHOOK_{}_AUTH'.format(provider_name.upper())
+        try:
+            auth_key = os.environ[env_name]
+        except KeyError:
+            logger.error('{} environment variable is not set'.format(env_name))
+            return False
+        return request.args.get('auth') == auth_key
+
+
+class FeedingServiceWebhookResource(Resource):
+    resource_methods = ['POST']
+    authentication = FeedinServiceWebhookAuth
+    # we don't want schema validation to accept any webhook data
+    allow_unknown = True
+
+
+class FeedingServiceWebhookService(BaseService):
+    """Service givin metadata on backend itself"""
+
+    def create(self, docs, **kwargs):
+        # we don't want to create anything
+        # we just use this service to trigget the provider
+        # and return a fake id
+        self.trigger_provider()
+        return [0]
+
+    def trigger_provider(self):
+        provider_name = request.args['provider_name']
+        lookup = {'name': provider_name}
+        for provider in superdesk.get_resource_service('ingest_providers').get(req=None, lookup=lookup):
+            kwargs = {
+                'provider': provider,
+                'rule_set': update_ingest.get_provider_rule_set(provider),
+                'routing_scheme': update_ingest.get_provider_routing_scheme(provider)
+            }
+            update_ingest.update_provider.apply_async(
+                expires=update_ingest.get_task_ttl(provider), kwargs=kwargs)
+
+
+def init_app(app):
+    service = FeedingServiceWebhookService()
+    FeedingServiceWebhookResource("webhook", app=app, service=service)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -53,7 +53,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'Superdesk'
-copyright = '2016, Sourcefabric'
+copyright = '2017, Sourcefabric'
 author = 'Sourcefabric'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/docs/ingest.rst
+++ b/docs/ingest.rst
@@ -88,3 +88,15 @@ Add new Parser
 ^^^^^^^^^^^^^^
 
 .. autofunction:: superdesk.io.registry.register_feed_parser
+
+Add a Webhook
+^^^^^^^^^^^^^
+
+Webhook are a way to trigger ingestion without polling an ingest provider: the service do a POST HTTP request on a given URL to trigger the ingestion, resulting in resources saving and quicker ingestion.
+Webhooks are using ``webhook`` endpoint. The service triggering the webhook must use this endpoint with 2 URLs parameters:
+
+* ``provider_name`` which is the name of the provider to trigger
+* ``auth`` which is an authentication key. This key is set in a ``WEBHOOK_[PROVIDER_NAME]_AUTH`` environment variable, when ``[PROVIDER_NAME]`` is the name of the provider in uppercase.
+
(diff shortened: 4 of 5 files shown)


jerome-poisson:
This P.R. hasn't been tested with Wufoo yet, it needs https://github.com/superdesk/superdesk-core/pull/799 to be merged, and a publicly available URL.

ioanpocol:
please rebase

superdesk/default_settings.py
```diff
@@ -571,5 +572,12 @@ def celery_queue(name):
 XMPP_AUTH_DOMAIN = env('XMPP_AUTH_DOMAIN', 'Superdesk')
 
 
+#: webhooks (push system for feeding services) 
+#: dict were key is provider name and value is webhook endpoint
+#: "webhook_" will always be prefixed to endpoint name to avoid name conflicts
+#: e.g. {'wufoo NTB': 'wufoo'} will create webhook_wufoo endpoint
+FEEDING_WEBHOOKS = {}
```
petrjasek: imo we should introduce an api for this, like we have to register ingest providers, outputs etc.

apps/io/feeding_service_webhook.py
```diff
@@ -0,0 +1,58 @@
...
+
+from superdesk.resource import Resource
+from superdesk.services import BaseService
+from superdesk.io.commands import update_ingest
+from superdesk.io.registry import registered_webhooks
+import superdesk
+
+
+class FeedingServiceWebhookResource(Resource):
+    public_methods = ['GET', 'POST']
```
petrjasek: there is missing authentication imo. POST shouldn't be public, and not sure if there is a usecase for GET requests?

jerome-poisson: @petrjasek this endpoint is there for being called by external services, it's just a generic public URL which will trigger the feeding service every time a POST or GET request is made there. It's public and without authentication on purpose.

petrjasek: which is imo wrong, because that makes it vulnerable to dos attacks

jerome-poisson: @petrjasek URL must be generic and is called by third party services, we can't use authentication there because we have no control on the entity doing the HTTP request. In the case of Wufoo, they have a handshake parameter which we could use, but this would be vendor specific.

Furthermore, the risk is limited, as the attacker need to know an URL, and with it he can only trigger a ingestion, there are probably more efficient ways to attack a superdesk instance.

We can add protections like limiting the number of webhook triggers (no more than 1 every X minutes for instance), or restrict to a specific request IP.

This P.R. only add a simple way to trigger ingestion with webhook, but if the security is a concern here, we should discuss it more deeply to decide how to implement the whole thing.

The POST doesn't actually store anything, there is no backend attached to the service.

petrjasek: if we control the url, we can provide auth tokens for it right? or at least add some limit.
besides this one there is only auth endpoint which is public and that one requires credentials, so not sure what you mean by more efficient ways.

naspeh: usually it's good practice to have some kind of  `Signature` for example: https://developer.github.com/webhooks/#delivery-headers
```
X-Hub-Signature	HMAC hex digest of the payload, using the hook's secret as the key (if configured).
```
that you can identify that the posted request is part that has secret key...

I don't understand why Wufoo don't use it, this `a handshake parameter` I think is just `GET` parameter like secret, but the secret shouldn't be transferred over internet...

so in case of Wufoo probably IP restriction should be used or so.

naspeh: or maybe really use this `a handshake parameter` but HTTPS should be used for sure in this case...

petrjasek: imo we can't assume what this entity making post requests will implement, but we control the url..

jerome-poisson: @petrjasek I'm not sure what you mean with auth token, the requesting entity (wufoo) will not provide it. By "more efficient way", I mean that if somebody really want to do a DoS, he'll better go with a DDoS tool like LOIC than guessing a specific url (which is not supposed to be public, even if for NTB it's currently hardcoded) to trigger ingestion.

The issue by using handshake or other shared secret is that it is vendor specific and not generic anymore. A hidden URL is a kind of shared secret anyway (we can autogenerate a key in the URL). An yes HTTPS is needed.

Maybe we can talk about this during the scrum?

petrjasek: for the record, what I mean is that we generate url for different push services like:
`superdesk.org/api/push?auth_token=abc`

jerome-poisson:
@petrjasek @naspeh after this morning discussion, I've added an authentication mechanism based on auth parameter and environment variable.
I'm not really satisfied with the hack done for the Service because no backend is attached, I'm not use if there is a simpler way to do that with Eve (maybe adding an URL directly with flask?). We can discuss it tomorrow.

Also @naspeh, I would like to set the environment variable for the test instance to try it with Wufoo, if we can check that together tomorrow, thanks :)

apps/io/feeding_service_webhook.py
```diff
@@ -0,0 +1,100 @@
...
+from superdesk.io.registry import registered_webhooks
+from flask import request, abort
+import os
+import superdesk
+import logging
+
+log = logging.getLogger(__name__)
+
+
+class FeedinServviceWebhookAuth(object):
```
petrjasek: typo Ser**vv**ice

apps/io/feeding_service_webhook.py
```diff
@@ -0,0 +1,100 @@
...
+        try:
+            auth_key = os.environ[env_name]
+        except KeyError:
+            log.error('{} environment variable is not set'.format(env_name))
+            return False
+        return request.args.get('auth') == auth_key
+
+
+class FeedingServiceWebhookResource(Resource):
+    resource_methods = ['GET', 'POST']
```
petrjasek: I would remove `GET` if possible, imo it triggers an action, so should be `POST`

superdesk/io/registry.py
```diff
@@ -115,3 +118,29 @@ def service(service_id):
         return ListCursor(
             [service(_id) for _id in registered_feeding_services]
         )
+
+
+def register_webhook(provider_name, webhook_name, webhook_service=None):
```
petrjasek: actually why would we create multiple resources for this? couldn't be provider a param as well?
`/webhook?provider=woofoo`

apps/io/feeding_service_webhook.py
jerome-poisson: right thanks

apps/io/feeding_service_webhook.py
jerome-poisson: I've put it just in case, but we can still add it in the future if needed, I'll remove

superdesk/io/registry.py
jerome-poisson: yes that's an option which is probably cleaner, I'll change that

petrjasek: maybe this should be provider id rather than type, it might make things simpler

apps/io/feeding_service_webhook.py
```diff
@@ -0,0 +1,100 @@
...
+log = logging.getLogger(__name__)
+
+
+class FeedinServviceWebhookAuth(object):
+
+    def authenticate(self):
+        abort(403, description='You are not authorized to access this resource')
+
+    def authorized(self, allowed_roles, resource, method):
+        env_name = '{}_auth'.format(resource)
```
naspeh: probably name should be in uppercase format

jerome-poisson: I was hesitating actually :). . OK I'll change

superdesk/io/registry.py
jerome-poisson: @petrjasek provider id is not visible in ingest setting UI. As we have to construct URL manually, I'll use provider name for now.
In a future step, when we change the UI, I think a webhook URL should be automatically generated (with auth key) and displayed, and provider id should then be used indeed.

jerome-poisson:
@petrjasek @naspeh I've refactored the patch and removed multiple endpoints and GET handling. As a result the P.R. is much simpler, and there is no more need for a registration API (variable environment is enough). I'll close the P.R. on ntb repos which is no more needed. I've also improved the doc.

jerome-poisson:
tested OK with Wufoo

superdesk/io/webhooks/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+class FeedinServiceWebhookAuth(object):
+
+    def authenticate(self):
+        abort(403, description='You are not authorized to access this resource')
+
+    def authorized(self, allowed_roles, resource, method):
+        try:
+            # provider's id may be used in the future
+            provider_name = request.args['provider_name']
```
petrjasek: imo this part shouldn't be in authorization handling, only later.

jerome-poisson: @petrjasek that's because there is a bug, I've renamed the variable but forgot to put it in the lookup, it should be used to start only the provider requested, and for now it's starting all providers, I'll fix that

jerome-poisson: @petrjasek actually forget my last message, I've misread the diff and the lookup is OK.
the provider_name is needed here for checking the environment variable, and the page is not authorised if nothing is set in environment, so that's kind of needed here.

It may still be changed later when we have some client UI, because at this time the environment variable should not be used anymore.

PR #808, opened 2017-02-20
Ingest supports generated items in `on_create`

vied12:


superdesk/io/commands/update_ingest.py
```diff
@@ -414,18 +415,18 @@ def ingest_items(items, provider, feeding_service, rule_set=None, routing_scheme
...
     # sync mongo with ingest after all changes
     ingest_collection = feeding_service.service if hasattr(feeding_service, 'service') else 'ingest'
     ingest_service = superdesk.get_resource_service(ingest_collection)
-    updated_items_ids = [item['_id'] for item in all_items if item[GUID_FIELD] not in failed_items]
-    updated_items = ingest_service.find({'_id': {'$in': updated_items_ids}}, max_results=len(updated_items_ids))
-    app.data._search_backend(ingest_collection).bulk_insert(ingest_collection, [item for item in updated_items])
-
+    updated_items = ingest_service.find({
+        '_id': {'$in': created_ids},
+        GUID_FIELD: {'$not': {'$in': failed_items}},
```
akintolga: I think we can still filter out failed item ids locally instead of asking mongo.

vied12: Thanks for the review. I thought Mongodb would have done this faster but I will edit

PR #812, opened 2017-02-22
[SDESK-745](feat):Adding product types to products.

mdhaman:
This feature adds Product Type attribute to products. A product can be of following types:
- Direct
- API
- Both

This classification allows the user to permission different products on a single subscriber based on API delivery or non-API delivery method. Enqueue process for publishing content into API will validate the articles against products of types `API` or `BOTH` whereas in case for non-API publishing the enqueue process will validate against the product type `Direct` or `Both`.

A record (placeholder) is created in the `publish_queue` for each subscriber to indicate content is published to content api.

- [x] Add product type to products.
- [x] Change the enqueue process to handle product types for normal publishing
- [ ] Change the enqueue process to handle product types for targeted publishing
- [ ] Resend to Subscribers
- [x] Change the publish queue to handle delivery_type = 'content_api'
- [ ] Behave and Unit tests


mdhaman:
change in requirement. not ready yet.

PR #815, opened 2017-02-22
fixed validate exception catching

jerome-poisson:
etree.get_text can now raise ValueError, so it need to be catched too.


jerome-poisson:
Note: there is no ticket for this issue, I just found it while testing the code.

PR #825, opened 2017-03-06
refactor(formatter): generate seq num before formatting it

petrjasek:
this part seems generic so there should be no need to do this inside formatter.
next step would be to avoid returning seq_num from format call, just
checking here if anyone can see any issue with that?
ping @akintolga @mdhaman @marwoodandrew

### DIFF
--- a/apps/publish/enqueue/enqueue_service.py
+++ b/apps/publish/enqueue/enqueue_service.py
@@ -344,8 +344,11 @@ class EnqueueService:
                             no_formatters.append(destination['format'])
                             continue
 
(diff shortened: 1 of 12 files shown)


mdhaman:
@petrjasek we are generating multiple `publish_queue` records for the same subscriber with different anpa categories for the same item. See https://github.com/superdesk/superdesk-aap/blob/master/server/aap/publish/formatters/aap_ipnews_formatter.py#L49. We need different sequence number for each record. Hence it was part of the formatter function.

petrjasek:
I see

PR #826, opened 2017-03-06
[content type] fixed overwritting of minlenght value

jerome-poisson:
minlength was overwritten when required was set, and always set to 1.
This commit fixes it by not overwritting minlenght if it is set and not
0.

fixes SDESK-698


jerome-poisson:
@petrjasek done, when the checks are OK I'll merge it

PR #828, opened 2017-03-09
feat(destinations): add internal destinations

petrjasek:
it allows you to define internal destinations and on publish
a copy of item is sent there. you can pick filter, desk, stage
and macro for each.

SDESK-613


mdhaman:
@petrjasek `archive_history` will record this action as `duplicate`. Should a record be added for `internal routed` content in the `archive_history` as well?

petrjasek:
duplicate is fine imo, that's what it does

PR #830, opened 2017-03-10
[sdesk-802] switch to lxml parser

marwoodandrew:


marwoodandrew:
Preserved content is html encode text wrapped in a <pre> tag, so use the html parser

PR #832, opened 2017-03-10
(feat)[SDESK-745] Adding product types to products.

mdhaman:
- Added `product type` field to product to differentiate between api and direct publishing. 
Direct publishing: All publishing modes except api 
- Added `api_products` to subscriber


mdhaman:
@akintolga can have look at this pr

akintolga:
Maybe these test cases would increase the coverage, but ignore the comment if they are already blended in to existing cases:
- update product to direct fails if set in subscriber as api (and vice versa)
- Resend to a subscriber having both api and direct products
- Correction goes to api again even if the api_product is removed from subscriber (i don't know if this is the valid requirement for api)

Also do we need a script to update existing products with product_type 'both'?

mdhaman:
@akintolga those test case are taken care of. I have added few more. 

We don't need a script, default product type is `both` on content validation.

PR #834, opened 2017-03-14
[FORMATTER] [NITF] <br/> handling

jerome-poisson:
`<br/>` element is now added to NITF output. If its parent is not a
possible parent of enrichedText group (see
https://www.iptc.org/std/NITF/3.6/documentation/nitf-3-6.html#Link5),
the element is removed.

`<br/>` are also removed if they are the last child (without trailing
text) of a `<p/>` element.

SDNTB-398

### DIFF
--- a/tests/publish/nitf_formatter_tests.py
+++ b/tests/publish/nitf_formatter_tests.py
@@ -105,6 +105,62 @@ class NitfFormatterTest(TestCase):
             </div>""").replace('\n', '').replace(' ', '')
         self.assertEqual(etree.tostring(nitf, encoding='unicode').replace('\n', '').replace(' ', ''), expected)
 
+    def test_html2nitf_br(self):
+        """Check that <br/> is kept if it is a child of and enrichedText parent element"""
+        html = etree.fromstring(dedent("""\
+            <div>
+                <br/>the previous tag should be removed (but not the text)
+                    <p>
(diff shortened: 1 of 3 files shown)


tests/publish/nitf_formatter_tests.py
```diff
@@ -105,6 +105,62 @@ def test_html2nitf(self):
             </div>""").replace('\n', '').replace(' ', '')
         self.assertEqual(etree.tostring(nitf, encoding='unicode').replace('\n', '').replace(' ', ''), expected)
 
+    def test_html2nitf_br(self):
+        """Check that <br/> is kept if it is a child of and enrichedText parent element"""
+        html = etree.fromstring(dedent("""\
+            <div>
+                <br/>the previous tag should be removed (but not the text)
+                    <p>
+                        the following tag <br/> should still be here
```
petrjasek: can you test it with `<br>` ? afaik that's what we get from editor

jerome-poisson: Ok

jerome-poisson: @petrjasek actually this method is only used by NTB which replaces all <br> by <br/>. I've added the same behaviour in Formatter so that we'll not get surprise if we move core NITF formatter to html2nitf. I keep the <br/>  because that's what html2nitf get.

PR #835, opened 2017-03-15
[sdesk-743] Previous versions of killed item should not be visible in…

marwoodandrew:
… the Content API


marwoodandrew:
WIP until I work out why the composite item kill are not getting published to the API

PR #836, opened 2017-03-15
fix(app): avoid mongo connections in multi process environment

petrjasek:
and thus avoid that pymongo warning. by default it will only create
indexes when running manage.py script.

SDESK-257

### DIFF
--- a/superdesk/factory/app.py
+++ b/superdesk/factory/app.py
@@ -15,6 +15,7 @@ import importlib
 import jinja2
 import eve
 import superdesk
+
 from flask_mail import Mail
 from eve.io.mongo import MongoJSONEncoder
 from eve.render import send_response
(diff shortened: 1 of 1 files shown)


naspeh:
I always had a question, so it's time to ask:
Why do we need this `init_elastic` when run `./manage.py`, it is a problem in some cases, when I run `./manage.py -h` with clean database yet, I'm not assuming than it creates elastic indexes after that, it's really strange.

We have `./manage.py app:initialize_data` during installation, so we can create all stuff needed here...

And when I run `./manage.py -h` it shouldn't even check elastic indexes are exist or not...

What is I missing?

petrjasek:
that could be a step further, atm it creates those indexes also when running honcho start, so also for celery and gunicorn processes.

naspeh:
Yep, lets put this stuff to `./manage.py app:initialize_data` next time

PR #840, opened 2017-03-16
feat(json): update json output

petrjasek:
- add copyrightholder and copyrightnotice using rightsinfo vocab
- fix place items to have code instead of qcode
- output filtered profile name instead of ObjectId
- add genre
- add ednote
- add signal for legal flag

SDESK-891 SDESK-807


petrjasek:
ping @mdhaman

PR #849, opened 2017-03-21
[SDESK-878] - Purge exported files from storage.

MarkLark86:
The following functionality was added to both **superdesk.storage.SuperdeskGridFSMediaStorage** and **superdesk.storage.amazon.AmazonMediaStorage**:

- **.put()** Ability to store the file in a 'folder' when calling 'put' on the service
- **.find()** Ability to 'find' files using a combination of 'prefix' and 'upload_date'

The **superdesk.download** blueprint now also supports the folder in the url via a second route rule.

The **apps.export.service** service now uses the folder functionality to store the temporary files under the 'temp' folder.
The **superdesk.commands.remove_exported_files** command is then able to efficiently find the files under 'temp' and ones that are older than 24hrs.

### DIFF
--- a/superdesk/storage/amazon/amazon_media_storage.py
+++ b/superdesk/storage/amazon/amazon_media_storage.py
@@ -15,6 +15,7 @@ import json
 import logging
 from mimetypes import guess_extension
 from superdesk.media.media_operations import download_file_from_url
+from superdesk.utc import query_datetime
 import time
 
 import boto3
@@ -213,7 +214,8 @@ class AmazonMediaStorage(MediaStorage):
                         logger.exception(ex)
         return headers
 
-    def put(self, content, filename=None, content_type=None, resource=None, metadata=None, _id=None, version=True):
+    def put(self, content, filename=None, content_type=None, resource=None, metadata=None, _id=None, version=True,
+            folder=None):
         """Save a new file using the storage system, preferably with the name specified.
 
         If there already exists a file with this name name, the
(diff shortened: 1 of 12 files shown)


mdhaman:
@petrjasek @ioanpocol can you guys review this PR.

superdesk/storage/amazon/amazon_media_storage.py
```diff
@@ -286,3 +305,45 @@ def remove_unreferenced_files(self, existing_files):
                 print('Failed to clean orphans: {}'.format(errors))
         else:
             print('There\'s nothing to clean.')
+
+    def find(self, prefix=None, upload_date=None, resource=None):
+        """Search for files in the S3 bucket
+
+        Searches for files in the GridFS using a combination of filename prefix and upload date
```
petrjasek: here the comment is about gridfs while this is S3 implementation

superdesk/storage/desk_media_storage.py
```diff
@@ -63,35 +63,84 @@ def url_for_download(self, media_id, content_type=None):
...
         :param content_type: mime type
         :param metadata: file metadata
         :param resource: type of resource
+        :param str folder: Folder that the file will be stored in
+        :return str: The ID that was generated for this object
         """
         if '_id' in kwargs:
             kwargs['_id'] = bson.ObjectId(kwargs['_id'])
+
+        if folder:
```
petrjasek: we could also use gridfs buckets for folder, that would avoid that regex query in gc, but not sure if it's a big deal

superdesk/storage/amazon/amazon_media_storage.py
```diff
@@ -286,3 +305,45 @@ def remove_unreferenced_files(self, existing_files):
                 print('Failed to clean orphans: {}'.format(errors))
         else:
             print('There\'s nothing to clean.')
+
+    def find(self, prefix=None, upload_date=None, resource=None):
```
petrjasek: imo we can call this `prefix` param `folder` as well, it's supposed to use same value right?

superdesk/storage/desk_media_storage.py
mdhaman: There is already a filename:asc uploadDate:asc index created on gridfs. Regex query is prefix query and the query hits appropriate index as well. I think that it should not be a problem.

petrjasek: :ok:

PR #850, opened 2017-03-22
[SDESK-847] Mark media items associated with a text item as 'used'

nrvikas:


apps/archive/archive.py
```diff
@@ -265,17 +265,25 @@ def _handle_media_updates(self, updates, original, user):
...
 
             self._validate_updates(stored_item, item_obj, user)
             if stored_item[ITEM_TYPE] == CONTENT_TYPE.PICTURE:  # create crops
                 CropService().create_multiple_crops(item_obj, stored_item)
+
+            if not media_item:
+                media_item = self.find_one(req=None, _id=_id)
+
+            if media_item and original.get(ITEM_TYPE) == CONTENT_TYPE.TEXT and not media_item.get('used'):
+                self.update(media_item['_id'], {'used': True}, media_item)
```
mdhaman: use `system_update`. don't want to change the etag.

akintolga:
Just double checking: If the media item is removed from text item then it doesn't become unused again. Is this as expected?

mdhaman:
@akintolga removed the check for the time being. Check the ticket. Users should be able to find the image thru search. There might be more changes coming in this area.

PR #853, opened 2017-03-22
[FORMATTER] [HTMLNewsMLG2] use self-closing tags for void elements, a…

jerome-poisson:
…nd start/end pairs otherwise

SDESK-947


jerome-poisson:
@naspeh yes I guessed it was not a pb on this branch, and the port of this P.R. on master is OK with tests

PR #874, opened 2017-04-05
fix(autosave): fix work being lost on unlock

petrjasek:
when user unlocks an item save autosave version (if any) as new
version.

SDESK-953

### DIFF
--- a/apps/item_lock/components/item_lock.py
+++ b/apps/item_lock/components/item_lock.py
@@ -7,15 +7,19 @@
 # For the full copyright and license information, please see the
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
-from superdesk.metadata.item import ITEM_STATE, CONTENT_STATE, ITEM_TYPE, CONTENT_TYPE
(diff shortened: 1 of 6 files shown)


apps/item_lock/components/item_lock.py
```diff
@@ -108,6 +111,13 @@ def unlock(self, item_filter, user_id, session_id, etag):
                 superdesk.get_resource_service('archive').delete_action(lookup={'_id': item['_id']})
                 push_content_notification([item])
             else:
+                autosave = superdesk.get_resource_service('archive_autosave').find_one(req=None, _id=item['_id'])
+                if autosave:
+                    autosave['force_unlock'] = True  # to avoid item locked check
+                    resolve_document_version(autosave, 'archive', 'PATCH', item)
+                    superdesk.get_resource_service('archive').patch(item['_id'], autosave)
```
mdhaman: There is a workflow transition check on patch call. Can you just check that this work for `correct` and `kill` action?

petrjasek:
updated to only make new version from autosave if item was not published yet

PR #878, opened 2017-04-10
[SDESK-873] - Fix celery schedule for purging temporary files, was no…

MarkLark86:
…t previously set

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -261,6 +261,7 @@ CELERY_BEAT_SCHEDULE = {
     },
     'temp_files:gc': {
         'task': 'superdesk.commands.temp_file_expiry',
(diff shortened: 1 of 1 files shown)


superdesk/default_settings.py
```diff
@@ -598,4 +599,4 @@ def celery_queue(name):
 #: The number of hours before temporary media files are purged
 TEMP_FILE_EXPIRY_HOURS = int(env('TEMP_FILE_EXPIRY_HOURS', 24))
 #: The number of days before content api items are removed, defaults to approximately 6 months
```
naspeh: Please changes the comment too in this case...

@petrjasek is it ok a such default?

petrjasek: yes

PR #883, opened 2017-04-12
fix(celery): set celery timezone to UTC

petrjasek:
otherwise it might stop processing tasks for an hour when there is
daylight savings time change.

SDESK-1057

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -16,8 +16,9 @@ Environment variables names match config name, with some expections documented b
 
 import os
 import json
+import pytz
 
-from datetime import timedelta
+from datetime import timedelta, datetime
 from celery.schedules import crontab
 from kombu import Queue, Exchange
 import tzlocal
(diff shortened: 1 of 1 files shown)


petrjasek:
@mdhaman when this will be deployed you should remove celerybeatschedule.db file, otherwise you would have ~10h break

marwoodandrew:
Might want to check the celery task that are scheduled to execute once per day at given times! (audit purge, api purge)

petrjasek:
@marwoodandrew good point, made a helper for this so it will use definition in local time and configure it with utc.
also seems like you don't have to remove that schedule db file, it finds out that there was a config change and resets db automatically.

superdesk/default_settings.py
```diff
@@ -55,6 +56,24 @@ def celery_queue(name):
...
+
+if not DEFAULT_TIMEZONE:
+    raise ValueError("DEFAULT_TIMEZONE is empty")
+
+
+def local_to_utc_hour(hour):
+    now = datetime.now()
+    tz = pytz.timezone(DEFAULT_TIMEZONE)
+    local = tz.localize(now.replace(hour=hour))
+    print('local', hour, 'utc', local.utctimetuple()[3])
```
naspeh: should be removed :)

petrjasek: ups

PR #884, opened 2017-04-14
Get rid hardcoded urls for AWS S3

naspeh:
Related: https://dev.sourcefabric.org/browse/SDESK-1099


superdesk/storage/amazon_media_storage.py
```diff
@@ -84,63 +83,35 @@ class AmazonMediaStorage(MediaStorage):
...
-            aws_secret_access_key=api_key,
-            region_name=self.region,
+            aws_access_key_id=self.app.config['AMAZON_ACCESS_KEY_ID'],
+            aws_secret_access_key=self.app.config['AMAZON_SECRET_ACCESS_KEY'],
+            region_name=self.app.config.get('AMAZON_REGION'),
             config=Config(signature_version='s3v4'),
         )
         self.user_metadata_header = 'x-amz-meta-'
 
     def url_for_media(self, media_id, content_type=None):
```
petrjasek: seems like this could be defined now on some base class we could use for both amazon and gridfs storage,
same for `url_for_download`

superdesk/storage/desk_media_storage.py
```diff
@@ -34,6 +34,9 @@ def get(self, _id, resource=None):
                     logger.exception('Failed to load metadata for file: %s with key: %s and value: %s', _id, k, v)
         return media_file
 
+    def get_redirect_url(self, media_id):
```
petrjasek: this could be also in the base class

naspeh:
> any performance impact

no, I don't think so, it only construct url for amazon and don't check if it exists or not

superdesk/storage/amazon_media_storage.py
naspeh: yep, I was thinking about that too, so I'll do it

naspeh: so they are bases on different classes
```
SuperdeskGridFSMediaStorage -> eve.io.mongo.media.GridFSMediaStorage
AmazonMediaStorage -> eve.io.media.MediaStorage
```
and I don't want to use multiple inheritance, that why I kept them like this

naspeh:
It was PR with wrong branch ([naspeh](https://github.com/superdesk/superdesk-core/tree/naspeh)) I'll create another one for master %)

PR #888, opened 2017-04-18
fix(legal): make legal archive optional

petrjasek:
use `LEGAL_ARCHIVE` env or `LEGAL_ARCHIVE` setting to disable it.

SDESK-1046

### DIFF
--- a/apps/legal_archive/commands.py
+++ b/apps/legal_archive/commands.py
@@ -35,6 +35,11 @@ from superdesk.utc import utcnow
 logger = logging.getLogger(__name__)
 
 
+def is_legal_archive_enabled():
+    """Test if legal archive is enabled."""
+    return app.config['LEGAL_ARCHIVE']
+
+
 class LegalArchiveImport:
     log_msg_format = "{{'_id': {_id}, 'unique_name': {unique_name}, 'version': {_current_version}, " \
                      "'expired_on': {expiry}}}."
@@ -405,6 +410,8 @@ def import_into_legal_archive(self, item_id):
     :param self: celery task
     :param str item_id: document id to import into legal_archive
     """
+    if not is_legal_archive_enabled():
+        return
     try:
         LegalArchiveImport().upsert_into_legal_archive(item_id)
     except Exception:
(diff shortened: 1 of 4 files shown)


mdhaman:
@petrjasek check for legal archive is needed for this function as well https://github.com/petrjasek/superdesk-core/blob/6964a552592d33331d249d1c7cee98af412b2608/apps/legal_archive/commands.py#L402

This change won't show legal archive but the side effect will be that items won't get expired.

apps/legal_archive/commands.py
```diff
@@ -425,6 +425,8 @@ class ImportLegalPublishQueueCommand(superdesk.Command):
     ]
 
     def run(self, page_size=None):
+        if not app.config['LEGAL_ARCHIVE']:
```
naspeh: I think it is not needed here, these commands won't registered with `LEGAL_ARCHIVE=0` option

superdesk/default_settings.py
```diff
@@ -116,6 +116,9 @@ def local_to_utc_hour(hour):
 #: full mongodb connection uri, overrides ``MONGO_DBNAME`` if set
 MONGO_URI = env('MONGO_URI', 'mongodb://localhost/%s' % MONGO_DBNAME)
 
+#: legal archive on switch, set to ``0`` to disable it
+LEGAL_ARCHIVE = env('SUPERDESK_LEGAL_ARCHIVE', '1') != '0'
```
naspeh: I think we should be consistent with config values, probably just empty string is enough for disabling.

Also I think env variable should be `LEGAL_ARCHIVE` instead of `SUPERDESK_LEGAL_ARCHIVE`, because we don't have `SUPERDESK_` prefix for setting like 
```
root@sds-naspeh:/opt/superdesk/server-core# env | grep LEGAL
LEGAL_ARCHIVE_URI=mongodb://data-sd/sds-naspeh_la
```
and many others.

apps/legal_archive/commands.py
petrjasek: it won't register it for manage.py, but it is registered also via celery beat and that would still run

naspeh: I see, thanks for clarification

petrjasek:
@naspeh right

PR #889, opened 2017-04-18
[ingest] [feed parsers] ingest:import + WordPress WXR feed parser

jerome-poisson:
new ingest:import command allows to import stories in "archived" database
using any registered feed parser.

mapping for XML files used in NITF has been moved to XMLFeedParser so it
can be used by other parsers.

XML mapping has been improved to take profit of lxml and its xpath
implementation, it is simplified:
	- filter_value and filter have been merged, filter is always
	  used
	- /@attribute is natively handled by xpath, the "attribute"
	  mapping key is not needed anymore and has been removed
	- text value are handled, so in addition to /@attribute, xpath
	  methods like text() can now be used
	- mapping can now use an OrderedDict if order of parsing is
	  important
	- new list mapping key, if several values are expected

SkipValue has been moved to superdesk.errors

etree.html_parser now return only <body> part (in a <div>)

WordPress WXR importer use XML mapping
All images are uploaded to superdesk. First one is used as featuremedia,
other ones as embeds.

SDTS-4


jerome-poisson:
tests added, I've also made a couple of improvements (default values for headline, alt_text and description_text for media, binary read of files in import, and other things I forgot :) ).

naspeh:
Awesome, thanks!

superdesk/io/importers/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+            parsed = etree.parse(f)
+        articles = feed_parser.parse(parsed)
+        if profile is not None:
+            for article in articles:
+                article['profile'] = profile
+        archived_service = get_resource_service("archived")
+        archived_service.post(articles)
+
+
+superdesk.command('ingest:import', ImportCommand())
```
petrjasek: imo the name should be more like `archive:import`

petrjasek: or maybe `xml:import` given that it only supports xml formats

jerome-poisson: right, I changed after for archived, good catch !

PR #891, opened 2017-04-19
Optional relative media urls for SaaS

naspeh:
Related: https://dev.sourcefabric.org/browse/SDESK-1099

Replacement for #884

After conversation with @omniton I've changed the code, so now we have relative urls (without host) in database and renamed endpoint `/api/upload-raw` which serves file or redirects to amazon, Andrey can override this endpoint with S3 proxy settings, so no additional redirects in this case

**UPDATED:**
So I added new setting `MEDIA_PREFIX` and removed `AMAZON_SERVER` and `AMAZON_SERVER_PROXY`, etc. The updated documentation is here [docs/settings.html#media-prefix](http://sdspr-891.test.superdesk.org/docs/settings.html#media-prefix)

Also there will be following task with a command to fix existing data with new `MEDIA_PREFIX` [SDESK-1100](https://dev.sourcefabric.org/browse/SDESK-1100).

BTW: PR instance http://sdspr-891.test.superdesk.org is configured with relative urls, so you can play a bit with it:
```sh
$ cat tpl/init/sdspr-891.sh 
{{>init/sd.sh}}

cat <<"EOF" >> {{config}}
{{>init/.amazon.sh}}
# use bucket from Frankfurt
AMAZON_CONTAINER_NAME='sd-frankfurt-test'
AMAZON_REGION=eu-central-1
MEDIA_PREFIX=/sdspr-891
EOF


cat <<"EOF" > /etc/nginx/conf.d/media.inc
location /sdspr-891/ {
    return 302 https://sd-frankfurt-test.s3-eu-central-1.amazonaws.com$request_uri;
}
EOF
```


naspeh:
I'm going to rework this and add new config option like `MEDIA_URL`, so it'll depend on this option relative urls will be saved to database or not.

naspeh:
@petrjasek I've updated this PR, please have a look again, thanks!

naspeh:
BTW: I picked name `MEDIA_PREFIX`, because [eve uses](http://python-eve.org/config.html)
```
MEDIA_BASE_URL
MEDIA_ENDPOINT
MEDIA_URL
```
and they have a little different meaning, so I decided to pick another one

superdesk/upload.py
```diff
@@ -28,7 +30,7 @@
 cache_for = 3600 * 24 * 30  # 30d cache
 
 
-@bp.route('/upload/<path:media_id>/raw', methods=['GET'])
```
petrjasek: imo we can keep this route for BC, could be redirect

naspeh: yep, I'll add redirect

naspeh:
Yes, we discussed this with @omniton, probably there will be two endpoints in nginx for client and api, so this is just configuration thing.

petrjasek:
and what about development env? atm server runs by default on `localhost:5000`,
client on `localhost:9000`

naspeh:
it uses full url by default, so no changes for dev needed now

naspeh:
I just added redirect for old url.

Also there will be following task with command or `data:upgrade` to fix existing data with new `MEDIA_PREFIX`, so probably  this redirect is not really needed...

naspeh:
Updated docs a little with notice about default behavior.

naspeh:
@akintolga, @mdhaman any feedback from you on this?

naspeh:
So if some installations have using direct serving from Amazon S3, it requires the change like
```
MEDIA_PREFIX='https://<bucket>.s3-<region>.amazonaws.com/<subfolder>'
```

PR #895, opened 2017-04-21
added new fields to aggregations list

ancafarcas:


akintolga:
@ancafarcas Are these aggregations for analytics only? If so maybe we should use them only if we're on analytics page.

mdhaman:
These aggregations are used by global search. We have close to 10 million documents in our system. It would be good to keep analytics related aggregates separate.

akintolga:
pinging @mugurrus

ancafarcas:
@akintolga:  Yes, the new aggregations are used only by analytics at the moment. 
@mdhaman: I'm going to talk to Mugur about the possibility of keeping this aggregations only for the analytics

PR #896, opened 2017-04-21
fix(init): fix init when legal archive resources are disabled

petrjasek:


naspeh:
there is still error
```
INFO:superdesk:Uhoh, an exception occured while running the command...
ERROR:superdesk:'legal_archive'
  File "/opt/superdesk/server-core/superdesk/__init__.py", line 61, in __call__
    res = self.run(*args, **kwargs)
  File "/opt/superdesk/server-core/apps/prepopulate/app_prepopulate.py", line 183, in run
    prepopulate_data(prepopulate_file, get_default_user())
  File "/opt/superdesk/server-core/apps/prepopulate/app_prepopulate.py", line 65, in prepopulate_data
    service = get_resource_service(resource)
  File "/opt/superdesk/server-core/superdesk/__init__.py", line 112, in get_resource_service
    return resources[resource_name].service
KeyError: 'legal_archive'
```
I think it's better to extend `__entities__` based on `LEGAL_ARCHIVE` option

petrjasek:
it's not about entities, we have it in sample data

PR #900, opened 2017-04-24
feat(entrypoints): provide entrypoints for procfile

petrjasek:
so that you can use core without extra code in superdesk repo.

usage:

    rest: gunicorn -c python:superdesk.gunicorn superdesk.wsgi
    capi: gunicorn -c python:superdesk.gunicorn content_api.wsgi
    wamp: python3 -u -m superdesk.ws
    work: celery -A superdesk.worker worker
    beat: celery -A superdesk.worker beat --pid=

### DIFF
--- a/superdesk/cli.py
+++ b/superdesk/cli.py
@@ -1,0 +1,15 @@
+#
+import superdesk
+
+from flask_script import Manager
+from superdesk.factory import get_app
+
+
+def main():
+    app = get_app()
+    manager = Manager(app)
+    manager.run(superdesk.COMMANDS)
+
+
+if __name__ == '__main__':
+    main()
--- a/superdesk/wsgi.py
+++ b/superdesk/wsgi.py
@@ -1,0 +1,20 @@
+"""WSGI entrypoint.
+
+Usage via Procfile:
+
+    rest: gunicorn -c pythong:superdesk.gunicorn superdesk.wsgi
+
+For debug mode, run it as: ``python -m superdesk.wsgi``.
+"""
+
+import os
+
+from superdesk.factory import get_app
+
+application = get_app()
+
+if __name__ == '__main__':
+    debug = True
+    host = '0.0.0.0'
+    port = int(os.environ.get('PORT', '5000'))
+    application.run(host=host, port=port, debug=debug, use_reloader=debug)
--- a/apps/archive/common.py
+++ b/apps/archive/common.py
@@ -32,11 +32,11 @@ from superdesk.metadata.item import GUID_NEWSML, GUID_FIELD, GUID_TAG, not_analy
 from superdesk.metadata.packages import PACKAGE_TYPE, TAKES_PACKAGE, SEQUENCE, ASSOCIATED_TAKE_SEQUENCE
 from superdesk.metadata.utils import generate_guid
 from superdesk.errors import SuperdeskApiError, IdentifierGenerationError
-from superdesk.logging import logger
 from apps.auth import get_user
 
 
 logger = logging.getLogger(__name__)
+
 ARCHIVE = 'archive'
 CUSTOM_HATEOAS = {'self': {'title': 'Archive', 'href': '/archive/{_id}'}}
 ITEM_OPERATION = 'operation'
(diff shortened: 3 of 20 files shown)


superdesk/cli.py
```diff
@@ -0,0 +1,15 @@
+#
+import superdesk
+
+from flask_script import Manager
+from superdesk.factory import get_app
+
+
+def main():
+    app = get_app(init_elastic=True)
```
naspeh: `init_elastic` is not needed anymore

superdesk/wsgi.py
```diff
@@ -0,0 +1,11 @@
...
+
+Usage via Procfile:
+
+    rest: gunicorn -c pythong:superdesk.gunicorn_config superdesk.wsgi
+
+"""
+
+from superdesk.factory import get_app
+
+application = get_app()
```
naspeh: I think it could be combined with `superdesk.worker`, probably in module `superdesk.honcho` and usage like
```
rest: gunicorn -c python:superdesk.gunicorn_config superdesk.honcho
work: celery -A superdesk.honcho worker
beat: celery -A superdesk.honcho beat --pid=
```

petrjasek: yep that would be possible, not sure about naming it honcho, it's not honcho specific in any way.
maybe just call it `superdesk` and do it there `if __name__ == '__main__':` or something:
```
rest: gunicorn -c python:superdesk.gunicorn_config superdesk
work: celery -A superdesk worker
beat: celery -A superdesk beat --pid=
```

naspeh: I think `if __name__ == '__main__':` won't work, but not sure, but if we can somehow, then sounds good

petrjasek: yep it won't work

petrjasek: I would rather keep the naming as is, it is more explicit imo.

naspeh: ok, lets keep the naming :+1:

naspeh:
`fire:www` is green now, because of this:
```
$ cat tpl/init/sdspr-900.sh 
{{>init/sd.sh}}

cat <<"EOF" >> {{config}}
NEW_RELIC_LICENSE_KEY=
EOF
```
:)

superdesk/wsgi.py
naspeh: by the way, maybe `superdesk.worker` should be renamed to `superdesk.celery` then for more explicit :)

petrjasek:
@naspeh I've added a check for that `newrelic.ini` file

naspeh:
it's still here:
```
+ python manage.py app:initialize_data
manage.py:15: ExtDeprecationWarning: Importing flask.ext.script is deprecated, use flask_script instead.
  from flask.ext.script import Manager
  File "manage.py", line 16, in <module>
    from app import get_app
  File "/opt/superdesk/server/app.py", line 21, in <module>
    newrelic.agent.initialize(os.path.abspath(os.path.join(os.path.dirname(__file__), 'newrelic.ini')))
  File "/opt/superdesk/env/lib/python3.5/site-packages/newrelic-2.66.0.49/newrelic/config.py", line 2532, in initialize
    log_file, log_level)
  File "/opt/superdesk/env/lib/python3.5/site-packages/newrelic-2.66.0.49/newrelic/config.py", line 791, in _load_configuration
    'environment "%s".' % (_config_file, _environment))
newrelic.api.exceptions.ConfigurationError: Configuration has already been done against differing configuration file or environment. Prior configuration file used was "newrelic.ini" and environment "None".
```
[from log](https://test.superdesk.org/logs/2/all/20170427-152721-11-sdspr-900/www.log.htm)

I think it's because we have this: https://github.com/superdesk/superdesk/blob/master/server/app.py#L21, so it inits two times

petrjasek:
I see, this file will be gone, so it's probably ok

naspeh:
yep, so probably your last changes is not needed then

petrjasek:
it would fail there if that env is configured but newrelic.ini is missing, but maybe it should fail then

naspeh:
ok, as you wish ;)

PR #902, opened 2017-04-26
additional date format for analytics behave tests

ancafarcas:


superdesk/tests/steps.py
```diff
@@ -248,7 +253,10 @@ def apply_placeholders(context, text):
                 if unit != -1:
                     value -= timedelta(days=int(placeholder[unit + 1]))
 
-            value = format_date(value)
+            if placeholder == 'DATE_FORMATTED':
```
petrjasek: can it be `ANALYTICS_DATE_FORMATTED` to make it more obvious?

ancafarcas: Sure, it can be. I'll modify it right away

PR #905, opened 2017-04-27
feat(ingest): test provided config on create/update

petrjasek:
ingest providers can implement `_test` method which is used
by ingest service to test given configuration.

in case there is an error it won't save the changes.

SDESK-638

### DIFF
--- a/superdesk/io/feeding_services/__init__.py
+++ b/superdesk/io/feeding_services/__init__.py
@@ -45,8 +45,39 @@ class FeedingService(metaclass=ABCMeta):
         :type update: dict
         :return: a list of articles which can be saved in Ingest Collection.
         """
-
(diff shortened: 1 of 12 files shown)


superdesk/io/feeding_services/__init__.py
```diff
@@ -45,9 +45,40 @@ def _update(self, provider, update):
...
+        :param provider: Ingest Provider Details.
+        """
+        return
+
+    def config_test(self, provider=None):
+        """Test provider configuration.
+
+        :param provider: provider data
+        """
+        if not provider:  # nosetests run this for some reason
```
naspeh: because the name `*_test`, I think we should move all tests under `tests` directory and change command like `nosetest tests`

PR #908, opened 2017-04-28
[media] fixed rendition when crop is not set

jerome-poisson:
rendition was wailing is crop was having empty string for "width" or
"height" key, this commit fix it

SDTS-24

### DIFF
--- a/superdesk/media/renditions.py
+++ b/superdesk/media/renditions.py
@@ -98,13 +98,23 @@ def can_generate_custom_crop_from_original(width, height, crop):
     if crop.get('ratio'):
         return True
 
-    if 'width' not in crop and 'height' not in crop:
+    try:
+        crop_width = int(crop['width'])
+    except (KeyError, TypeError):
+        crop_width = None
+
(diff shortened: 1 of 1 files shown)


superdesk/media/renditions.py
```diff
@@ -179,7 +181,7 @@ def _resize_image(content, size, format='png', keepProportions=True):
     assert isinstance(size, tuple)
     img = Image.open(content)
     width, height = img.size
-    new_width, new_height = [to_int(x) for x in size]
+    new_width, new_height = [int(x) for x in size]
```
naspeh: there is no `try...except`, so could fail in some point

jerome-poisson: I don't think so, this is weight and size fro PIL, so I expect it to be int. I don't even think that int() casting is needed, but I kept because it was there before.

jerome-poisson: I've actually misread, it's size and not img.size, so I've restored the orginal code for this part

PR #909, opened 2017-05-01
[SDESK-1206] [SDESK-1178] - Deleted and disabled content profiles

akintolga:
- [x] Deleting a content profile will remove profile data from template
- [x] Deleting a content profile will remove default_content_profile from desks
- [x] Content profile won't be disabled (as this creates issues in templates and desk i.e. fetched stories)
- [x] Remove the value from referencing templates if a field is removed from content profile
- [x] Remove the values of missing fields in template if content profile field of the template is changed


akintolga:
Addresses [SDESK-1178]

mdhaman:
@akintolga really really approved....

akintolga:
Unbelievable :)

PR #915, opened 2017-05-03
fix(media): fix clean images command

petrjasek:
- use prefix to filter out images not created by superdesk
- check legal archive only if enabled
- delete items in 1000 item chunks to avoid s3 limits

SDESK-1119

### DIFF
--- a/superdesk/commands/clean_images.py
+++ b/superdesk/commands/clean_images.py
@@ -10,6 +10,8 @@
 
 
 import superdesk
+
+from flask import current_app as app
 from superdesk.metadata.item import ASSOCIATIONS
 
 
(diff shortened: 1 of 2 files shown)


naspeh:
It works great with `AMAZON_S3_SUBFOLDER`

But still error with `GridFS`
```
root@sdspr-915:/opt/superdesk/server# ./manage.py app:clean_images
./manage.py:15: ExtDeprecationWarning: Importing flask.ext.script is deprecated, use flask_script instead.
  from flask.ext.script import Manager
Starting image cleaning.
unhashable type: 'list'
INFO:superdesk:Command finished with: None
```

PR #917, opened 2017-05-04
[SDESK-1211](fix) Modified command to initialize index only.

mdhaman:


apps/prepopulate/app_initialize.py
```diff
@@ -210,15 +210,17 @@ class AppInitializeWithDataCommand(superdesk.Command):
         superdesk.Option('--full-path', '-p', dest='path'),
         superdesk.Option('--sample-data', action='store_true'),
         superdesk.Option('--force', '-f', action='store_true'),
+        superdesk.Option('--init-index-only', '-i', dest='init_index_only'),
```
naspeh: `dest='init_index_only'` will be by default, so it's not really needed here

petrjasek: it should be probably with `action='store_true'` instead

mdhaman: done...

PR #918, opened 2017-05-04
Fix "app:index_from_mongo" with explicit option for full reindex

naspeh:


superdesk/commands/index_from_mongo.py
```diff
@@ -26,19 +26,22 @@ class IndexFromMongo(superdesk.Command):
...
+        superdesk.Option('--full', action='store_true'),
+        superdesk.Option('--page-size', '-p')
     ]
     default_page_size = 500
 
-    def run(self, mongo_collection_name, page_size):
-        if mongo_collection_name == 'all':
+    def run(self, collection_name, full, page_size):
+        if not collection_name and not full:
+            raise SystemExit('Specify --full to reindex all indexes')
```
petrjasek: maybe it should be `--all` then

naspeh: done

PR #926, opened 2017-05-11
fix(stage): abort changes to items on readonly stage

petrjasek:
SDESK-1243


mdhaman:
@petrjasek this would cause issues at our end where routing scheme uses this local read only stage for publishing. This is mainly used for auto publishing where no one allowed to touch the article.  

Can we have a restrictions that `working_stage` cannot be `local_read_only`? Maybe content creation is not allowed on local_read_only stage if its done by user but background process can create content & publish.

Same issue is with `Duplicate To` functionality.

petrjasek:
@mdhaman so it should be possible to duplicate content into read only stage right? or create content there. just no changes to content by users?

mdhaman:
No content creation by users should be fine.

petrjasek:
updated again, so user can duplicate content there or fetch from ingest, but can't create a new one there nor edit

PR #928, opened 2017-05-12
(fix): Updating mapping for associations & fixing AP ANPA dateline parsing

mdhaman:
- To use filtering based on featuremedia.
- Fixed the failing AP ANPA dateline parsing.

### DIFF
--- a/superdesk/io/feed_parsers/ap_anpa.py
+++ b/superdesk/io/feed_parsers/ap_anpa.py
@@ -117,7 +117,7 @@ class AP_ANPAFeedParser(ANPAFeedParser):
             html = item.get('body_html')
             if html:
                 parsed = parse_html(html, content='html')
-                for par in parsed.xpath('/html/div/child::*'):
+                for par in parsed.xpath('/html/body/child::*'):
                     if not par.text:
                         continue
                     city, source, the_rest = par.text.partition(' (AP) _ ')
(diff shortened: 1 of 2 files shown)


superdesk/io/feed_parsers/ap_anpa.py
```diff
@@ -117,7 +117,7 @@ def ap_derive_dateline(self, item):
             html = item.get('body_html')
             if html:
                 parsed = parse_html(html, content='html')
-                for par in parsed.xpath('/html/div/child::*'):
+                for par in parsed.xpath('/html/body/child::*'):
```
naspeh: is this change occasional?

naspeh: yep, I see, just misunderstood with subject (which says only about mapping)

mdhaman: PR was about mapping only but unit test failed. 
In version 1.7 xpath needs to `/html/body` but in version 1.8 it needs to be `/html/div`. It looks like changes were done after `root element` changes in version 1.8. (https://github.com/superdesk/superdesk-core/blob/master/superdesk/etree.py#L121)

naspeh: thanks for clarification, now I certainly understand this.

PR #933, opened 2017-05-16
Fix for legacy auth endpoint

naspeh:


naspeh:
Seems it can be closed

PR #934, opened 2017-05-16
doc(auth): add docs regarding google sign in

petrjasek:
SDESK-1268

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -63,3 +63,23 @@ Following functions can be used:
 
 .. automodule:: apps.auth
     :members:
+
+
+Google Authentication
+---------------------
+
+.. versionadded:: 1.8
+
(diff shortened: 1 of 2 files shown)


naspeh:
I don't like setting in `superdesk.config.js`, I like the way @jerome-poisson did for XMPP
- [with ActivatedService](https://github.com/superdesk/superdesk-core/blob/master/apps/auth/xmpp/auth.py#L84)

I would want to setup only settings in server part instead of two ones.
Also it's better to be consistent with approaches, now we have two ways and it's not good, I think.

Maybe it's better to use one endpoint for such things with all enabled types of authentication in response...

naspeh:
Probably we can improve it in another PR, because we also have the same thing for
```
settings.py: NO_TAKES
superdesk.config.js: features.noTakes
```
so would be good to have one way to get such settings from server (and define them only on server)...

petrjasek:
yep if that service is generic I would add it there, but it's only made for xmpp atm

PR #951, opened 2017-06-07
[SDESK-1315] (Fix):Expose expiry setting via superdesk api.

mdhaman:


mdhaman:
@petrjasek I merged this PR. can we have this pr in 1.8?

petrjasek:
yep np

PR #953, opened 2017-06-07
[INGEST] [FEEDING SERVICE] [FTP] better failure handling

jerome-poisson:
FTP feeding service was failing the whole ingestion if one file could
not be ingested e.g. because of parsing error. This commit fixes it.

In addition, the new move path on error feature is handled: if a file
can't be ingested, and the move is activated and move path error is set,
the file will be moved to the requested directory.

SDESK-1452

### DIFF
--- a/superdesk/io/feeding_services/ftp.py
+++ b/superdesk/io/feeding_services/ftp.py
@@ -31,6 +31,8 @@ except ImportError:
     from urlparse import urlparse
 
 logger = logging.getLogger(__name__)
+DEFAULT_SUCCESS_PATH = "_PROCESSED"
+DEFAULT_FAILURE_PATH = "_ERROR"
 
 
 class FTPFeedingService(FeedingService):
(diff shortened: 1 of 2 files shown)


superdesk/io/feeding_services/ftp.py
```diff
@@ -91,52 +115,52 @@ def _update(self, provider, update):
...
-                                reason=e))
+                            with open(local_file_path, 'xb') as f:
+                                try:
+                                    ftp.retrbinary('RETR %s' % filename, f.write)
+                                except ftplib.all_errors as ex:
+                                    os.remove(local_file_path)
+                                    logger.exception('Exception retrieving file from FTP server (%s)', filename)
+                                    continue
+                        except FileExistsError:
+                            logger.error('Exception retrieving from FTP server, file already exists (%s)',
```
petrjasek: imo this shouldn't be an error, warning maybe

superdesk/io/feeding_services/ftp.py
```diff
@@ -79,8 +97,14 @@ def _update(self, provider, update):
             except KeyError:
                 logger.warning('missing move_path')
                 move_dest_path = None
+            try:
+                move_dest_path_error = os.path.join(config.get('path', ''), config['move_path_error'])
```
petrjasek: maybe we can use default values for those move paths to be consistent with file ingest,
there is `_PROCESSED` and `_ERROR` folder created

jerome-poisson: the trick is that empty value is used as a way to say "don't move files on error". So we can't use default (or we need an other way to disable on error).
Also the directory needs to exists, if we use default value I guess we need to create the directory, is it a good thing?

superdesk/io/feeding_services/ftp.py
jerome-poisson: OK

jerome-poisson:
I've changed logger.error to warning as requested by @petrjasek 

I've also modified **continue** on **RETR** and if a local file already exists to raise an error instead. This way it will trigger the _move on error_ code.

See my comment for default `_PROCESSED` and `_ERROR`, as I'm not convinced it would be a good thing.

superdesk/io/feeding_services/ftp.py
petrjasek: what would be the usecase to keep those files there on error?

jerome-poisson: it has not been requested, I just thought one may want to move on success but not on error, but I may be wrong.

jerome-poisson: @petrjasek so what do you think? I can remove the ability to disable _move on error_.
If I add default dir, I'll add mkdir if the dir doesn't exist, I think it would make sense in this case.

So do I let as it is or add default dir and dir creation if it doesn't exist? Either way is OK to me.
@karelpetrak your input is welcome here.

petrjasek: I would prefer the default and create that folder if it's not there, but lets see what karel thinks, or @hlmnrmr ?

jerome-poisson: @karelpetrak @hlmnrmr so what do we do?

jerome-poisson: I've pinged @karelpetrak about this, he rather like the default directory to be consistent with files, so same opinion as @petrjasek . So I'll go this way and change the P.R. later today.

superdesk/io/feeding_services/ftp.py
```diff
@@ -69,74 +71,120 @@ def _test(self, provider):
...
+
+    def _create_if_missing(self, ftp, path):
+        """Check if a dir exists, and create it else
+
+        :param ftp: FTP instance to use
+        :type ftp: ftplib.FTP
+        :param src: dir path to check
+        """
+        try:
+            next(ftp.mlsd(path))
```
petrjasek: I'm not sure if it's the best way to check it, when there are lots of files it could take some time, or even timeout.
What about trying to create that dir and ignore any conflict error?

petrjasek: or it could try to `ftp.cwd()` there

jerome-poisson: @petrjasek I was hesitating for the command to use and for whatever reason I've missed cwd which is definitely a better choice, I'm changing it.

jerome-poisson: @petrjasek I remember now, I've not uses cwd to avoid changing directory, because we then have to come back. But big listing can a worst issue indeed, specially in SD case.

jerome-poisson: @petrjasek creating the dir directly is not possible because we don't have conflict error, but more generic ones, so we can't distinguish a conflict (which would be ignored) for existing path from a setting/permission error (which should make the moving fail).

An error code is returned which can probably be used, but it would need extensive test to be sure that most FTP server return the same one, it's a lot of complication when the current workflow works OK.

superdesk/io/feeding_services/ftp.py
```diff
@@ -69,74 +71,124 @@ def _test(self, provider):
...
+        base_path = ftp.pwd()
+        try:
+            ftp.cwd(path)
+        except ftplib.all_errors:
+            # path probably doesn't exist
+            # catching all_errors is a bit overkill,
+            # but ftplib doesn't really have precise error
+            # for missing directory
+            ftp.mkd(path)
+        except StopIteration:
```
petrjasek: I think this part is not needed anymore

PR #956, opened 2017-06-12
fix(feature media): Fix publish when feature media and/or media descr…

ioanpocol:
…iption are required

### DIFF
--- a/apps/validate/validate.py
+++ b/apps/validate/validate.py
@@ -159,6 +159,17 @@ class ValidateService(superdesk.Service):
                     # fails for json fields like subject, genre
                     pass
 
+    def _process_media(self, doc):
+        """If feature media is required it should be present for validation on doc not only on associations
+        If media_description is required it should be present for validation on doc not only on
+        associations->featuremedia
+        :param doc: Article to be validated
+        """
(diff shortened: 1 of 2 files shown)


apps/validate/validate.py
```diff
@@ -159,6 +159,17 @@ def _sanitize_fields(self, doc, validator):
...
+    def _process_media(self, doc):
+        """If feature media is required it should be present for validation on doc not only on associations
+        If media_description is required it should be present for validation on doc not only on
+        associations->featuremedia
+        :param doc: Article to be validated
+        """
+        if 'associations' in doc and 'featuremedia' in doc['associations']:
+            doc['feature_media'] = doc['associations']['featuremedia']
+            if 'description_text' in doc['associations']['featuremedia']:
+                doc['media_description'] = doc['associations']['featuremedia']['description_text']
```
petrjasek: shouldn't it be stored there then?

ioanpocol: I think no, this is just for validation and to keep the validation simple. There is a well defined logic around associations so I don't know if we have to change it.

petrjasek: ok

ioanpocol:
done

PR #964, opened 2017-06-21
[SDESK-1501] (fix):Fixing item history for scheduled items.

mdhaman:
- We increment the version on publishing scheduled item and it causes item history to be out of sync with `publish_queue`.


petrjasek:
pls rebase to get travis fixed

PR #981, opened 2017-07-11
feat(vocabularies): Activate DELETE for vocabularies endpoint

ioanpocol:


superdesk/vocabularies/vocabularies.py
```diff
@@ -73,7 +73,6 @@ class VocabulariesResource(Resource):
         },
         'service': {
             'type': 'dict',
-            'default': {'all': True},
```
petrjasek: why is this removed? @ioanpocol

ioanpocol: Duplicates genre and other dropdowns that had default vocabulary. If the service is present means that it is a custom vocabulary.

petrjasek: ok, this will need some ui changes so that new vocabularies will show up in profile management

PR #991, opened 2017-07-14
[storage] add file extension to url for local storage

jerome-poisson:
File extension is needed in local storage to avoid trouble with
Wordpress plugin.
As it is not possible to change ID (Mongo expect a unique ObjectID with 24 hex bytes), this patch work around id by adding the file extension to the URL, and ignoring it when getting the file.

media_id has also been removed from SuperdeskGridFSMediaStorage as it is
not used anymore.

SDFID-47

### DIFF
--- a/superdesk/storage/desk_media_storage.py
+++ b/superdesk/storage/desk_media_storage.py
@@ -10,10 +10,10 @@
 
 import logging
 import json
+import mimetypes
 import bson
 import gridfs
 from eve.io.mongo.media import GridFSMediaStorage
-from superdesk.utils import sha
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


jerome-poisson:
Here is a summary of the situation:

1. https://dev.sourcefabric.org/browse/SDFID-47 has been created, the issue request that a file extension is added for local storage
2. I've merged a fix (https://github.com/superdesk/superdesk-core/pull/980), after testing it OK
3. @ioanpocol  tells me that the fix is not working on Fidelity test instance with AWS storage, so we revert the fix immediately (https://github.com/superdesk/superdesk-core/pull/984) for further testing, test instance works after the revert
4. I ask AWS test credentials and try the patch. It seems to be working OK with AWS and local storage
5. after a call with Ingo we decide to create a test instance. So I've made a new P.R. and asked Infrastructure to activate AWS there, this P.R. is the current one (https://github.com/superdesk/superdesk-core/pull/991). I also make a P.R. without AWS, this is https://github.com/superdesk/superdesk-core/pull/992 and is for test only, not to be merged.
6. we check with Ingo and set up Wordpress plugin, everything works fine for either https://github.com/superdesk/superdesk-core/pull/991 and https://github.com/superdesk/superdesk-core/pull/992
7. so the patch seems to be working, I remove _wip_ label for https://github.com/superdesk/superdesk-core/pull/991 and let both instances for further testing. If everything is alright, https://github.com/superdesk/superdesk-core/pull/991 can be merged and https://github.com/superdesk/superdesk-core/pull/992 can be closed

Further remarks:

- the patch should not affect AWS (it acting on **/upload-raw/** URL and on local storage, i.e. _desk_media_storage.py_), so I was surprised of the issue on AWS. I suspect the trouble can come from either an other patch or a setting somewhere
- the settings of the P.R. with AWS activated can be checked on https://dev.sourcefabric.org/browse/IN-2947

PR #1006, opened 2017-08-05
SDESK-1631 Create Concept Items type person

mugurrus:


apps/concept_items/resource.py
```diff
@@ -0,0 +1,44 @@
...
+
+from superdesk.resource import Resource
+
+
+CONCEPT_ITEMS_PRIVILEGE = 'concept_items'
+
+
+class ConceptItemResource(Resource):
+    schema = {
+        'first_name': {
```
petrjasek: I think we should have just `name` otherwise this doesn't work for other types of concepts

apps/concept_items/resource.py
```diff
@@ -0,0 +1,44 @@
...
+            'type': 'string',
+            'required': True
+        },
+        'by_line': {
+            'type': 'string'
+        },
+        'email': {
+            'type': 'string'
+        },
+        'biography': {
```
petrjasek: this should be `description` to work for other types

apps/concept_items/resource.py
```diff
@@ -0,0 +1,44 @@
...
+            'type': 'string'
+        },
+        'role': {
+            'type': 'string'
```
petrjasek: this should be a list

petrjasek:
obsolete

PR #1010, opened 2017-08-16
[users] added "author" to user_type

jerome-poisson:
SDESK-1754


jerome-poisson:
Closed as the feature is done in a different way (using a `is_author` flag instead of new type), cf. https://github.com/superdesk/superdesk-core/pull/1012

PR #1011, opened 2017-08-17
[SDESK-1798] Refactor purging the audit collection

marwoodandrew:


superdesk/data_updates/00003_20170814-114652_audit.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class DataUpdate(DataUpdate):
+
+    resource = 'audit'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        for audit in mongodb_collection.find({}):
+            audit_id = get_resource_service(self.resource)._extract_doc_id(audit.get('extra'))
+            print(mongodb_collection.update({'_id': audit.get(config.ID_FIELD)},
```
petrjasek: imo would make sense to add an index on audit_id once it gets populated

marwoodandrew: @petrjasek Yes we've been tuning it on our UAT system, It's just a painfully slow process! 

Do you think the index should be coded in the resource. or just create it in Mongo externally?

petrjasek: I would rather create it here once the migration is done

marwoodandrew: Yep, no problem!

PR #1012, opened 2017-08-18
[users] added "is_author" flag, and biography/job_title fields

jerome-poisson:
SDESK-1754


jerome-poisson:
this second version of  SDESK-1754 fix, modified after discussions with @hlmnrmr

PR #1013, opened 2017-08-20
Changes for content-api user schema

akintolga:
- Introducing basic authentication as initial set up
- Modified users end point schema to let public users to sign up
- Renamed clients endpoint to companies

### DIFF
--- a/content_api/users/resource.py
+++ b/content_api/users/resource.py
@@ -9,6 +9,7 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 from superdesk.resource import Resource
+from content_api import MONGO_PREFIX
 
 
 class UsersResource(Resource):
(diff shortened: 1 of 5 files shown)


content_api/users/resource.py
```diff
@@ -17,18 +18,50 @@ class UsersResource(Resource):
...
+            'nullable': True
+        },
+        'signup_details': {
+            'type': 'dict'
         },
-        'client': Resource.rel('clients')
+        'country': {
+            'type': 'string'
+        },
+        'company': Resource.rel('roles', embeddable=True, required=False),
```
petrjasek: here it should be rel to `companies` I guess

content_api/auth/service.py
```diff
@@ -0,0 +1,54 @@
...
+from flask import g, current_app as app
+from apps.auth.service import AuthService
+from superdesk import get_resource_service
+from apps.auth.errors import CredentialsAuthError
+
+from superdesk.utc import utcnow
+from datetime import timedelta
+
+
+class ContentApiAuthService(AuthService):
```
petrjasek: I'm not sure if we need such service, users should be authenticated using sessions, and for api we should use some tokens that users could generate, but wouldn't authenticate on the api with email/password

akintolga:
For some reason the checks stalled for this PR

PR #1018, opened 2017-08-27
content api companies enpoint

akintolga:
- Renamed the clients as companies
- Created company token end point

### DIFF
--- a/tests/content_api_test.py
+++ b/tests/content_api_test.py
@@ -27,6 +27,16 @@ class ContentAPITestCase(TestCase):
         self.capi.testing = True
         self.subscriber = {'_id': 'sub1'}
 
+    def _auth_headers(self, sub=None):
+        if sub is None:
+            sub = self.subscriber
+        service = superdesk.get_resource_service('company_token')
+        payload = {'company': sub.get('_id')}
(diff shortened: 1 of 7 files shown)


tests/content_api_test.py
```diff
@@ -27,6 +27,16 @@ def setUp(self):
         self.capi.testing = True
         self.subscriber = {'_id': 'sub1'}
 
+    def _auth_headers(self, sub=None):
+        if sub is None:
+            sub = self.subscriber
+        service = superdesk.get_resource_service('company_token')
+        payload = {'company': sub.get('_id')}
+        service.create([payload])
+        token = payload.get('_id')
```
jerome-poisson: either `_id` key can be non existent and the following line will fail (string + None), or it will always exist and a get is not needed here.

akintolga: `_id` cannot be non-existent because when we call create function of the `company_token` service we create a new recor in `company_token` collection. Every record represents a token and _id contains the token value. We need to get it as well as we'll use the token value in the request header to get items.

jerome-poisson: so can you change it to `token = payload['_id']` then?

PR #1020, opened 2017-09-01
added "authors" content type

jerome-poisson:
author can now be used in items.

Added "author_roles" in controlled vocabularies.
users endpoint now allow filtering by authors using "is_author=1"
parameter.

SDESK-1703 SDESK-1800 SDESK-1801

### DIFF
--- a/superdesk/metadata/item.py
+++ b/superdesk/metadata/item.py
@@ -285,6 +285,19 @@ metadata_schema = {
         'type': 'string',
         'nullable': True,
     },
+    'authors': {
+        'type': 'list',
+        'nullable': True,
+        'mapping': {
+            'type': 'object',
+            'properties': {
+                'parent': not_analyzed,
+                'name': not_analyzed,
+                'role': not_analyzed
+            }
+
+        }
+    },
     'description_text': {
         'type': 'string',
         'nullable': True
--- a/superdesk/users/services.py
+++ b/superdesk/users/services.py
@@ -346,6 +346,18 @@ class UsersService(BaseService):
     def set_privileges(self, user, role):
         user['active_privileges'] = get_privileges(user, role)
 
+    def get(self, req, lookup):
+        try:
+            is_author = req.args['is_author']
+        except (AttributeError, TypeError, KeyError):
+            pass
+        else:
+            if is_author in ('0', '1'):
+                lookup['is_author'] = bool(int(is_author))
+            else:
+                logger.warn('bad value of is_author argument ({value})'.format(value=is_author))
+        return super().get(req, lookup)
+
     def get_users_by_user_type(self, user_type='user'):
         return list(self.get(req=None, lookup={'user_type': user_type}))
 
--- a/apps/comments/user_mentions.py
+++ b/apps/comments/user_mentions.py
@@ -23,8 +23,8 @@ def get_mentions(text):
     Returns the names of desks and users from a given text
     User names starts with '@' and desk names starts with '#'
     """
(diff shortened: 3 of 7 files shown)


superdesk/users/services.py
```diff
@@ -346,6 +346,11 @@ def get_role(self, user):
     def set_privileges(self, user, role):
         user['active_privileges'] = get_privileges(user, role)
 
+    def get(self, req, lookup):
+        if req is not None and 'is_author' in req.args and req.args['is_author'] in ('0','1'):
+            lookup['is_author'] = bool(int(req.args['is_author']))
```
petrjasek: this assumes `is_author` is present on all users, but it won't return existing users without that flag.

jerome-poisson: Hi, there is a default value of "True" for this field, so it should work.

petrjasek: yep but would it work for users created before?

jerome-poisson: indeed it would not work in this case. But I think in this case it should be handled by an update script. Do you think it should be handled differently?

petrjasek: yep that's another option, you can either do the lookup like `{'$eq': {'is_author': True}}` / {`'$neq': {'is_author': True}}` or add the update script

jerome-poisson: after talking with @hlmnrmr we decided that I'll go with the upgrade script

naspeh:
@jerome-poisson I just did checkout for this PR and investigated a bit, you did something wrong with mapping
```
root@sds-naspeh--check-nose:/opt/superdesk/server-core# git diff superdesk/metadata/item.py
diff --git a/superdesk/metadata/item.py b/superdesk/metadata/item.py
index d4c5453..9c04831 100644
--- a/superdesk/metadata/item.py
+++ b/superdesk/metadata/item.py
@@ -291,11 +291,7 @@ metadata_schema = {
         'mapping': {
             'type': 'object',
             'properties': {
-                '_id': {
-                    'type': 'list',
-                    'minlength': 2,
-                    'maxlength': 2,
-                    'required': True},
+                '_id': not_analyzed,
                 'parent': not_analyzed,
                 'name': not_analyzed,
                 'role': not_analyzed
```
this change can fix `nosetests`
```
root@sds-naspeh--check-nose:/opt/superdesk/server-core# nosetests 
...................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 867 tests in 365.864s

OK
```
So you need to look into this mapping again.

jerome-poisson:
hi @naspeh thanks for that. It's weird because _id is indeed a list of 2 elements here (parent's id + role). I can change that to a string concatenating the 2 values.

jerome-poisson:
Ooops sorry, I though it was OK but I was on a older page, not refreshed, fixing it now.

superdesk/metadata/item.py
```diff
@@ -285,6 +285,19 @@
         'type': 'string',
         'nullable': True,
     },
+    'authors': {
+        'type': 'list',
+        'nullable': True,
+        'mapping': {
+            'type': 'object',
+            'properties': {
+                'parent': not_analyzed,
```
petrjasek: who is authors parent?

jerome-poisson: it's the user (user id). Parent was being used in client for subjects to have the tree, so I took the same mechanism (parent are users, children are roles).

PR #1021, opened 2017-09-02
SDESK-1816 As a admin I want to have a setting for "Publish Queue" expiry

mugurrus:


apps/archive/commands.py
```diff
@@ -48,6 +49,17 @@ def run(self):
...
+    def _remove_expired_publish_queue_items(self):
+        system_settings_service = get_resource_service('system_settings')
+        expire_interval = system_settings_service.find_one(req=None, name='publish_queue_expire_interval')
+        if not expire_interval:
+            return
+
+        expire_time = utcnow() - system_settings_service.validate_timedelta(expire_interval)
+        logger.info('{} Removing publish queue items created before {}'.format(self.log_msg, str(expire_time)))
+
+        get_resource_service('publish_queue').delete({'_created': {'$lte': expire_time}})
```
mdhaman: Can you define the index on the `_created` field for the `publish_queue` resource?

superdesk/system_settings/resource.py
```diff
@@ -0,0 +1,31 @@
...
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+from superdesk.resource import Resource
+
+
+class SystemSettingsResource(Resource):
+    schema = {
+        'name': {
```
petrjasek: I would rather use string `_id` instead of name, it would make it use an index

apps/archive/commands.py
petrjasek: or you can use something like `{'_id': {'$lte': ObjectId.from_datetime(expire_time)}}`

naspeh:
We have https://github.com/superdesk/superdesk-core/blob/master/superdesk/server_config.py
Probably `superdesk/system_settings` should go there...

There are too many endpoints for config related things:
- https://github.com/superdesk/superdesk-core/blob/master/superdesk/server_config.py
- https://github.com/superdesk/superdesk-core/blob/master/apps/client_config.py

I've even created ticket to tidy up this stuff: https://dev.sourcefabric.org/browse/SDESK-1883 (probably I should postpone it)

apps/archive/commands.py
mdhaman: I thought that index might not exists but we already have index on `_created`.

mugurrus:
@naspeh The server_config module has no validation for settings; all values must be dict in server_config, which is not acceptable. I think what I started (system_settings) is more flexible and has more features. I can refactor all configuration endpoints into a single on if it's OK with you.

petrjasek:
actually I'm not sure about this setting part, imo it should be another setting in default settings like `CONTENT_EXPIRY_MINUTES`, not sure why we would need a UI for this one?

superdesk/system_settings/service.py
```diff
@@ -0,0 +1,59 @@
...
+        updated = copy.deepcopy(original)
+        updated.update(updates)
+        self._validate_value(updated)
+
+    def on_replace(self, document, original):
+        self._validate_value(document)
+
+    def _validate_value(self, doc):
+        if doc['type'] == 'string' and type(doc['value'] != 'string'):
+            SuperdeskApiError.badRequestError('Property \'' + doc[config.ID_FIELD] + '\' must be a string.')
```
naspeh: there is no `raise` here

superdesk/system_settings/service.py
```diff
@@ -0,0 +1,59 @@
...
+        if doc['type'] == 'integer' and not doc['value'].isdigit():
+            SuperdeskApiError.badRequestError('Property \'' + doc[config.ID_FIELD] + '\' must be an integer.')
```
naspeh: also there is no `raise` here

superdesk/system_settings/service.py
naspeh: Probably additional test case could catch up this...

mugurrus:
Because the requirement is for the admin user to be able to change it.

naspeh:
@mugurrus I'm ok to refactor these settings modules later, this why I postponed [SDESK-1883](https://dev.sourcefabric.org/browse/SDESK-1883)

But probably UI is not really needed for this task, as metioned @petrjasek, admin can change just config file in such case, no?

PR #1024, opened 2017-09-05
Remove SERVER_NAME and URL_PROTOCOL settings

naspeh:
- they outdated since [eve==0.5](https://github.com/pyeve/eve/blob/98a1febdae04312e9f4be96e3b2394ca2efb2315/eve/default_settings.py#L33-L35)
- SERVER_NAME is used in Flask, but seems we don't need such usage, so
  using SERVER_DOMAIN instead (it's already in use in some places)
- possibility to specify URL_PREFIX and CONTENTAPI_URL_PREFIX from env
  variables, can be useful in some deployment schemes

[SDESK-1892](https://dev.sourcefabric.org/browse/SDESK-1892)

### DIFF
--- a/superdesk/upload.py
+++ b/superdesk/upload.py
@@ -17,7 +17,7 @@ from superdesk.errors import SuperdeskApiError
 from werkzeug.wsgi import wrap_file
 from .resource import Resource
 from .services import BaseService
-from flask import url_for, request, current_app as app, redirect
+from flask import request, current_app as app, redirect
 from superdesk.media.renditions import generate_renditions, delete_file_on_error
 from superdesk.media.media_operations import (
     download_file_from_url, download_file_from_encoded_str,
@@ -67,10 +67,8 @@ def url_for_media(media_id, mimetype=None):
 
 
 def upload_url(media_id, view='upload_raw.get_upload_as_data_uri'):
-    media_prefix = app.config.get('MEDIA_PREFIX')
-    if media_prefix:
-        return '%s/%s' % (media_prefix.rstrip('/'), media_id)
-    return url_for(view, media_id=media_id, _external=True)
+    media_prefix = app.config.get('MEDIA_PREFIX').rstrip('/')
+    return '%s/%s' % (media_prefix, media_id)
 
 
 def init_app(app):
--- a/content_api/app/settings.py
+++ b/content_api/app/settings.py
@@ -39,8 +39,9 @@ CONTENTAPI_DOMAIN = {}
 
 # NOTE: no trailing slash for the CONTENTAPI_URL setting!
 CONTENTAPI_URL = env('CONTENTAPI_URL', 'http://localhost:5400')
+MEDIA_PREFIX = '%s/assets' % CONTENTAPI_URL.rstrip('/')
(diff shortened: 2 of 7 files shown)


petrjasek:
I think we might need `SERVER_NAME` still, not sure if we have always request context when generating urls (like when it's publishing something), but not sure

naspeh:
ok, I'll replace `SERVER_DOMAIN` with `SERVER_NAME` then

naspeh:
nope, when I set `SERVER_NAME` I must do request from the same domain, for example
- I run `honcho` with `port=5000` and set `SERVER_NAME=localhost`
- I can't access urls from `localhost:5000`
but without `SERVER_NAME` I can access both... And I need this ability to simplify deployment when proxing with `nginx`

@petrjasek what do you think?

P.S. I can't understand who do this match `flask`, `werkzeug`, or `gunicorn` %) and how avoid this

naspeh:
anyway it's bad idea to generate urls without request context, for that we have `SUPERDESK_URL` setting

naspeh:
@petrjasek I just updated the code
- removed `SERVER_NAME` from `TestCase`
- removed usage of `app.url_for` from `superdesk.upload.upload_url`, it uses in some commands

so now it should be good

superdesk/upload.py
```diff
@@ -67,10 +67,8 @@ def url_for_media(media_id, mimetype=None):
 
 
 def upload_url(media_id, view='upload_raw.get_upload_as_data_uri'):
-    media_prefix = app.config.get('MEDIA_PREFIX')
-    if media_prefix:
-        return '%s/%s' % (media_prefix.rstrip('/'), media_id)
-    return url_for(view, media_id=media_id, _external=True)
+    media_prefix = app.config.get('MEDIA_PREFIX').rstrip('/')
+    return '%s/%s' % (media_prefix, media_id)
```
petrjasek: it's not the same - `url_for` with `_external=True` will give you full url, while you're returning relative one, but on the client relative urls won't work because it's different domain serving api and ui

naspeh:
```
MEDIA_PREFIX = env('MEDIA_PREFIX', '%s/upload-raw' % SERVER_URL.rstrip('/'))
```
so `MEDIA_PREFIX` has absolute part from `SUPERDESK_URL`, and this is the same as `_external=True`

petrjasek: I see. should be fine then

naspeh:
I found where this server name matching goes: [werkzeug/routing.py#L1312-L1322](https://github.com/pallets/werkzeug/blob/40bfd13b4383512160a139aea3721c3746a4d7ae/werkzeug/routing.py#L1312-L1322)
```
            if cur_server_name[offset:] != real_server_name:
                # This can happen even with valid configs if the server was
                # accesssed directly by IP address under some situations.
                # Instead of raising an exception like in Werkzeug 0.7 or
                # earlier we go by an invalid subdomain which will result
                # in a 404 error on matching.
                subdomain = '<invalid>'
```

So this `SERVER_NAME` settings should be really avoided, otherwise you need to fix hostname when app behind a proxy... %)

PR #1025, opened 2017-09-06
SDESK-1816 As a admin I want to have a setting for "Publish Queue" expiry

mugurrus:


apps/archive/commands.py
```diff
@@ -48,6 +52,14 @@ def run(self):
 
         remove_locks()
 
+    def _remove_expired_publish_queue_items(self):
+        expire_interval = app.config.get('PUBLISH_QUEUE_EXPIRY_TIME', 0)
+        if expire_interval:
+            expire_time = utcnow() - timedelta(seconds=expire_interval)
```
petrjasek: imo would be good to be consistent with the other options so the config should be in minutes

petrjasek: like `PUBLISH_QUEUE_EXPIRY_MINUTES`

mugurrus: Then I would have to set the sleep time for the behave test to 61 seconds, otherwise it wouldn't work.

petrjasek: I think you can set an _id manually when you post to publish queue

mugurrus: But this has nothing to do with _id

naspeh: I agreed with @petrjasek regarding the name

Probably you can use `PUBLISH_QUEUE_EXPIRY_MINUTES = -1` for tests like https://github.com/superdesk/superdesk-core/blob/master/tests/audit_purge_test.py#L16

naspeh: using this [superdesk.tests.steps.given_config_update](https://github.com/superdesk/superdesk-core/blob/master/superdesk/tests/steps.py#L376)

petrjasek: if the _id is older it will pick it or am I missing something?

superdesk/default_settings.py
```diff
@@ -618,3 +618,5 @@ def local_to_utc_hour(hour):
 # SAML Auth settings
 SAML_PATH = env('SAML_PATH')
 SAML_LABEL = env('SAML_LABEL', 'Single Sign On')
+
+PUBLISH_QUEUE_EXPIRY_TIME = env('PUBLISH_QUEUE_EXPIRY_TIME', 864000)
```
petrjasek: imo it should be 0 by default

PR #1027, opened 2017-09-10
[SDAN-52] - User validation and password change token

akintolga:
- Fields needed to validate user email and password change requests


petrjasek:
I'm not sure if single token will be enough, but we can give it a try.

akintolga:
I am not also 100% sure but for now let's try as you say.

PR #1030, opened 2017-09-12
SDESK-1919 Superdesk ninjs extension schema is incorrect

mugurrus:


docs/superdesk-ninjs-schema.json
```diff
@@ -15,11 +15,11 @@
 			"type" : "string"
   		}
 	},
+	"required": ["guid"],
```
mdhaman: Do we need to change this line as well https://github.com/superdesk/superdesk-core/blob/master/docs/superdesk-ninjs-schema.json#L2 ?

mugurrus: That contains the description of the keywords used in the schema. We added some custom keywords so we may create our own document but I think this should be done in another task.

mdhaman: that's fine if you want to do in a separate task.

PR #1032, opened 2017-09-13
Fix Publish Queue expiry

naspeh:
[SDESK-1816](https://dev.sourcefabric.org/browse/SDESK-1816)

### DIFF
--- a/apps/archive/commands.py
+++ b/apps/archive/commands.py
@@ -8,10 +8,10 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-
-from flask import current_app as app
-import superdesk
+import functools as ft
 import logging
+import superdesk
+from flask import current_app as app
 from eve.utils import config, ParsedRequest
 from copy import deepcopy
 from apps.packages import PackageService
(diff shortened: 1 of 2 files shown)


apps/archive/commands.py
```diff
@@ -28,6 +28,16 @@
 logger = logging.getLogger(__name__)
 
 
+def log_exeption(fn):
```
petrjasek: I don't get why it's needed, there is no logging otherwise?

petrjasek: should be logged here https://github.com/superdesk/superdesk-core/blob/master/superdesk/__init__.py#L57-L64

naspeh: we have two calls
```
+        self._remove_expired_publish_queue_items()
 +        self._remove_expired_items(now)
```
in previous version if the first fails second one never call, so I wrapped these function to be sure that second function will be called

petrjasek: I see, would be good imo to put this into the comment

naspeh: comment is here
```diff
+        # next functions wrapped with log_exeption, so no try there
```

apps/archive/commands.py
```diff
@@ -40,18 +51,20 @@ def run(self):
...
-        try:
-            logger.info('{} Removing expired content for expiry.'.format(self.log_msg))
-            self._remove_expired_items(now)
-            self._remove_expired_publish_queue_items()
-        finally:
-            unlock(lock_name)
+
+        logger.info('{} Removing expired content for expiry.'.format(self.log_msg))
+        # both functions should be called, even the first one throw exception,
+        # so they are wrapped with log_exeption
```
naspeh: here is I updated comment with better description

PR #1033, opened 2017-09-13
[FORMATTER] [NINJS] output authors is SD extension "authors" field

jerome-poisson:
Authors is added to NINJS output, and superdesk NINJS extension schema
has been updated accordingly.

SDESK-1928


jerome-poisson:
Putting `WIP` label as we are not sure yet about what to do with job title, cf. https://dev.sourcefabric.org/browse/SDESK-1930)

PR #1035, opened 2017-09-20
[SDESK-1912] Media Contacts

marwoodandrew:
Media Contacts is implemented as a simple instance of a Knowledge Base/Concept.
A knowledge base is simply an entry in the knowledge base vocabulary.
The concept collection can be extended to include additional concepts.


marwoodandrew:
Wasn't working out, there were problems calling it from the front end. Will go back to a single use contacts endpoint!

PR #1037, opened 2017-09-20
[FORMATTER] [NINJS]: added time to read to output

jerome-poisson:
time to read (estimated number of minute to read body of item) is added
in readTime property. The property has been added to Superdesk ninjs
schema too.

SDFID-118

### DIFF
--- a/superdesk/publish/formatters/ninjs_formatter.py
+++ b/superdesk/publish/formatters/ninjs_formatter.py
@@ -40,7 +40,7 @@ from superdesk.metadata.packages import RESIDREF, GROUP_ID, GROUPS, ROOT_GROUP, 
 from superdesk.utils import json_serialize_datetime_objectId
 from superdesk.media.renditions import get_renditions_spec
 from apps.archive.common import get_utc_schedule
-from superdesk.etree import get_text
+from superdesk import text_utils
 
 
 def filter_empty_vals(data):
@@ -173,7 +173,7 @@ class NINJSFormatter(Formatter):
         if article.get('abstract'):
             abstract = article.get('abstract', '')
             ninjs['description_html'] = abstract
-            ninjs['description_text'] = get_text(abstract)
+            ninjs['description_text'] = text_utils.get_text(abstract)
         elif article.get('description_text'):
             ninjs['description_text'] = article.get('description_text')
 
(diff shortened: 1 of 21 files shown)


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -200,6 +200,18 @@ def _transform_to_ninjs(self, article, subscriber, recursive=True):
         if article.get('attachments'):
             ninjs['attachments'] = self._format_attachments(article)
 
+        if 'body_html' in ninjs or 'body_text' in ninjs:
+            if 'body_html' in ninjs:
+                word_count = etree.get_word_count(ninjs.get('body_html'))
+            else:
+                word_count = etree.get_text_word_count(ninjs['body_text'])
+            reading_time_float = word_count / 250
```
petrjasek: imo this should go into a function with some docs explaining the formula

jerome-poisson: @petrjasek : OK, do you know about a module where I can put that ? Etree seems not adapted since it's only from word count (or I can add get_reading_time in etree, but it would need to count the word several times if word count is also needed).

petrjasek: right, not sure why we have `get_word/text_count` in etree anyhow, so imo all those should go into textutils or something

petrjasek:
can you add some info about that renaming into changelog? it's a breaking change

jerome-poisson:
@petrjasek sure

jerome-poisson:
@petrjasek also do you think text_utils is good name while methods are for text and html? Not sure if there is any better.

petrjasek:
I think the naming is good

jerome-poisson:
@petrjasek done

jerome-poisson:
I've put in 1.8.5 marked as not released yet, is it good that way?

naspeh:
P.S. We have some text/string related functions in [superdesk/utils.py](https://github.com/superdesk/superdesk-core/blob/master/superdesk/utils.py)

naspeh:
what about package `superdesk.utils` and separate module `superdesk.utils.text` and add
```
# utils.__init__.py
from .text import *
from .other import *
```

jerome-poisson:
@naspeh checking `utils.py`, it's more generic methods, only `plaintext_filter` would make sense in `text_utils` IMHO.

Also renaming/moving a method imply changing many files and fixing tests, and potentially break things, if we have to rename things more deeply, we would need a ticket and have a more careful look at all the modules. It would be a good thing, but a bit time consuming.

So for my point of view, it's good enough for now.

naspeh:
I'm suggesting creating package from `superdesk.utils` module, but all import stay `from superdesk.utils import plaintext_filter` as is, whatever you put it under `utils` package... just import all stuff from underlining modules in `utils.__init__`...

Anyway, I'm OK with that PR (I already approved it).

jerome-poisson:
@naspeh oh OK, may be a good idea indeed. But still I think this should be in a separate ticket, it's refactoring and not really linked to this `time to read` P.R.

PR #1040, opened 2017-09-25
[SDESK-1957] Simplify email validator

naspeh:
@petrjasek we need new release for liveblog, they still use `1.6`


naspeh:
I cherry-picked `pydocstyle` fix also

PR #1046, opened 2017-10-09
[SDESK-1998](feat): Adding a reference to assignment to superdesk item.

mdhaman:
required by https://github.com/superdesk/superdesk-planning/pull/484


petrjasek:
seems like there was some change in style checking packages, you can put `D107` into setup.cfg ignore section, or put there those comments ;)

mdhaman:
@petrjasek ignore done...

PR #1047, opened 2017-10-09
SDESK-1965 As a user I want to use custom CVs when I create a "filter condition"

mugurrus:


apps/content_filters/filter_condition/filter_condition_resource.py
```diff
@@ -75,3 +76,7 @@ class FilterConditionResource(Resource):
     privileges = {'POST': 'content_filters',
                   'PATCH': 'content_filters',
                   'DELETE': 'content_filters'}
+
+    def on_init(self):
+        for vocabulary in get_resource_service('vocabularies').get(req=None, lookup={}):
+            self.schema['field']['allowed'].append(vocabulary[config.ID_FIELD])
```
petrjasek: not sure we should do this on init, it would need a restart when there is a new field added later

mugurrus: Yep, wanted to add this but I forgot. Please check the latest update, now the filter allowed values are updated on vocabulary create/delete.

apps/content_filters/filter_condition/filter_condition_field.py
```diff
@@ -205,3 +215,29 @@ def get_elastic_query(self):
...
+
+
+class FilterConditionControlledVocabularyField(FilterConditionField):
+    def __init__(self, field):
+        self.field = type('_ControlledVocabulary', (object,), {'name': field})()
+        self.entity_name = 'subject.qcode'
+        self.field_type = str
+
+    def is_in_article(self, article):
+        return any([self.field.name == subject['scheme'] for subject in article.get('subject', [])])
```
petrjasek: afaik there is no scheme set for default subjectcodes, so this would be keyerror

apps/content_filters/filter_condition/filter_condition_parameters.py
```diff
@@ -26,85 +26,99 @@ class FilterConditionParametersResource(Resource):
...
+                   'operators': ['eq', 'ne'],
+                   'values': values['embargo'],
+                   'value_field': 'name'
+                   }]
+        fields.extend(self._get_vocabulary_fields(values))
+        return ListCursor(fields)
+
+    def _get_vocabulary_fields(self, values):
+        for vocabulary in get_resource_service('vocabularies').get(req=None, lookup=None):
+            if vocabulary[config.ID_FIELD] not in values and vocabulary['type'] == 'manageable':
```
petrjasek: imo there should be some filtering, otherwise this gets also crop sizesa, rights info etc

apps/content_filters/filter_condition/filter_condition_resource.py
petrjasek: I think the syncing won't work because it will only sync it in the process which did the changes, but not for other processes. would be probably better to check the field in the service

apps/content_filters/filter_condition/filter_condition_parameters.py
mugurrus: Added a filter, please review

apps/content_filters/filter_condition/filter_condition_resource.py
```diff
@@ -75,3 +76,17 @@ class FilterConditionResource(Resource):
     privileges = {'POST': 'content_filters',
                   'PATCH': 'content_filters',
                   'DELETE': 'content_filters'}
+
+    def on_init(self):
+        excluded_vocabularies = [vocabulary.strip() for vocabulary in
+                                 app.config.get('EXCLUDED_VOCABULARY_FIELDS', '').split(',')]
+        for vocabulary in get_resource_service('vocabularies').get(req=None, lookup={}):
+            if vocabulary[config.ID_FIELD] not in excluded_vocabularies:
```
petrjasek: this filtering could be done in db

apps/content_filters/filter_condition/filter_condition_resource.py
```diff
@@ -75,3 +76,17 @@ class FilterConditionResource(Resource):
...
+                self.schema['field']['allowed'].append(vocabulary[config.ID_FIELD])
+
+    def on_created_vocabularies(self, docs):
```
petrjasek: this has still the issue with multiple processes

apps/content_filters/filter_condition/filter_condition_resource.py
```diff
@@ -75,3 +76,11 @@ class FilterConditionResource(Resource):
     privileges = {'POST': 'content_filters',
                   'PATCH': 'content_filters',
                   'DELETE': 'content_filters'}
+
+    def pre_request_get(self, request, lookup):
```
petrjasek: imo this schema update is only needed for patch/post requests to do validation, or is there any other usecase?

PR #1048, opened 2017-10-11
fix newsmlg2 parser for ansa

petrjasek:
there is different date format, fix it for now


petrjasek:
ping @mdhaman I've put this on 1.8 but shouldn't be a problem for you right?

PR #1049, opened 2017-10-11
SDESK-1734 text in error message references text custom field name before modifications

mugurrus:


apps/validate/validate.py
```diff
@@ -214,6 +215,12 @@ def _get_validator_schema(self, validator):
         """
         return {field: get_validator_schema(schema) for field, schema in validator['schema'].items() if schema}
 
+    def _get_vocabulary_display_name(self, vocabulary_id):
+        vocabulary = get_resource_service('vocabularies').find_one(req=None, _id=vocabulary_id)
+        if vocabulary:
+            return vocabulary['display_name']
```
ioanpocol: Not all vocabularies had 'display_name'

PR #1052, opened 2017-10-17
feat(profile): Add support for custom media fields

ioanpocol:


apps/content_types/content_types.py
```diff
@@ -188,8 +188,11 @@ def prepare_for_edit_content_type(doc):
 def init_extra_fields(editor, schema):
     fields = get_resource_service('vocabularies').get_extra_fields()
     for field in fields:
+        field_type = 'string'
+        if field.get('field_type', None) == 'media':
```
petrjasek: I don't like this special case much, imo we should have either type=media and then content_type=picture as a new schema field or we should create the field with type=picture right away. not sure actually if there would be a usecase for a field where you can put different type of content, so maybe type=picture would be ok

ioanpocol: The requirements was to support media fields that can be image(picture), video and embed. So because of this we have type="media" and in "field_options" we have "media_type".

petrjasek: I see, but why is the type here set to `picture` then?

ioanpocol: On content profile we set the type picture because the type of the rendered field will be picture (like feature media)

petrjasek: I see, maybe we should make the ui expect type=media then and not picture, or is it for backward compatibility?

ioanpocol: If featured media support all 3 types (picture, video and embed) we can do this. Otherwise we need a way to identify what type of media is.

ioanpocol: Ok, so the conclusion is to have only type that will have one of the following values: text, picture, video or embed.

apps/content_types/content_types.py
```diff
@@ -188,8 +188,11 @@ def prepare_for_edit_content_type(doc):
 def init_extra_fields(editor, schema):
     fields = get_resource_service('vocabularies').get_extra_fields()
     for field in fields:
+        field_type = field.get('field_type', None)
+        if (field_type == 'text'):
+            field_type = 'string'
```
petrjasek: maybe we should change it to string instead of text, it's not text anyhow

apps/content_types/content_types.py
```diff
@@ -188,8 +188,11 @@ def prepare_for_edit_content_type(doc):
 def init_extra_fields(editor, schema):
     fields = get_resource_service('vocabularies').get_extra_fields()
     for field in fields:
+        field_type = field.get('field_type', None)
```
petrjasek: there should be some default value, otherwise the type would be left empty, while before it was always set

PR #1053, opened 2017-10-17
SDESK-1965 As a user I want to use custom CVs when I create a "filter condition"

mugurrus:


superdesk/default_settings.py
```diff
@@ -616,3 +616,7 @@ def local_to_utc_hour(hour):
 # SAML Auth settings
 SAML_PATH = env('SAML_PATH')
 SAML_LABEL = env('SAML_LABEL', 'Single Sign On')
+
+EXCLUDED_VOCABULARY_FIELDS = env('EXCLUDED_VOCABULARY_FIELDS', 'default_categories,signal,rightsinfo,'
```
petrjasek: imo this could be a list, I don't think we would ever change this via environment, it's easier to change it via settings file

apps/content_filters/filter_condition/filter_condition_resource.py
```diff
@@ -75,3 +81,18 @@ class FilterConditionResource(Resource):
...
+
+    def pre_request_post(self, request):
+        self._init_allowed_filters()
+
+    def pre_request_patch(self, request, lookup):
+        self._init_allowed_filters()
+
+    def _init_allowed_filters(self):
+        self.schema['field']['allowed'] = copy.copy(default_allowed_filters)
+        excluded_vocabularies = [vocabulary.strip() for vocabulary in
```
petrjasek: but if that config is a string this handling to split and strip should be in one place and not duplicated

superdesk/resource.py
```diff
@@ -152,6 +152,15 @@ def __init__(self, endpoint_name, app, service, endpoint_schema=None):
         app.register_resource(self.endpoint_name, endpoint_schema)
         superdesk.resources[self.endpoint_name] = self
 
+        for attr in self.__dir__():
+            for request_method in ['GET', 'POST', 'PATCH', 'PUT', 'DELETE']:
```
petrjasek: it would be probably enough to do:
```
 for method in ['GET', 'POST', 'PATCH', 'PUT', 'DELETE']:
    if hasattr(self, 'pre_request_%s' % method.lower()):
        ...
```

PR #1054, opened 2017-10-17
SDESK-1982 Modify handling of custom fields in ninjs output

mugurrus:


docs/superdesk-ninjs-schema.json
```diff
@@ -204,6 +204,14 @@
 					"code" : {
 						"description": "The code for the subject in a scheme (= controlled vocabulary) which is identified by the scheme property",
 						"type" : "string"
+					},
+					"scheme": {
+						"description" : "The controlled vocabulary (scheme) identifier",
+						"type" : "string"
+					},
+					"parent": {
```
petrjasek: not sure if we need the `parent` here, any usecase?

docs/superdesk-ninjs-schema.json
```diff
@@ -301,7 +305,21 @@
 			"type": "object",
 			"additionalProperties" : false,
 			"patternProperties" : {
-				"^[a-zA-Z0-9]+" :  {}
+				"^[a-zA-Z0-9]+" : {},
```
petrjasek: I think there should be no `patternProperties`, just `properties`

docs/superdesk-ninjs-schema.json
```diff
@@ -301,7 +305,21 @@
...
+				"properties" : {
```
petrjasek: I think extra is just `'key': 'val'` without any properties

PR #1055, opened 2017-10-18
put ansa on superdesk core

petrjasek:
here are changes done for ansa, idea is to move those to master where it makes sense or introduce apis so we can put this into their superdesk repo

so eventually this should have empty diff and will get closed


petrjasek:
summary:
- add newsmlg2 format with html body
- add highlights info to newsmlg2
- updates to package items publishing

SDANSA-99

PR #1061, opened 2017-10-23
[FORMATTER] [NINJS]: jobtitle is now an object with both "qcode" and …

jerome-poisson:
…"name" properties

This commit add "name" in addition to "qcode" for authors' job title in
ninjs formatter.
ninjs Superdesk schema extension has been updated.

SDESK-1928

### DIFF
--- a/docs/superdesk-ninjs-schema.json
+++ b/docs/superdesk-ninjs-schema.json
@@ -382,7 +382,15 @@
 					},
 					"jobtitle" : {
 						"description" : "author's Job Title qcode",
-						"type" : "string"
+						"type" : "object",
+						"properies" :  {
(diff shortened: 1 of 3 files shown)


docs/superdesk-ninjs-schema.json
```diff
@@ -382,7 +382,15 @@
 					},
 					"jobtitle" : {
 						"description" : "author's Job Title qcode",
-						"type" : "string"
+						"type" : "object",
+						"properies" :  {
+							"qcode": {
```
petrjasek: we use code for the other properties, would be good imo to be consistent.
if there is also scheme it should go into `scheme` field, otherwise just `code`

PR #1064, opened 2017-10-25
SDFID-138 Custom vocabulary cause to stop publishing

mugurrus:
When adding subject_custom to a content type the allowed values are
restricted to 'subject_custom'. Custom vocabularies, which are saved
also in the subject field generate an error on publishing as the scheme
allowed values does not contain the custom vocabularies.

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -359,7 +359,8 @@ def compose_subject_schema(schema, fields_map):
     default = []
     for old_field, field in fields_map.items():
         if (old_field == field or old_field == 'subject') and schema.get(field, None):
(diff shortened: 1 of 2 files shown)


ioanpocol:
@mugurrus  Please take a look at this PR: https://github.com/superdesk/superdesk-core/pull/1065
Some vocabularies got field_type=null and because of this those fields are not added to allowed.

mugurrus:
@ioanpocol I applied your changes without my fix and they don't fix the publish error.

ioanpocol:
@mugurrus  You have to edit the profile before to check. Also it is possible to be some additional bugs.

PR #1065, opened 2017-10-25
fix(vocabularies): Fix the get of custom vocabularies and custom fields

ioanpocol:
After edit some vocabularies get field_type=null

### DIFF
--- a/superdesk/vocabularies/vocabularies.py
+++ b/superdesk/vocabularies/vocabularies.py
@@ -236,10 +236,10 @@ class VocabulariesService(BaseService):
             return {}
 
     def get_extra_fields(self):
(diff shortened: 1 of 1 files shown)


superdesk/vocabularies/vocabularies.py
```diff
@@ -236,10 +236,10 @@ def get_rightsinfo(self, item):
             return {}
 
     def get_extra_fields(self):
-        return list(self.get(req=None, lookup={'field_type': {'$exists': True}}))
+        return list(self.get(req=None, lookup={'field_type': {'$exists': True, '$ne': None}}))
```
petrjasek: imo `$ne: None` should be enough, such field must exist

ioanpocol: Yes, true

PR #1070, opened 2017-11-02
[CONTENT_API] [PUBLISH_QUEUE]: fixed packages publishing for content_api subscribers

jerome-poisson:
packages were not correctly published to content_api, but delivery
status was always set to success. This patch fixes it by publishing the
package just before changing the delivery state.

SDESK-2073

### DIFF
--- a/superdesk/publish/publish_queue.py
+++ b/superdesk/publish/publish_queue.py
@@ -108,7 +108,7 @@ class PublishQueueResource(Resource):
     }
 
     additional_lookup = {
-        'url': 'regex("[\w,.:-]+")',
+        'url': r'regex("[\w,.:-]+")',
         'field': 'item_id'
     }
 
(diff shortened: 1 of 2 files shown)


superdesk/publish/publish_queue.py
```diff
@@ -142,9 +142,7 @@ def on_updated(self, updates, original):
                               )
 
     def _set_queue_state(self, updates, original):
-        destination = updates.get('destination', original.get('destination')) or {}
-        updates['state'] = QueueState.SUCCESS.value if destination.get('delivery_type') == 'content_api' \
-            else updates.get('state') or QueueState.PENDING.value
+        updates['state'] = updates.get('state') or QueueState.PENDING.value
```
petrjasek: this looks like `updates.setdefault('state, QueueState.PENDING.value)`

jerome-poisson: yes I've kept the way it was initially done, but a setdefault is better, or putting the value in second parameter of the get.

PR #1073, opened 2017-11-08
[CONTENT_API] [PUBLISH]: fixed associations

jerome-poisson:
associations where not set correclty due to missing subscribers in data, in
addition a filter was blocking associations processing for packages

SDESK-2073

### DIFF
--- a/content_api/publish/service.py
+++ b/content_api/publish/service.py
@@ -17,7 +17,7 @@ from superdesk.utc import utcnow
 from superdesk.services import BaseService
 from superdesk.publish.formatters.ninjs_formatter import NINJSFormatter
 from superdesk import get_resource_service
-from superdesk.metadata.item import ITEM_TYPE, CONTENT_TYPE
+from superdesk.metadata.item import ASSOCIATIONS
 
 logger = logging.getLogger('superdesk')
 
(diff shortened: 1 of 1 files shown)


content_api/publish/service.py
```diff
@@ -146,7 +150,6 @@ def _assign_associations(self, item, doc):
         """
         if item[ITEM_TYPE] != CONTENT_TYPE.TEXT:
```
petrjasek: I would remove this part for consistency

content_api/publish/service.py
```diff
@@ -45,6 +45,10 @@ def publish(self, item, subscribers=[]):
             doc.setdefault('firstcreated', now)
             doc.setdefault('versioncreated', now)
             doc.setdefault(config.VERSION, item.get(config.VERSION, 1))
+            item[ASSOCIATIONS] = doc.get(ASSOCIATIONS, {}).copy()
+            for _, assoc in item[ASSOCIATIONS].items():
+                assoc['subscribers'] = [str(subscriber[config.ID_FIELD]) for subscriber in subscribers]
```
petrjasek: and here would put probably `assoc.setdefault('subscribers', subscribers)`

petrjasek: right probably without the `str` call which should be handled by assign_associations without that condition

content_api/publish/service.py
```diff
@@ -45,6 +45,10 @@ def publish(self, item, subscribers=[]):
             doc.setdefault('firstcreated', now)
             doc.setdefault('versioncreated', now)
             doc.setdefault(config.VERSION, item.get(config.VERSION, 1))
+            item[ASSOCIATIONS] = doc.get(ASSOCIATIONS, {}).copy()
```
petrjasek: this line is probably not needed

content_api/publish/service.py
```diff
@@ -45,6 +45,10 @@ def publish(self, item, subscribers=[]):
...
+            for _, assoc in item[ASSOCIATIONS].items():
+                assoc['subscribers'] = [str(subscriber[config.ID_FIELD]) for subscriber in subscribers]
+            item[ASSOCIATIONS]['subscribers'] = subscribers
```
petrjasek: this adds another assocation `'subscribers'`

PR #1074, opened 2017-11-08
[PUBLISH] [CONTENT_API] moved code handling content_api packages to p…

jerome-poisson:
…ublish_package

SDESK-2073

### DIFF
--- a/apps/publish/enqueue/enqueue_service.py
+++ b/apps/publish/enqueue/enqueue_service.py
@@ -345,6 +345,13 @@ class EnqueueService:
             if temp_queued:
                 queued = temp_queued
 
+            delivery_types = [d['delivery_type'] for d in self.get_destinations(subscriber)]
+            is_content_api_delivery = 'content_api' in delivery_types
+            # packages for content_api will not be transmitted
+            # so we need to publish them here
+            if is_content_api_delivery and subscriber.get('api_enabled'):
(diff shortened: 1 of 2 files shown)


petrjasek:
would be good to add some test for this

apps/publish/enqueue/enqueue_service.py
```diff
@@ -345,6 +345,13 @@ def publish_package(self, package, target_subscribers):
             if temp_queued:
                 queued = temp_queued
 
+            delivery_types = [d['delivery_type'] for d in self.get_destinations(subscriber)]
+            is_content_api_delivery = 'content_api' in delivery_types
```
petrjasek: afaik there is no content_api delivery type

jerome-poisson: @petrjasek it's used in several places (e.g. [there](https://github.com/superdesk/superdesk-core/blob/master/superdesk/publish/publish_queue.py#L146) and [there](https://github.com/superdesk/superdesk-core/blob/master/apps/publish/enqueue/enqueue_service.py#L499)

It is set [there](https://github.com/superdesk/superdesk-core/blob/master/apps/publish/enqueue/enqueue_service.py#L353) and the code is working anyway.

petrjasek: I see, thx

PR #1076, opened 2017-11-10
SDFID-171 Create "custom embed" field

mugurrus:


superdesk/data_updates/00005_20171110-170620_archive.py
```diff
@@ -0,0 +1,22 @@
...
+from superdesk.commands.data_updates import DataUpdate
+from superdesk.commands.rebuild_elastic_index import RebuildElasticIndex
+
+
+class DataUpdate(DataUpdate):
+
+    resource = 'archive'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        RebuildElasticIndex()()
```
petrjasek: I wouldn't do this as part of data update, imo that's a side effect and would make an update on ntb take ~1h

petrjasek:
@mugurrus imo we should revert this, why would we do the rebuild? also was this part of any release that we should consider an update?

PR #1079, opened 2017-11-14
SDESK-2057 Enable removing custom fields from Metadata settings

mugurrus:


superdesk/resource.py
```diff
@@ -160,6 +160,26 @@ def __init__(self, endpoint_name, app, service, endpoint_schema=None):
...
 
+        operations_events = {'fetched': 'fetched_resource',
+                             'fetched_item': 'fetched_item',
+                             'create': 'insert',
+                             'created': 'inserted',
+                             'update': 'update',
+                             'updated': 'updated',
+                             'delete': 'delete_item',
+                             'deleted': 'deleted_item'}
+        for method in [method for method in dir(service) if method.startswith('on_')]:
```
petrjasek: not sure if this is needed, you could subscribe explicitly in content_types init_app, any benefits to do it like this? also should have some documentation

superdesk/vocabularies/vocabularies.py
```diff
@@ -164,6 +172,17 @@ def on_replaced(self, document, original):
...
         self._send_notification(document)
 
+    def on_delete(self, doc):
+        """
+        Overriding to validate vocabulary deletion
+        """
+        if 'field_type' not in doc:
+            raise SuperdeskApiError.badRequestError('Default vocabularies cannot be deleted')
+
+    def delete(self, lookup):
```
petrjasek: there is [soft delete](http://python-eve.org/features.html#soft-delete) we could use that

superdesk/resource.py
mugurrus: This is a generic implementation for callbacks on other resources operations and seems more elegant to me than creating the subscription in init_app manually.

superdesk/vocabularies/vocabularies.py
```diff
@@ -97,9 +97,13 @@ class VocabulariesResource(Resource):
         },
         'field_options': {
             'type': 'dict',
-        },
+        }
     }
 
+    datasource = {
+        'filter': {'removed': {'$ne': True}}
```
petrjasek: is this still needed?

mugurrus: oops, forgot

PR #1080, opened 2017-11-14
fix(custom media): Fix the custom media validation

ioanpocol:


apps/validate/validate.py
```diff
@@ -55,8 +55,8 @@ def get_validator_schema(schema):
 
 
 class SchemaValidator(Validator):
-    def _validate_type_picture(self, field, value):
-        """Allow type picture in schema."""
+    def _validate_type_media(self, field, value):
```
petrjasek: imo we should keep both for backward compatibility

PR #1081, opened 2017-11-14
fix(editor3): Update associations with images from editor3

ioanpocol:


petrjasek:
can you add test?

ioanpocol:
Done

apps/archive/archive.py
```diff
@@ -100,6 +100,19 @@ def update_image_caption(body, name, caption):
     return body[0:startDescription + len('<figcaption>')] + caption + body[endDescription:]
 
 
+def update_associations(doc):
+    if 'editor_state' not in doc:
+        return
+    entityMap = doc['editor_state'].get('entityMap', {})
+    mediaList = {
+        entity[1]['data']['media']['guid']: entity[1]['data']['media']
+        for entity in entityMap.items()
```
petrjasek: imo this `entity[1]` is a bit confusing, what about `for entity in entityMap.values()` ?

PR #1082, opened 2017-11-14
[PUBLISHING] [CONTENT_API] [NINJS]: added "firstpublished" field

jerome-poisson:
"firstpublished" is set to current datetime on first publication, then
it is kept unchanged.
NINJS schema extension has been updated accordingly

SDFID-175

### DIFF
--- a/apps/archive/archive_test.py
+++ b/apps/archive/archive_test.py
@@ -11,6 +11,7 @@
 
 import copy
 import unittest
+import superdesk
 
 from datetime import timedelta, datetime
 from unittest.mock import MagicMock, patch
@@ -19,6 +20,7 @@ from bson import ObjectId
 from pytz import timezone
 
 from apps.archive.archive import SOURCE as ARCHIVE, update_image_caption, update_associations
+from superdesk.metadata.item import CONTENT_STATE
 from apps.archive.common import (
     validate_schedule, remove_media_files,
     format_dateline_to_locmmmddsrc, convert_task_attributes_to_objectId,
@@ -30,6 +32,10 @@ from superdesk.errors import SuperdeskApiError
 from superdesk.media.crop import CropService
 from superdesk.tests import TestCase
 from superdesk.utc import get_expiry_date, utcnow
+from apps.publish.content import publish
+from unittest import mock
+
+NOW = utcnow()
 
 
 class RemoveSpikedContentTestCase(TestCase):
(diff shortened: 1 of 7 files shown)


content_api/publish/service.py
```diff
@@ -38,10 +38,18 @@ def publish(self, item, subscribers=[]):
 
         :param item: item to publish
         """
-
         if not self._filter_item(item):
-            doc = self.formatter._transform_to_ninjs(item, self.subscriber)
             now = utcnow()
+            try:
+                # do we already have an item id db?
+                stored_item = self.find_one(req=None, _id=item['_id'])
```
petrjasek: would rather put this into _create_doc which is already checking previous version so it's not fetched twice

apps/publish/enqueue/__init__.py
```diff
@@ -95,8 +95,11 @@ def enqueue_item(published_item):
             # if scheduled then change the state to published
             # change the `version` and `versioncreated` for the item
             # in archive collection and published collection.
-            versioncreated = utcnow()
+            now = utcnow()
+            versioncreated = now
```
mugurrus: You can put these two in one line

jerome-poisson:
@petrjasek need to get `firstpublished` before generating ninjs. So I still get the item at the same place, but not I transmit it to _create_doc so it's not fetched 2 times.

petrjasek:
why is firstpublished needed before doing ninjs?

jerome-poisson:
@petrjasek because it is required in ninjs, that's what is requested in the issue (the schema is updated by this patch)

jerome-poisson:
@petrjasek Hi, waiting for your green light before merging, let me know if you think it should be modified, thanks.

petrjasek:
I don't see it in the requirements, it says it should be there when you fetch something from content api, and it says it's missing in ninjs, but not that it should be added there

content_api/publish/service.py
```diff
@@ -38,10 +38,20 @@ def publish(self, item, subscribers=[]):
...
             now = utcnow()
+            try:
+                # do we already have an item id db?
+                original_item = self.find_one(req=None, _id=item.get('guid') or item['_id'])
+            except KeyError:
+                original_item = None
+                check_original = True
+            else:
+                # if yes, we need to use previous firstpublished data
+                item.setdefault('firstpublished',
```
petrjasek: I'm still not sure why is this needed, it should be set on item when publishing and corrections shouldn't update it anyhow, so what's the usecase here?

petrjasek: imo it should be set when user clicks on publish, before it enters the output handling, and not when it's pushed to contentapi. so if you get an item from content api or via ftp or whatever there will be the same timestamp.

jerome-poisson: OK I see, I'll move it.

jerome-poisson:
checked with @petrjasek it's alright now. As soon as all is green we can merge this patch.

PR #1086, opened 2017-11-16
Reverted soft_delete option

jerome-poisson:
SOFT_DELETE option is documented at
http://python-eve.org/features.html#soft-delete it allows to keep
"deleted" documents in database so they can be restored, and it has been
introduced in rev. 3b63b198c .

But this option add a _delete field which is then returned by client,
resulting in request not validating against schema, and bugs like SDESK-2140 .

So this option is disabled here to have a quick fix to SDESK-2140, and
it needs further testing before being activated again.

fix SDESK-2140

### DIFF
--- a/superdesk/vocabularies/vocabularies.py
+++ b/superdesk/vocabularies/vocabularies.py
@@ -100,8 +100,8 @@ class VocabulariesResource(Resource):
         }
     }
 
-    soft_delete = True
-    item_url = 'regex("[-_\w]+")'
+    soft_delete = False
+    item_url = r'regex("[-_\w]+")'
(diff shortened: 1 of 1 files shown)


jerome-poisson:
@mugurrus this patch revert an option you introduced in https://github.com/superdesk/superdesk-core/commit/3b63b198cb7c93bcfe11669057fac8f363c81f53 because it is the source of a bug (check description). Can you check it please? Thanks :)

mugurrus:
Please do not merge. I created this PR to fix the issue: https://github.com/superdesk/superdesk-client-core/pull/1914

PR #1092, opened 2017-11-22
fix(correction): Fix the update of image metadata on correction

ioanpocol:


ioanpocol:
It is needed only where the body images can be changed. I think kill is not the case. I don't know other place to add this update.

PR #1095, opened 2017-11-24
[NINJS] [FORMATTER]: handle annotations

jerome-poisson:
Annotations are transmitted using DraftJS internal editor
representation, which separate content from style instead of using HTML.
To handle that, the drafsjs_exporter module has been choosen: after some
testing, it seems to work quite well, so it is a new dependency (using
lxml version).

To render annotations in the body_html, annotation are converted to
special entities, which are then rendered as <span> elements in the
body, using an identifier (annotation-id) local to the item.

The Superdesk ninjs schema extension has been updated accordingly.

SDFID-201

### DIFF
--- a/content_api/items/resource.py
+++ b/content_api/items/resource.py
@@ -76,6 +76,17 @@ schema = {
             'type': 'dict',
         },
     },
+    'annotations': {
+        'type': 'list',
+        'mapping': {
+            'type': 'object',
+            'properties': {
+                'id': not_analyzed,
+                'type': not_analyzed,
+                'body': not_analyzed
+            }
+        }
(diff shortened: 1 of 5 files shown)


petrjasek:
I think we will also need `annotations` in content api item mapping, other than that looks good

jerome-poisson:
@petrjasek indeed was missing, this is done now, thx.

ioanpocol:
I think we have to keep any change on body_html on editor3 inside export component.

jerome-poisson:
found an issue with media while doing another test, will fix it should be quick, please dot not merge for now.

PR #1096, opened 2017-11-24
SDESK-2145 Content profile: changes in editor 3 formatting options

mugurrus:


ioanpocol:
Some tests should be updated also

PR #1097, opened 2017-11-25
Lists for recording user actions on content api item

akintolga:


petrjasek:
just wondering if we should add something like `actions` with properties type and user, so we don't modify schema when there is another one added.

PR #1098, opened 2017-11-26
Introducing a label attribute to place field in ninjs output

akintolga:
So that label could be used to display user friendly names in filters on Newsroom application

[SDAN-150]


akintolga:
Closing this as I noticed it is not ninjs I needed to change...

petrjasek:
there is no label in the standard, imo name should be the user friendly field

akintolga:
I agree, the way the `name` property is used is not correct. It should be used to store the full name of the place, whereas in superdesk we have the same value as the `code`. 

`place` fieldhas the attribute `additionalProperties=false` so a new attribute will make the json invalid. So I'll make the change to use the name field now.

PR #1107, opened 2017-12-05
feat(translations): Set the translations option for some vocabularies…

ioanpocol:
… and add an example of translation for Categories

### DIFF
--- a/apps/prepopulate/data_init/vocabularies.json
+++ b/apps/prepopulate/data_init/vocabularies.json
@@ -5,10 +5,11 @@
         "type": "manageable",
         "unique_field": "qcode",
         "items": [
-            {"is_active": true, "name": "FIXME", "qcode": "f", "subject": ""}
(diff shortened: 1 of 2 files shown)


apps/prepopulate/data_init/vocabularies.json
```diff
@@ -187,7 +193,8 @@
         "schema": {
             "name": {},
             "width": {"type": "int"},
-            "height": {"type": "int"}
+            "height": {"type": "int"},
+            "translations": {}
```
petrjasek: not sure we need translations here

apps/prepopulate/data_init/vocabularies.json
```diff
@@ -243,7 +250,8 @@
         "type": "manageable",
         "schema" : {
             "name" : {},
-            "qcode" : {}
+            "qcode" : {},
+            "translations": {}
```
petrjasek: not sure we need translations here

PR #1108, opened 2017-12-06
[SDESK-2200] Add a variant of the NINJS formatter that includes the A…

marwoodandrew:
…PI product codes that the article matches

### DIFF
--- a/superdesk/publish/formatters/ninjs_newsroom_formatter.py
+++ b/superdesk/publish/formatters/ninjs_newsroom_formatter.py
@@ -27,7 +27,7 @@ class NewsroomNinjsFormatter(NINJSFormatter):
         :return:
         """
         result = superdesk.get_resource_service('product_tests').test_products(article, lookup={'product_type': 'api'})
-        return [p['product_id'] for p in result if p.get('matched', False)]
+        return [{'code': p['product_id'], 'name': p.get('name')} for p in result if p.get('matched', False)]
 
     def _transform_to_ninjs(self, article, subscriber, recursive=True):
         ninjs = super()._transform_to_ninjs(article, subscriber, recursive)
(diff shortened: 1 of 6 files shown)


superdesk/publish/formatters/ninjs_newsroom_formatter.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def _format_products(self, article):
+        """
+        Return a list of API product id's that the article matches.
+
+        :param article:
+        :return:
+        """
+        result = superdesk.get_resource_service('product_tests').test_products(article, lookup={'product_type': 'api'})
+        return [p['product_id'] for p in result if p.get('matched', False)]
```
petrjasek: it might be useful to include names, so there will be labels for ids if needed.
what about products having name and code like other similar fields?

marwoodandrew: @petrjasek We discussed it, the  thinking was that the size of the ninjs response would potentially  be large if the names were included. 

I think the customer admin permissioning will need to retrieve a complete list of id's and names, I don't think we'd worked out how that would work. 

I guess I can include the names, we can always remove them if it is a problenm

petrjasek: I see, using names from items would not let you setup some products in newsroom upfront without having items from superdesk using it, not sure if this could be a problem, but if not lets add the names

superdesk/publish/formatters/ninjs_newsroom_formatter.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def _format_products(self, article):
+        """
+        Return a list of API product id's that the article matches.
+
+        :param article:
+        :return:
+        """
+        result = superdesk.get_resource_service('product_tests').test_products(article, lookup={'product_type': 'api'})
+        return [{'id': p['product_id'], 'name': p.get('name')} for p in result if p.get('matched', False)]
```
petrjasek: could it be `code` and `name` so it's consistent with other fields?

marwoodandrew:
@petrjasek Are we OK to merge this into master?

petrjasek:
yep :wink:

PR #1109, opened 2017-12-06
fix(editor3): Prevent the merge of editor state in backend by wrappin…

ioanpocol:
…g it in a list


petrjasek:
what exactly is the issue here?

PR #1110, opened 2017-12-07
fix(ninjs): If the abstract is reset (it is '') also reset descriptio…

ioanpocol:
…n_html and description_text


petrjasek:
but can be done later if this is urgent

ioanpocol:
Done

PR #1112, opened 2017-12-08
fix(associations): Fix the delete of items in associations

ioanpocol:


content_api/items/service.py
```diff
@@ -253,6 +253,8 @@ def _process_item_associations(self, item):
         allowed_items = {}
         if item.get('associations'):
             for _k, v in item.get('associations', {}).items():
+                if v is None:
```
petrjasek: maybe `if not v:` would be better, it would also handle `{}` or `False` besides `None`

PR #1113, opened 2017-12-08
feat(vocabularies): Add support for vocabularies translation on ninjs…

ioanpocol:
… output

### DIFF
--- a/superdesk/publish/formatters/ninjs_formatter.py
+++ b/superdesk/publish/formatters/ninjs_formatter.py
@@ -149,7 +149,7 @@ class NINJSFormatter(Formatter):
             ninjs['description_html'] = self.append_body_footer(article)
 
         if article.get('place'):
-            ninjs['place'] = self._format_place(article['place'])
+            ninjs['place'] = self._format_place(article)
 
         if article.get('profile'):
             ninjs['profile'] = self._format_profile(article['profile'])
@@ -310,8 +310,12 @@ class NINJSFormatter(Formatter):
     def _format_qcodes(self, items):
         return [{'name': item.get('name'), 'code': item.get('qcode')} for item in items]
 
-    def _format_place(self, items):
-        locator_map = superdesk.get_resource_service('vocabularies').find_one(req=None, _id='locators')
+    def _format_place(self, article):
+        vocabularies_service = superdesk.get_resource_service('vocabularies')
+        locator_map = vocabularies_service.find_one(req=None, _id='locators')
+        if locator_map and 'items' in locator_map:
+            locator_map['items'] = vocabularies_service.get_locale_vocabulary(
+                locator_map.get('items'), article.get('language'))
 
         def getLabel(item):
             locators = [l for l in locator_map.get('items', []) if l['qcode'] == item.get('qcode')]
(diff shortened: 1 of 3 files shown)


superdesk/vocabularies/vocabularies.py
```diff
@@ -228,6 +228,7 @@ def get_rightsinfo(self, item):
         if not all_rights or not all_rights.get('items'):
             return {}
         try:
+            all_rights['items.'] = self.get_locale_vocabulary(all_rights.get('items'), item.get('language'))
```
petrjasek: seems like a typo `items.`

tests/vocabularies_tests.py
```diff
@@ -100,5 +100,17 @@ def test_get_rightsinfo(self):
...
+    def test_get_locale_vocabulary(self):
+        items = [
+            {'is_active': True, 'name': 'FIXME1', 'qcode': 'f', 'subject': '',
+             'translations': {'name': {'fr': 'FIXME1-fr', 'es': 'FIXME1-es'}}},
+            {'is_active': True, 'name': 'FIXME2', 'qcode': 'f', 'subject': '',
+             'translations': {'name': {'fr': 'FIXME2-fr', 'es': 'FIXME2-es'}}}
+        ]
+        result = VocabulariesService().get_locale_vocabulary(items, 'fr')
+
+        self.assertEqual(result[0]['name'], 'FIXME1-fr')
```
petrjasek: can you check it like `self.assertEqual(result[0], {'name': 'FIXME1-fr', 'code': 'f'})` so we make sure there are no extra fields like translations

ioanpocol: Why we have to drop translations since it is already on original vocabulary?

petrjasek: because it shouldn't make it to the output

ioanpocol: I think it is not the case, the code that uses vocabularies just pick the name or any other value that is needed. The vocabulary is not put in output as a whole.

superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -311,10 +311,13 @@ def _format_qcodes(self, items):
...
 
     def _format_place(self, items):
-        locator_map = superdesk.get_resource_service('vocabularies').find_one(req=None, _id='locators')
+        vocabularies_service = superdesk.get_resource_service('vocabularies')
+        locator_map = vocabularies_service.find_one(req=None, _id='locators')
 
         def getLabel(item):
-            locators = [l for l in locator_map.get('items', []) if l['qcode'] == item.get('qcode')]
+            locale_locators = vocabularies_service.get_locale_vocabulary(
+                locator_map.get('items'), item.get('language')) or []
```
petrjasek: item here is a value from `item.place` so imo it won't have language set

tests/vocabularies_tests.py
petrjasek: you're right, it's not used directly.

superdesk/publish/formatters/ninjs_formatter.py
ioanpocol: Yes, true
I updated the code

PR #1115, opened 2017-12-11
add annotations field to item meta

petrjasek:
it will be populated by client

SDFID-201


petrjasek:
required by https://github.com/superdesk/superdesk-client-core/pull/1938

PR #1124, opened 2017-12-15
SDANSA-95 Create CV for publishing destinations

mugurrus:


superdesk/metadata/item.py
```diff
@@ -557,6 +557,9 @@
     'assignment_id': {
         'type': 'string',
         'mapping': not_analyzed
+    },
+    'destination': {
```
petrjasek: why is this needed?

mugurrus: To store the publishing destination

petrjasek: but why not use custom vocabularies?

PR #1127, opened 2017-12-21
fix(correction): Allow the dateline to be changed on corrections

ioanpocol:


petrjasek:
@mdhaman can you check this pls? I think it was there for a reason

ioanpocol:
AAP confirmed that dateline can be changed on corrections.

PR #1129, opened 2017-12-28
fix(ninjs): Fix the translations for subject and service

ioanpocol:


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -60,11 +60,15 @@ def filter_empty_vals(data):
     return dict(filter(lambda x: x[1], data.items()))
 
 
-def format_cv_item(item):
+def get_locale_name(item, language):
+    return item.get('translations', {}).get('name', {}).get(language, None) or item.get('name', '')
+
```
gbbr: Extra blank line.

ioanpocol: On python it is required 2 lines between functions outside class

PR #1138, opened 2018-01-10
(fix): Removing the duplicate operators for headline filter parameter.

mdhaman:


petrjasek:
I see you've added it to 1.8, do you need it also on 1.9 or is it ok to have it on next release?

mdhaman:
@petrjasek I did the PR for 1.8 as `cherry-pick` as it was having merge conflicts. You need that change in versions >= 1.8.

PR #1141, opened 2018-01-10
feat(validation): Add support for validation of date custom field

ioanpocol:


apps/validate/validate.py
```diff
@@ -65,6 +67,12 @@ def _validate_type_embed(self, field, value):
         """Allow type media in schema."""
         pass
 
+    def _validate_type_date(self, field, value):
+        try:
+            datetime.strptime(value or '', '%d/%m/%Y')
```
petrjasek: we use `%Y-%m-%d` usually, why this format here?

ioanpocol: This was the default format on frontend, I changed on frontend to send always in  %Y-%m-%d format.

petrjasek:
can you add some test where it would check any values?

ioanpocol:
Done

PR #1145, opened 2018-01-15
[DOCUMENTATION] added a readthedoc config file to activate latest doc…

jerome-poisson:
…ker image

This image should have plantuml installed, which we use in auth
documentation.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,2 @@
+build:
+    image: latest


petrjasek:
can you make it hidden?

jerome-poisson:
@petrjasek done

petrjasek:
we can skip CI here

PR #1146, opened 2018-01-16
[PUBLISHING] [CORRECTION] update word_count on correction

jerome-poisson:
the "update_word_count" function is now in text_utils as it is a
generally useful function.

SDFID-194


petrjasek:
imo it's missing tests like the client PR

PR #1147, opened 2018-01-17
feat(vocabularies): Update vocabularies values on content item from a…

ioanpocol:
…rchive and published


petrjasek:
I don't see the need for those new resources, why it's there? also when getting items we can filter by language so we only get items which might use translations and it could also check if there is any value set for those affected fields

ioanpocol:
Those new resource are added in order to prevent the adding of new versions that is a mandatory requirement. Also it prevent to run code that usually run on update and that code can slow a little the process.
The set value from a vocabulary contains all metadata (name, qcode, translations), so it is not affected by the set language.

petrjasek:
there is a `service.system_update` that should prevent any extra modifications

ioanpocol:
OK I will check, I hope that also elastic is updated.

ioanpocol:
Done

PR #1150, opened 2018-01-19
SDESK-2306 An error message should be displayed for all the custom required fields in case if those are missing

mugurrus:


petrjasek:
test pls

features/validate.feature
```diff
@@ -244,3 +244,57 @@ Feature: Validate
...
+    }}]
+    """
+
+    When we post to "/validate"
+    """
+    {"act": "publish", "type": "text", "validate": {"profile": "foo"}}
+    """
+    Then we get existing resource
+    """
+    {"errors": ["Text 1 is a required field", "MEDIA1 is a required field",
```
petrjasek: it should be using display name for media1 field

PR #1159, opened 2018-01-30
Twitter ingest

mgogoulos:
Adds Twitter as Ingest provider (works with multiple screen_names and hashtags). 

I open this PR hoping to start a discussion on what would be required for this to get merged. 

There are a few changes on the client (https://github.com/mgogoulos/superdesk-client-core/tree/twitter_ingest) and a quick way to test this PR is described here:
https://medium.com/@mgogoulos/adding-twitter-feeds-on-superdesk-newsroom-7208d1e938b8

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,4 +12,5 @@ sphinxcontrib-plantuml
 docutils==0.12
 requests-mock
 responses
+python-twitter
 -e .
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,5 +12,4 @@ sphinxcontrib-plantuml
 docutils==0.12
 requests-mock
 responses
-python-twitter
 -e .
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ install_requires = [
     'PyYAML>=3.11,<3.13',
     'lxml>=3.8,<3.9',
     'draftjs_exporter[lxml]>=1.1.0,<2.0',
-    'python-twitter>=3.3',
+    'python-twitter==3.3',
 ]
 
 package_data = {
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,7 @@ install_requires = [
     'PyYAML>=3.11,<3.13',
     'lxml>=3.8,<3.9',
     'draftjs_exporter[lxml]>=1.1.0,<2.0',
+    'python-twitter>=3.3',
 ]
 
 package_data = {
--- a/superdesk/io/feeding_services/twitter.py
+++ b/superdesk/io/feeding_services/twitter.py
@@ -1,0 +1,160 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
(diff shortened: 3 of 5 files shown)


requirements.txt
```diff
@@ -12,4 +12,5 @@ sphinxcontrib-plantuml
 docutils==0.12
 requests-mock
 responses
+python-twitter
```
petrjasek: this should go into setup.py, in requirements we keep test dependencies

petrjasek:
hey, it looks good to me, you might have to sign a cla, @hlmnrmr should give you more info about it

setup.py
```diff
@@ -42,6 +42,7 @@
     'PyYAML>=3.11,<3.13',
     'lxml>=3.8,<3.9',
     'draftjs_exporter[lxml]>=1.1.0,<2.0',
+    'python-twitter>=3.3',
```
petrjasek: would be better to fix the version to `<4.0` or something, so it won't break when they release an incompatible version

superdesk/io/feeding_services/twitter.py
```diff
@@ -0,0 +1,164 @@
...
+                item = {}
+                item['headline'] = headline
+                item['type'] = 'text'
+                item['guid'] = guid
+                item['versioncreated'] = d
+                item['firstcreated'] = d
+
+                item['body_html'] = status.text
+                # include URL on body
+                urls = re.findall('http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+', status.text)
```
jerome-poisson: the regex contains invalid sequence `\(`. You can use a raw string (`r"…"`) to avoid that

superdesk/io/feeding_services/twitter.py
```diff
@@ -0,0 +1,164 @@
...
+                if urls:
+                    item['body_html'] += '<p><a href="%s" target="_blank">%s</a></p>' % (urls[0], urls[0])
```
jerome-poisson: I know that dangerous chars are not in the current regex, but I would be more comfortable with some quoting, just in case somebody touch the regex in the future, to avoid XSS.

petrjasek:
there are a few format issues:
```
./superdesk/io/feeding_services/twitter.py:60:64: F821 undefined name 'ex'
./superdesk/io/feeding_services/twitter.py:91:121: E501 line too long (128 > 120 characters)
./superdesk/io/feeding_services/twitter.py:109:25: E722 do not use bare except'
./superdesk/io/feeding_services/twitter.py:113:1: D403 First word of the first line should be properly capitalized
./superdesk/io/feeding_services/twitter.py:164:1: E305 expected 2 blank lines after class or function definition, found 1
./superdesk/io/feeding_services/__init__.py:189:1: E402 module level import not at top of file
./superdesk/io/feeding_services/__init__.py:189:1: F401 'superdesk.io.feeding_services.twitter.TwitterFeedingService' imported but unused
```

petrjasek:
last one:
```
/superdesk/io/feeding_services/twitter.py:124:1: D403 First word of the first line should be properly capitalized
```

petrjasek:
cool thx

mgogoulos:
Nice! Just opened the PR for the client side,  https://github.com/superdesk/superdesk-client-core/pull/2068

PR #1160, opened 2018-01-30
SDFID-284 Labels on stories in a package

mugurrus:


mugurrus:
@petrjasek We need the upgrade to add the vocabulary to existing instances.

petrjasek:
we run app:initialize_data there afaik, so new vocabulary should make it there automatically

PR #1163, opened 2018-02-01
Sdesk 2328

jerome-poisson:


tests/io/fixtures/ritzau_feed.xml
```diff
@@ -0,0 +1,122 @@
...
+        <town>K&#xF8;benhavn</town>
+        <origin>Via Ritzau</origin>
+        <sectionID>42</sectionID>
+        <section>Int. Pressemed.</section>
+        <sectionShort>Y</sectionShort>
+        <version>1</version>
+        <Replaces i:nil="true"/>
+        <format>inpublic</format>
+        <content>&lt;P&gt;&lt;/P&gt;&#xD;
+	&lt;P&gt;GLASGOW, Scotland, Feb. 01, 2018 (GLOBE NEWSWIRE) -- The UK Space Agency has awarded more than &#xC2;&#xA3;4 million to Spire Global, a satellite powered data company, to demonstrate cutting-edge space technology including &amp;amp;apos;parallel super-computing&amp;amp;apos;, UK Government ministers Lord Henley and Lord Duncan announced today.&#xD;
```
petrjasek: does it unescape those html entities?

tests/io/fixtures/ritzau_feed.xml
```diff
@@ -0,0 +1,122 @@
+<?xml version="1.0"?>
+<XMLNewsQueuedResponse xmlns="http://tempuri.org/">
+  <XMLNewsQueuedResult xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
+    <error xmlns="">0</error>
+    <errormsg xmlns=""/>
+    <ListOfRBNews xmlns="">
+      <RBNews>
+        <NewsID>83e4e779-e3b7-46da-b061-34ab92bc6ec4</NewsID>
+        <PublishDate>2018-02-01T18:44:52.973</PublishDate>
```
petrjasek: is this utc date?

tests/io/fixtures/ritzau_feed.xml
jerome-poisson: yes, it's done by the xpath. Actually I've put the tests in NTB part because there are service and categories modified there, but I should maybe add test for the feed here too.

tests/io/fixtures/ritzau_feed.xml
jerome-poisson: It is actually not specified in the documentation.

tests/io/fixtures/ritzau_feed.xml
jerome-poisson: I've added a test for the feed parser.

tests/io/feed_parsers/ritzau_test.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertTrue(RitzauFeedParser().can_parse(self.root_elt))
+
+    def test_content(self):
+        item = self.item
+        self.assertEqual(item['version'], 1)
+        self.assertEqual(item['byline'], 'Via Ritzau')
+        self.assertEqual(item['headline'], 'PR / Spire Global, Inc.')
+        self.assertEqual(item['abstract'], 'UK Space Agency Backs Spire&apos;s Proposal to European Space Agency as'
+                                           ' Space Mission Provider')
+        self.assertEqual(hashlib.md5(item['body_html'].encode('utf-8')).hexdigest(), '8cda6746538782654a150c029b197fc0')
```
petrjasek: imo it would be better to test it for something like `assertIn('<p>GLASGOW, Scotland', item['body_html'])`, I have no idea where this hash 8cda comes from

jerome-poisson: it was to avoid copying huge part of text in the test. I can limit the test to just a part in clear text then

tests/io/fixtures/ritzau_feed.xml
jerome-poisson: @petrjasek checking again this <content>, I realize that the format on the feeds I have is not the same on the doc (it's escaped on the feeds, not in the doc), and the content here is broken (uncomplete tags) while it's fine on other news I've got. I think it's a feed issue from Ritzau, but I'll check this with @karelpetrak once he's back.

jerome-poisson: Also found broken HTML, I can clean it, but that doesn't smell good. I've explained that to @karelpetrak so he can check what to do with NTB. Maybe it's better to wait before merging this P.R. then.

jerome-poisson:
added `WIP` label as we are waiting for details

PR #1164, opened 2018-02-07
fix(ninjs): avoid empty assocations in ninjs output

petrjasek:
when embed is removed it shouldn't be in ninjs as `null`

SDESK-2376

### DIFF
--- a/superdesk/publish/formatters/ninjs_formatter.py
+++ b/superdesk/publish/formatters/ninjs_formatter.py
@@ -289,8 +289,6 @@ class NINJSFormatter(Formatter):
         for key, item in (article.get(ASSOCIATIONS) or {}).items():
             if item:
                 associations[key] = self._transform_to_ninjs(item, subscriber)
-            else:
-                associations[key] = None
         return associations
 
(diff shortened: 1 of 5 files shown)


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -289,8 +289,6 @@ def _format_related(self, article, subscriber):
         for key, item in (article.get(ASSOCIATIONS) or {}).items():
             if item:
                 associations[key] = self._transform_to_ninjs(item, subscriber)
-            else:
-                associations[key] = None
```
ioanpocol: This will generate a problem on updating the content api after an item from associations is deleted.

petrjasek: that's why I've added this test - https://github.com/superdesk/superdesk-core/pull/1164/files#diff-57be8e755f04f5d58c1694174ac4baafR200 - doesn't seem to be a problem, because that code which merges it with original is only applied when doing patch request, while the content api is writing it directly to the storage

PR #1173, opened 2018-02-12
[CONTENT API] fixed schema

jerome-poisson:
"wordcount", "charcount" and "readtime" were missing in item schema

fix SDESK-2414


petrjasek:
just wondering how it's related to 2414 ?

PR #1184, opened 2018-02-20
fix(ninjs): fix formatter with new category items

petrjasek:
when you add an item it has `translations` set to `null`
whick breaks formatting.


petrjasek:
fixes https://github.com/superdesk/superdesk/issues/2828

PR #1185, opened 2018-02-20
SDESK-1881 Load the JID field in the user profile when XMPP auth enabled only

mugurrus:


apps/auth/xmpp/__init__.py
```diff
@@ -14,7 +14,8 @@
 
 
 def init_app(app):
-    endpoint_name = 'auth_xmpp'
-    service = XMPPAuthService('auth', backend=superdesk.get_backend())
-    XMPPAuthResource(endpoint_name, app=app, service=service)
     app.client_config['xmpp_auth'] = bool(app.config['XMPP_AUTH_URL'])
```
petrjasek: the config for client is already set here so I don't think any change is needed

PR #1192, opened 2018-02-25
SDANSA-119 Media workflow

mugurrus:


apps/publish/content/common.py
```diff
@@ -538,28 +542,46 @@ def apply_kill_override(self, item, updates):
         except Exception:
             logger.exception('Failed to apply kill header template to item {}.'.format(item))
 
-    def _refresh_associated_items(self, original):
+    def _refresh_associated_items(self, original, updates={}, publish=False):
```
petrjasek: imo would be better to have publish_associated_items, it would be more consistent with packages,
now it's a bit misleading when refresh does publishing

mdhaman: You are right it is misleading refresh does publishing. Those 2 should be separate.

apps/publish/content/common.py
```diff
@@ -538,28 +542,46 @@ def apply_kill_override(self, item, updates):
...
-                    is_db_item_bigger_ver = updates['_current_version'] > original['_current_version']
+                    is_db_item_bigger_ver = original_item['_current_version'] > original['_current_version']
                 except KeyError:
-                    update_item_data(item, updates, keys)
+                    update_item_data(item, original_item, keys)
                 else:
-                    update_item_data(item, updates, keys, keep_existing=not is_db_item_bigger_ver)
+                    update_item_data(item, original_item, keys, keep_existing=not is_db_item_bigger_ver)
+
+                if publish_service and publish and item['type'] in MEDIA_TYPES:
```
mdhaman: We don't publish media separately when the `Text` item is published. Can we have config setting to publish media or not?

apps/publish/content/common.py
```diff
@@ -538,28 +542,46 @@ def apply_kill_override(self, item, updates):
...
-                    updates = original
+                    original_item = original
                     keys = FIELDS_TO_COPY_FOR_ASSOCIATED_ITEM
                 else:
-                    updates = super().find_one(req=None, _id=item[config.ID_FIELD]) or {}
+                    original_item = super().find_one(req=None, _id=item[config.ID_FIELD]) or {}
 
                 try:
-                    is_db_item_bigger_ver = updates['_current_version'] > original['_current_version']
+                    is_db_item_bigger_ver = original_item['_current_version'] > original['_current_version']
```
mdhaman: `original_item['_current_version'] > original['_current_version']` always be `false` as we are comparing the same item. If you have the config setting `COPY_METADATA_FROM_PARENT` and it won't copy/overwrite the `metadata` from the `item` to the `associated media`.

@mugurrus these changes were done as part of SDFID-211.

mugurrus: I just renamed the variable, didn't change the logic here.

apps/publish/content/common.py
mugurrus: Done

apps/publish/content/common.py
```diff
@@ -94,7 +99,7 @@ class BasePublishService(BaseService):
     package_service = PackageService()
 
     def on_update(self, updates, original):
-        self._refresh_associated_items(original)
+        self._publish_associated_items(original)
```
petrjasek: I think here it should still do the refresh, so there should be both refresh and publish methods, and from publish_associated_items you can call refresh_associated_items if needed

apps/publish/content/common.py
```diff
@@ -538,28 +542,46 @@ def apply_kill_override(self, item, updates):
         except Exception:
             logger.exception('Failed to apply kill header template to item {}.'.format(item))
 
-    def _refresh_associated_items(self, original):
+    def _publish_associated_items(self, original, updates={}, publish=False):
         """Refresh associated items before publishing
```
petrjasek: description is off now

apps/publish/content/common.py
mugurrus: Initially I implemented it as 2 methods but I wanted to avoid reading the associated item from the database 2 times. I think the method is quite clear as it is now, will update the description.

apps/publish/content/common.py
mdhaman: @mugurrus I know that. Hence, I have stated that in my comment above. Would it be possible to change it?

mugurrus: @mdhaman How would you like to change it?

mdhaman: @mugurrus thanks. Logic is correct now as you have removed `is_db_item_bigger_ver 
 = original_item['_current_version'] > original['_current_version']` check.

PR #1197, opened 2018-02-28
fix(ingest): allow ingestion of json files via ftp

petrjasek:


superdesk/io/feeding_services/ftp.py
```diff
@@ -44,8 +44,8 @@ class FTPFeedingService(FeedingService):
     ERRORS = [IngestFtpError.ftpUnknownParserError().get_error_description(),
               IngestFtpError.ftpError().get_error_description()]
 
-    FILE_SUFFIX = '.xml'
     DATE_FORMAT = '%Y%m%d%H%M%S'
+    ALLOWED_EXT = set(('.json', '.xml'))
```
jerome-poisson: using `{'.json', '.xml'}` would be cleared and would avoid converting tuple to set.

PR #1198, opened 2018-03-02
Adapt to changed comment data structure

tomaskikutis:


petrjasek:
you will have to fix the test payload in https://github.com/superdesk/superdesk-core/blob/master/features/archive_comments.feature#L146

tomaskikutis:
Thanks for the tip, I'll try to do that

PR #1199, opened 2018-03-05
[INGEST] remove content_expiry if value is negative

jerome-poisson:
when disabled, content_expiry has a negative value resulting in an
expiration date in the past. This patch fix the backend part of it by
adding extra checks.

SDESK-2539


jerome-poisson:
This patch fix it and log a warning. I'm not sure if the warning is really needed, we can just remove negative value, which would avoid changing the client.

PR #1205, opened 2018-03-07
[FEED PARSER] [RITZAU] removed "place" field

jerome-poisson:
"place" was not ingested with the right data type (a list is expected
while a string was used), resulting in ingestion troubles.
As the field is not really needed, it has been removed.

SDNTB-479


jerome-poisson:
Note: I've checked with @karelpetrak , the `place` field can be removed.

PR #1211, opened 2018-03-12
feat(translate): translate package items

petrjasek:
when translating a package translate each item contained in it.

SDFID-270

### DIFF
--- a/features/content_translate.feature
+++ b/features/content_translate.feature
@@ -66,3 +66,49 @@ Feature: Translate Content
       {"type":"text", "headline": "test1", "state": "submitted", "sign_off": "abc", "language": "de", "source": "AAP", 
        "subject":[{"qcode": "17004000", "name": "Statistics"}], "body_html": "Test Document body"}
       """
+
+    @auth
+    Scenario: Translate package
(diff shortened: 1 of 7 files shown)


features/content_translate.feature
```diff
@@ -66,3 +66,54 @@ Feature: Translate Content
...
+      Then we get new resource
+      When we post to "/archive/translate"
+      """
+      {"guid": "package", "language": "de", "desk": "#desks._id#"}
+      """
+  
+      When we get "/archive/#translate._id#"
+      Then we get existing resource
+      """
+      {"translated_from": "package", "operation": "translate", "task": {"stage": "#desks.working_stage#"}}
```
ioanpocol: How do we know that the new items are really part of the new package?

petrjasek: I didn't know how to test this because those new item ids are generated, any suggestions?

ioanpocol: You can set for  "residRef" the value "__any_value__", and we at least know that there are 2 items in package.

petrjasek: added a new step definition to check package items

PR #1212, opened 2018-03-12
[INGEST] AP Feeding service and NewsMLG2 updates

jerome-poisson:
NITF content in NewsML-G2 parser is now ingested, and the parser is
not crashing anymore if priority is missing.
New "private" field allows providers to keep any useful data, it's used
here to keep position in feeds.

SDNTB-474


tests/io/feed_parsers/ap_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def test_can_parse(self):
+        self.assertTrue(NewsMLTwoFeedParser().can_parse(self.root_elt))
+
+    def test_content(self):
+        item = self.items[0]
+        self.assertEqual(item['version'], '2')
+        self.assertEqual(item['headline'], 'Nutrisystem beats 4Q earnings and revenue expectations')
+        self.assertEqual(item['slugline'], 'BC-US--Earns-Nutrisystem, 1st Ld-Writethru')
+        self.assertEqual(item['body_html'], '<div>\n              <p>FORT WASHINGTON, Pa. (AP) _ Nutrisystem Inc. (NTR'
```
petrjasek: would be better to avoid that div around the content

PR #1216, opened 2018-03-20
SDESK-2389 Editor3 - Change the generation of html to include annotations

mugurrus:


ioanpocol:
Why we still call _parse_editor_state at the following line:
https://github.com/mugurrus/superdesk-core/blob/0c85f86d6f1c84e2ece870b9dcaa5d4d858f9de8/superdesk/publish/formatters/ninjs_formatter.py#L136 ?

jerome-poisson:
if annotations are now generated in client, we can remove the whole thing indeed, and the dependency in setup.py

jerome-poisson:
I've checked the code, we can get rid of `_parse_editor_state`, all `_render_*` methods, `draftjs_exporter` import and `draftjs_exporter[lxml]>=1.1.0,<2.0` in `setup.py`.

PR #1220, opened 2018-03-22
Migration scripts for annotations

tomaskikutis:
* Run a mongo query to find all articles having `editor_state`
* take `editor_state` serialize to JSON, execute node script with that JSON as stdin
* node script transforms annotations to the new structure of highlights
* back to python, it takes the stdout and writes back to mongo item "editor_state" field


ioanpocol:
Looks ok, there are some small flake8 issues.
It will help if we test it on the fil test server and only then to deploy on UAT server.

superdesk/data_updates/00007_20180321-092824_archive.py
```diff
@@ -0,0 +1,42 @@
...
+
+import os
+import subprocess
+import json
+from superdesk.commands.data_updates import DataUpdate
+
+def get_updated_editor_state(editor_state):
+    try:
+        p1 = subprocess.Popen(["echo", json.dumps(editor_state)], stdout=subprocess.PIPE)
+        p2 = subprocess.Popen(["node", os.path.abspath('../env/src/superdesk-core/superdesk/data_updates/00007_20180321-092824_archive.dist.js')], stdin=p1.stdout, stdout=subprocess.PIPE)
```
jerome-poisson: normally you can use `__file__` to get module location, but I've made a quick try and it's not working with data updates (it depends how the script is launched). As an alternative, you can do following imports:
```python
import superdesk
from os.path import realpath, join, dirname
```
then do:

```python
join(dirname(realpath(superdesk.__file__)), "data_updates", "00007_20180321-092824_archive.dist.js")
```

superdesk/data_updates/00007_20180321-092824_archive.py
```diff
@@ -0,0 +1,42 @@
...
+        p1.stdout.close()  # Allow p1 to receive a SIGPIPE if p2 exits.
+        output,err = p2.communicate()
+
+        return json.loads(output.decode("utf-8"))
+
+    except Exception as exception:
```
jerome-poisson: you don't need to put `as exception` if you don't use it (in fact, flake8 tests wont like it).
Also it's good practice to put explicitly the Exceptions you are expecting, but as we are on a one shot upgrade script, it's probably not a big deal here.

tomaskikutis:
Fixed a few minor issues, tested the latest change by checking out fi 1.4, creating articles with annotations and observing them disappear after using master's code and reappear again after running the migration script.

I was testing against master, but the data structure for annotations is the same for fi 1.5 so I think we are good.

superdesk/data_updates/00007_20180321-092824_archive.py
```diff
@@ -0,0 +1,42 @@
...
+# Creation: 2018-03-21 09:28
+
+import os
+import subprocess
+import json
+from superdesk.commands.data_updates import DataUpdate
+
+def get_updated_editor_state(editor_state):
+    try:
+        p1 = subprocess.Popen(["echo", json.dumps(editor_state)], stdout=subprocess.PIPE)
```
jerome-poisson: Don't use `echo` to send data, you have `communicate` method for that (which you use below), just pass your data to it.
Also you should use the `with` statement: 
```python
with subprocess.Popen(…) as p:
    p.do_something
```

superdesk/data_updates/00007_20180321-092824_archive.py
tomaskikutis: I wasn't sure if it's required, removed them.

superdesk/data_updates/00007_20180321-092824_archive.py
tomaskikutis: Fixed it :)

PR #1228, opened 2018-03-27
[SDESK-2614] fix(amazon_media_storage) Use file extension when getting from Amazon S3 storage

MarkLark86:


superdesk/storage/amazon_media_storage.py
```diff
@@ -317,3 +317,6 @@ def find(self, folder=None, upload_date=None, resource=None):
             next_marker = objects[-1]['Key']
 
         return files
+
+    def getFilename(self, media_id):
```
petrjasek: by [convetions](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) this should be `get_filename`

PR #1235, opened 2018-03-30
Add fields_meta and an upgrade script to move existing editor_state

tomaskikutis:
SDFID-345

MUST BE MERGED TOGETHER WITH CLIENT CHANGES https://github.com/superdesk/superdesk-client-core/pull/2180

### DIFF
--- a/superdesk/metadata/item.py
+++ b/superdesk/metadata/item.py
@@ -524,13 +524,6 @@ metadata_schema = {
         'type': 'boolean'
     },
 
-    # draft-js editor state
-    'editor_state': {
-        'type': 'list',
-        'nullable': True,
-        'mapping': not_enabled,
-    },
-
     'fields_meta': {
         'type': 'dict',
         'mapping': {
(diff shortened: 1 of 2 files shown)


superdesk/metadata/item.py
```diff
@@ -524,11 +524,11 @@
...
 
-    # draft-js editor state
-    'editor_state': {
-        'type': 'list',
-        'nullable': True,
-        'mapping': not_enabled,
+    'fields_meta': {
+        'type': 'dict',
+        'mapping': {
+            'type': 'object',
```
petrjasek: can you specify the properties? it would be better to avoid any dynamic mapping

tomaskikutis: It can contain custom fields created by the user, so it's not possible to have a static mapping

petrjasek: can you set it to `not_enabled` then? it shouldn't get indexed

tomaskikutis:
Closed in favor of #1238

PR #1238, opened 2018-04-04
[METADATA] schema update: replaced "editor_state" by "fields_meta"

jerome-poisson:
SDFID-360

### DIFF
--- a/superdesk/data_updates/00008_20180404-165521_archive.py
+++ b/superdesk/data_updates/00008_20180404-165521_archive.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8; -*-
+# This file is part of Superdesk.
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+#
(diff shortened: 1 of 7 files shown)


jerome-poisson:
data migration script is still missing, an update will follow

superdesk/data_updates/00008_20180404-165521_archive.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        for resource in ['archive', 'archive_autosave', 'published']:
+
+            collection = mongodb_database[resource]
+
+            for item in collection.find({'editor_state': {'$exists': True}}):
+                state = item['editor_state']
+                fields_meta = {'body_html': {'draftjsState': state}}
+                print(collection.update({'_id': item['_id']}, {
```
petrjasek: this won't make the update visible in elastic, it should use service

jerome-poisson: after talking with @petrjasek this change is not needed in this particular case (fields are not indexed in elastic).

PR #1239, opened 2018-04-05
[SDANSA-142] Accept default desk monitoring view

pablopunk:


apps/desks.py
```diff
@@ -97,6 +97,12 @@ class DeskTypes(SuperdeskBaseEnum):
         'type': 'string'
     },
 
+    'desk_default_view': {
```
petrjasek: maybe this could be `monitoring_default_view` so we know where it's used

PR #1242, opened 2018-04-12
SDESK-2680 contentapi is always reporting success even when something is going wrong

mugurrus:
The publishing to content API should be done through
the regular workflow, instead of custom implementation.


petrjasek:
what's the idea regarding configuration? would it require each subscriber to have a content api type destination? and it would push there an item per subscriber or just single one?

petrjasek:
closing it for now

PR #1248, opened 2018-04-18
SDESK-2763 Empty Qcode not checked for some vocabularies

mugurrus:


superdesk/vocabularies/vocabularies.py
```diff
@@ -64,7 +64,15 @@ class VocabulariesResource(Resource):
...
         'items': {
             'type': 'list',
-            'required': True
+            'required': True,
+            'schema': {
+                'type': 'dict',
+                'allow_unknown': True,
+                'schema': {
+                    'name': {'type': 'string', 'required': True, 'nullable': False},
+                    'qcode': {'type': 'string', 'required': True, 'nullable': False}
```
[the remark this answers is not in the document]
petrjasek: imo would be better to check it only if it's set required in vocabulary schema

superdesk/vocabularies/vocabularies.py
```diff
@@ -115,14 +124,24 @@ class VocabulariesService(BaseService):
 
     system_keys = set(DEFAULT_SCHEMA.keys()).union(set(DEFAULT_EDITOR.keys()))
 
+    def _validate_items(self, update):
+        if 'schema' in update and 'items' in update and \
+                ('name' in update['schema'] or 'qcode' in update['schema']):
```
petrjasek: what about setting what fields are required in the schema like:
```
{'name': {'required': true}, 'qcode': {'required': true}}
```
instead of hardcoding there validation for only name and qcode. so we could use it also for other vocabularies which don't use name/qcode but have required fields

superdesk/data_updates/00009_20180425-010702_vocabularies.py
```diff
@@ -0,0 +1,30 @@
...
+from superdesk.commands.data_updates import DataUpdate
+
+
+class DataUpdate(DataUpdate):
+
+    resource = 'vocabularies'
+    update_fields = ['name', 'qcode']
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        for vocabulary in mongodb_collection.find({}):
```
petrjasek: you could do `{'schema.qcode': {'$exists': True}}` filtering via query

PR #1255, opened 2018-04-26
[SDESK-2798] Add endpoint to support auto suggest for contact organis…

marwoodandrew:
…ation names

### DIFF
--- a/apps/contacts/service.py
+++ b/apps/contacts/service.py
@@ -12,6 +12,7 @@
 from superdesk.services import Service
 from superdesk.notification import push_notification
 from eve.utils import config
+from eve.utils import ParsedRequest
 
 
 class ContactsService(Service):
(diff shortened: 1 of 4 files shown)


mdhaman:
remove `'repo': 'contacts'` from the query string otherwise it looks fine.

apps/contacts/service.py
```diff
@@ -48,3 +49,29 @@ def on_deleted(self, doc):
...
+    def get(self, req, lookup):
+        """
+        Search for organisation matching the passed q parameter
+        :param req:
+        :param lookup:
+        :return: List of matching organisation strings
+        """
+        new_req = ParsedRequest()
+
+        q_str = 'organisation:' + '* organisation:'.join(req.args.get('q', '').split()) + '*'
```
petrjasek: is it supposed to be `organisation:* organisation:<q>*` ? I think [match_phrase_query](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-match-query.html#query-dsl-match-query-phrase-prefix) could be better for the prefix search

marwoodandrew: The idea is to wildcard each term. Not strictly a phrase search as I'm trying to match say "insurance australia" or "australian insurance". Each term is prefixed with "organisation:" An alternative might have been to set the default field to 'organisation'

petrjasek: I see, makes sense then

mdhaman:
@petrjasek don't merge this PR. We will merge it in the morning. @sjunaid was having issues with `endpoint name`.

marwoodandrew:
Changed the title of the end point.

PR #1257, opened 2018-05-01
SDANSA-113 Custom media field multi-items in the ninjs output

mugurrus:


petrjasek:
it's only tests right?

PR #1258, opened 2018-05-01
SDESK-2640 It is taking enormously long to open a published image file

mugurrus:


apps/archive/archive.py
```diff
@@ -197,6 +197,8 @@ def on_create(self, docs):
             self.validate_embargo(doc)
 
             update_associations(doc)
+            for assoc in doc.get(ASSOCIATIONS, {}).values():
+                assoc[config.LAST_UPDATED] = doc[config.LAST_UPDATED]
```
ioanpocol: Some tests fails because doc don't have config.LAST_UPDATED key.

PR #1274, opened 2018-05-21
[VOCABULARIES] fixed priority and urgency qcodes

jerome-poisson:
qcodes were stored as int instead of string for priority and urgency,
resulting in issues when trying to update vocabularies or to use
them in articles.

This patch fixes the prepopulate data and add a migration script to fix
existing values.

SDESK-2922

### DIFF
--- a/superdesk/vocabularies/vocabularies.py
+++ b/superdesk/vocabularies/vocabularies.py
@@ -71,7 +71,6 @@ class VocabulariesResource(Resource):
                 'allow_unknown': True,
                 'schema': {
                     'name': {'type': 'string', 'required': False, 'nullable': True},
-                    'qcode': {'type': 'string', 'required': False, 'nullable': True}
                 }
             }
         },


jerome-poisson:
@petrjasek why? qcode is an id isn't it?

jerome-poisson:
@petrjasek in NewsML, qcode is a string: https://iptc.org/std/NewsML-G2/2.27/specification/XML-Schema-Doc-Power/NewsItem.html#Link52 (it should even contain a \:\ there), I don't think an integer should be used in our case.

petrjasek:
yep qcode is a string, but not urgency or priority values

jerome-poisson:
but they are stored as qcode, and client send string when modifying/updating those values. To allow integer I would have to remove type from qcode, which would make validation less strict. I think that if we want to use integer we should use something else, not named "qcode" (e.g.: "value" would make sense).

petrjasek:
thing is this change might break some output formats, so far it was assuming there is int, and not sure how we could change it without breaking output formatting

jerome-poisson:
@petrjasek I've restored the former behaviour and qcode type is not checked anymore, fixing SDESK-2922, but this is IMO wrong: if user add `priority` or `urgency` by hand in metadata, new items will be sent by client as strings, resulting in a mix of int and string in mongo. This is the former behaviour and it's working (for now), but it's not good.

I keep thinking that my first patch is the right way to do, but I understand the worries to break publication workflow in production. But on the other hand, having a mix of int and string is a bomb which can explode later.

petrjasek:
right, I think we should specify the type in the [vocabulary schema](https://github.com/superdesk/superdesk-core/blob/master/apps/prepopulate/data_init/vocabularies.json#L82) and use that to validate/convert

jerome-poisson:
after talk with @petrjasek , I merge as it for now, and I'll create a separate PR later to handle type in vocabularies.

PR #1276, opened 2018-05-22
SDESK-715 Feeding service labels and fields should be set server side, not in client

mugurrus:


superdesk/io/feeding_services/rss.py
```diff
@@ -38,13 +38,46 @@ class RSSFeedingService(FeedingService):
...
+            'placeholder': 'Username', 'required_expression': '{auth_required}',
+            'show_expression': '{auth_required}'
+        },
+        {
+            'id': 'password', 'type': 'password', 'label': 'Password',
+            'placeholder': 'Password', 'required_expression': '{auth_required}',
+            'show_expression': '{auth_required}'
+        },
+        {
+            'id': 'field_aliases', 'type': 'tuple-multivalue', 'label': 'Content Field Aliases',
```
petrjasek: what about something like `type='mapping'` and instead of `first_field_label` having "recursive" fields:
```
{
    'label': 'Content Field Aliases',
    'type': 'mapping',
    'fields': [
        {'type': 'select', 'label': 'RSS field name', 'options': [...]},
        {'type': 'text', 'label': 'Superdesk field'},
    ],
}
```
I think this would make the api a bit cleaner and wouldn't require ui changes when we need more columns

petrjasek:
pls add some info into docs about the fields

PR #1282, opened 2018-05-24
[SDESK-2941] Change annotation ID to integer instead of string

pablopunk:
This should be merged together with client's PR


petrjasek:
I think this will need a change in [feature](https://github.com/superdesk/superdesk-core/blob/master/features/archive.feature#L780)

PR #1285, opened 2018-05-29
SDESK-716 We should be able to restrict feed parser for feeding services

mugurrus:


apps/io/feeding_services/wufoo.py
```diff
@@ -47,7 +47,7 @@ class WufooFeedingService(FeedingService):
         }
     ]
 
-    force_values = {'feed_parser': 'wufoo'}
+    restricted_values = {'feed_parser': ['wufoo']}
```
petrjasek: it there any other usecase for restricted value? I think it could be more explicit via:
```python3
feed_parser = 'wufoo'
```
but not sure if there would be a restriction to more than 1 parsers.
this should be also defined imo for tap, ritzau, rss and twitter.

mugurrus: The code supports now one value lists (in which case it hides the feed parser field) and also multi-valued lists, together with tests. I don't see any logical reason to reduce the functionality once it's already done.
From the task: "we know that a feeding service can only be used with some feed parsers". I understand from this text that there may be multiple values so I implemented it this way.

petrjasek: simpler code is easier to maintain and read, that would be the reason.
what about other restricted values? I don't see that in the task.

superdesk/errors.py
```diff
@@ -237,6 +241,10 @@ def parserNotFoundError(cls, exception=None, provider=None):
     def notConfiguredError(cls, exception=None, provider=None):
         return SuperdeskIngestError(2001, exception, provider)
 
+    @classmethod
+    def invalidFeedParserValue(cls, exception=None, provider=None, restricted_fields=None):
```
petrjasek: seems like `restricted_fields` param is not used

PR #1287, opened 2018-05-31
[MEDIA] new "media_editor" endpoint

jerome-poisson:
client can use "media_editor" API to transform an image.
A request need to specify "item_id" and an "edit" object with edition
instruction.

Edition instructions keys are operations names, and values are
operations parameters.

So far, the following operations are available:

- rotate (counter clockwise degrees)
- flip (value can be "vertical" or "horizontal")
- brightness (value is a float)
- contrast (value is a float)
- grayscale (value is not used)

In the answer, the API will return an "item" object with updated
renditions. Renditions media are temporary, and will be cleaned after at
most 24 hours.

SDANSA-160


petrjasek:
imo rotate should get a number, so in ui they can press rotate 1/2/3/4 times and it would still be single operation.

jerome-poisson:
@petrjasek OK I can change no problem.

To do a single operation, user should press several times in a short delay, and client would have to not do the API request before this delay, is it not complicating for not so usual case ?

jerome-poisson:
I've updated the P.R., now it access `item` and `item_id` so client can use intermediate item to modify again a previously modified image.
rotation now use counter clockwise degrees for parameter.
Also raise SuperdeskAPIError when param is not what is expected or if both `item` and `item_id` are missing

petrjasek:
afaik client will only send it when user is done with editing, it handles preview on its own

jerome-poisson:
updated again, to add `grayscale` operation which was missing, and also added some documentation to describe the API.

superdesk/media/media_editor.py
```diff
@@ -0,0 +1,118 @@
...
+            'mapping': not_analyzed
+        },
+        'edit': {
+            'type': 'dict',
+            'required': True,
+            'mapping': not_analyzed
+        }
+    }
+    internal_resource = False
+    resource_methods = ['GET']
```
petrjasek: why is it actually using `GET` method? imo it should be `POST` - imo it changes state

jerome-poisson: It doesn't store items in database, files are temporary, I though `GET` was more appropriate in this specific case.

petrjasek: I think we also use post with picture crop, so it would be more consistent
https://github.com/superdesk/superdesk-core/blob/master/apps/picture_crop/__init__.py#L50

jerome-poisson: alright, I'll change it.

jerome-poisson: @petrjasek done

PR #1289, opened 2018-05-31
[VOCABULARIES] fixed error sending on validation error

jerome-poisson:
error was missing payload, resulting in message not displayed in client
in case of validation error.

SDESK-2923

### DIFF
--- a/superdesk/vocabularies/vocabularies.py
+++ b/superdesk/vocabularies/vocabularies.py
@@ -140,7 +140,9 @@ class VocabulariesService(BaseService):
                 for field, desc in update.get('schema', {}).items():
                     if ((desc.get('required', False) or unique_field == field) and (
                             field not in item or not item[field])):
-                        raise SuperdeskApiError.badRequestError('Required ' + field + ' in item ' + str(index))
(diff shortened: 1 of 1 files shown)


superdesk/vocabularies/vocabularies.py
```diff
@@ -140,7 +140,9 @@ def _validate_items(self, update):
                 for field, desc in update.get('schema', {}).items():
                     if ((desc.get('required', False) or unique_field == field) and (
                             field not in item or not item[field])):
-                        raise SuperdeskApiError.badRequestError('Required ' + field + ' in item ' + str(index))
+                        msg = 'Required ' + field + ' in item ' + str(index)
+                        payload = {'validator exception': msg}
```
petrjasek: for translation it would be better if the error is constant and it gets the rest as params, something like:
```
payload = {'error': 'Required {{ field }} is missing in item {{ item }}', 'params': {'field': field, 'item': index}}
```
or even just send some error code and put the whole string on client (this will simplify translate string extraction):
```
payload = {'error': {'required_field': 1}, 'params': {'field': field, 'item': index}}
```

PR #1293, opened 2018-06-05
SDESK-2940 specify qcode type in vocabulary schema

mugurrus:


petrjasek:
it should be enough to put this into https://github.com/superdesk/superdesk-core/blob/master/apps/prepopulate/data_init/vocabularies.json data updates don't run on new instances afaik

PR #1299, opened 2018-06-07
[MEDIA EDITOR] added "saturation" filter and do permanent saving on e…

jerome-poisson:
…ach API call

SDESK-3010

### DIFF
--- a/superdesk/media/media_editor.py
+++ b/superdesk/media/media_editor.py
@@ -12,10 +12,9 @@
 from superdesk.resource import Resource, not_analyzed
 from superdesk.services import BaseService
 from superdesk.media.renditions import generate_renditions, get_renditions_spec
-from superdesk.utc import utcnow
 from superdesk import get_resource_service
 from superdesk import errors
-from flask import request, current_app
+from flask import current_app
 from PIL import Image, ImageEnhance
 from io import BytesIO
 from eve.utils import config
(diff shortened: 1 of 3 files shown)


superdesk/media/media_editor.py
```diff
@@ -72,52 +71,62 @@ def transform(self, im, operation, param):
...
         elif operation == 'grayscale':
             return im.convert('L')
+        elif operation == 'saturation':
+            return ImageEnhance.Color(im).enhance(float(param))
         logger.warning('unhandled operation: {operation} {param}'.format(
             operation=operation,
             param=param))
         return im
 
     def create(self, docs):
```
petrjasek: I this this could be put into a few methods, it's kind of hard to follow what it does

jerome-poisson: for now it's only calling one method by operation, I could put in separate methods, but that would be on line each time. Do you think it's would really improve readability ?

superdesk/media/media_editor.py
```diff
@@ -61,63 +60,97 @@ def transform(self, im, operation, param):
...
+                                             current_app.media.url_for_media,
+                                             temporary=True)
+
+            # we update item in db, and the item we'll return to client
+            updates = {'renditions': renditions}
+            ids.append(item_id)
+            archive.update(item_id, updates, item)
+            item.update(updates)
+
+            # we can now delete old media
```
petrjasek: it should be probably kept so it's visible in previous versions

jerome-poisson:
@petrjasek I'm not sure to get it right, there is not more references to those renditions, were should I link them ?

jerome-poisson:
and AFAIK history is not requested.

superdesk/media/media_editor.py
```diff
@@ -61,63 +60,97 @@ def transform(self, im, operation, param):
...
+            for operation, param in edit.items():
+                try:
+                    out = self.transform(out, operation, param)
+                except ValueError as e:
+                    raise errors.SuperdeskApiError.badRequestError('invalid edit instructions: {msg}'.format(msg=e))
+            buf = BytesIO()
+            out.save(buf, format=im.format)
+
+            # we set metadata
+            buf.seek(0)
```
petr: May be it's better to hide in specific function ? like 
```
def set_metadata
```
And you make code more readeable than. And of course you don't need to write this comment

petrjasek:
I mean to keep the links working for previous item versions

superdesk/media/media_editor.py
jerome-poisson: Hi, thanks for your feedback.
Your suggestion make sense, but I'm not really fond of using methods for 3 lines, I find it more readable to see what's happening directly without having to jump in the code.
Question of taste I guess.

PR #1301, opened 2018-06-11
[MEDIA EDITOR] use item from archive when item is given by client

jerome-poisson:
using item given by client was resulting in item not being updated
correctly. This patch fixes it by getting item from database before
updating it.

SDESK-3010


petrjasek:
just wondering - does it work with autosave changes?

jerome-poisson:
@petrjasek keep this answer from @pavlovicnemanja  for PR history:

> I think no, I will do further tests for it, but I think autosave is not causing any changes on images

jerome-poisson:
I'm waiting from green light from @pavlovicnemanja before merging, in case any change is needed.

jerome-poisson:
after tests with @pavlovicnemanja it seems OK, I'm merging now.

PR #1302, opened 2018-06-12
SDESK-3023 [INGEST] Ingest source creation: Configure restriction for feed parser dropdown per feeding service

ride90:
PR for client side https://github.com/superdesk/superdesk-client-core/pull/2369


ride90:
Instance for testing https://sdpr-2909.test.superdesk.org

PR #1304, opened 2018-06-15
SDESK-2483 [USER] [NINJS] Add new values to author's profile

ride90:


ride90:
Instance for testing with client side included https://sdpr-2914.test.superdesk.org/

superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -407,6 +407,9 @@ def _format_authors(self, article):
                 "name": user.get('display_name', author.get('name', '')),
                 "role": author.get('role', ''),
                 "biography": user.get('biography', author.get('biography', '')),
+                "facebook": user.get('facebook', author.get('facebook', '')),
```
petrjasek: I think we should only add those if it's non-empty, otherwise this might break some clients even if they don't use the feature

PR #1311, opened 2018-06-27
Error in data update 00013 20180618 090440 archive py script  (sdesk 3105)

tomaskikutis:
Couldn't find any issues with the upgrade script except for bad default value for `draftjsState`, but `len` of an empty object is 0 the same as for empty array, so it should not make a difference.

Added a try-catch to log the failing item on the next run

### DIFF
--- a/superdesk/data_updates/00013_20180618-090440_archive.py
+++ b/superdesk/data_updates/00013_20180618-090440_archive.py
@@ -21,34 +21,42 @@ class DataUpdate(DataUpdate):
             collection = mongodb_database[resource]
 
             for item in collection.find({'fields_meta': {"$gt": {}}}):
-                for field_meta_name in item.get('fields_meta', {}):
-                    draft_js_state = item['fields_meta'].get(field_meta_name, {}).get('draftjsState', [])
-                    entity_map = draft_js_state[0].get('entityMap', {}) if len(draft_js_state) > 0 else {}
 
-                    for entity_key in entity_map:
-                        entity = entity_map[entity_key]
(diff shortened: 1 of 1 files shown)


superdesk/data_updates/00013_20180618-090440_archive.py
```diff
@@ -21,34 +21,42 @@ def forwards(self, mongodb_collection, mongodb_database):
...
-                        if(selector is not None and selector.startswith('#qumu')):
-                            entity_map[entity_key] = {
-                                'type': 'EMBED',
-                                'mutability': 'MUTABLE',
-                                'data': {
+
+                try:
+                    for field_meta_name in item.get('fields_meta', {}):
+                        draft_js_state = item['fields_meta'].get(field_meta_name, {}).get('draftjsState', [])
+                        entity_map = draft_js_state[0].get('entityMap', {}) if len(draft_js_state) > 0 else {}
```
petrjasek: I think it would be better to do it like:
```
try:
    entity_map = ...
except KeyError:
    entity_map = {}
```
don't like the try catch for Exception on the whole thing

tomaskikutis: Why? I would agree if it was usual code, but for an upgrade scripts, I think it makes sense to wrap the entire thing, since we don't know what is in the DB and we want to log the item no matter what the issue is.

petrjasek: yep I think it's just bad practice, no matter where it is. but I can live with this, np

PR #1315, opened 2018-07-02
[SDESK-3113](fix): Adding registry for restricted parser for feeding service.

mdhaman:
Adding registry for the `restricted parsers` for feeding service so that different parser can be used for then`restricted parser feed service`.

@petrjasek @jerome-poisson

### DIFF
--- a/apps/io/feeding_services/reuters.py
+++ b/apps/io/feeding_services/reuters.py
@@ -16,7 +16,7 @@ from flask import current_app as app
 
 from superdesk.errors import IngestApiError
 from superdesk.etree import etree, ParseError
-from superdesk.io.registry import register_feeding_service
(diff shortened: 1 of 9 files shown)


petrjasek:
what does it actually allows that was not possible before?

mdhaman:
It allows assigning of overloaded parsers. It was allowed before till commit https://github.com/superdesk/superdesk-core/commit/6aeff1c2dffee587c53a1e3508fcd40109d44508 by @ride90 which enforces that `newsml2` parser to be used.
Currently is hard coded https://github.com/superdesk/superdesk-core/blame/master/apps/io/feeding_services/reuters.py#L68 . These changes will allow you to add new parser to `parser_restricted_values` for the feeding service.

PR #1316, opened 2018-07-02
Preview API

tomaskikutis:


superdesk/io/format_document_for_preview.py
```diff
@@ -0,0 +1,43 @@
...
+    subscriber_id = request.args.get('subscriber_id')
+    formatter_qcode = request.args.get('formatter')
+
+    subscriber = get_resource_service('subscribers').find_one(req=None, _id=subscriber_id)
+    doc = get_resource_service('archive').find_one(req=None, _id=document_id)
+
+    formatter = get_formatter(formatter_qcode, doc)
+    formatted_docs = formatter.format(article=apply_schema(doc),
+                                      subscriber=subscriber,
+                                      codes=None) # `codes` doesn't seem to be used in any formatter,
```
tomaskikutis: Can we remove the argument or is it used somewhere?

superdesk/io/format_document_for_preview.py
```diff
@@ -0,0 +1,43 @@
+from flask import Blueprint, request, Response, current_app as app
+import superdesk
+from superdesk import get_resource_service
+from superdesk.publish.formatters import get_formatter
+from apps.content_types import apply_schema
+
+
+bp = Blueprint('format_document', __name__)
+
+def get_mime_type(formatter_qcode):
```
tomaskikutis: Any suggestions on a cleaner way of doing this? Are there any other types except XML and JSON?

superdesk/io/format_document_for_preview.py
mdhaman: `codes` are used by https://github.com/superdesk/superdesk-aap/blob/master/server/aap/publish/formatters/anpa_formatter.py#L39. 

These `codes` are used by upstream distribution platform at AAP and they are calculated based on the products assigned to the subscriber.

Please don't remove the argument.

superdesk/io/format_document_for_preview.py
```diff
@@ -0,0 +1,43 @@
...
+from superdesk import get_resource_service
+from superdesk.publish.formatters import get_formatter
+from apps.content_types import apply_schema
+
+
+bp = Blueprint('format_document', __name__)
+
+
+def get_mime_type(formatter_qcode):
+    if(formatter_qcode == 'newsmlg2'):
```
tomaskikutis: Any suggestions on a cleaner way of doing this? Are there any other types except XML and JSON?

superdesk/io/format_document_for_preview.py
```diff
@@ -0,0 +1,43 @@
...
+
+    document_id = request.args.get('document_id')
+    subscriber_id = request.args.get('subscriber_id')
+    formatter_qcode = request.args.get('formatter')
+
+    subscriber = get_resource_service('subscribers').find_one(req=None, _id=subscriber_id)
+    doc = get_resource_service('archive').find_one(req=None, _id=document_id)
+
+    formatter = get_formatter(formatter_qcode, doc)
+    formatted_docs = formatter.format(article=apply_schema(doc),
```
petrjasek: I thought this will make a request to that `preview_endpoint_url` and return the response, not sure why it's doing the internal formatting?

tomaskikutis: It needs to return ninjs/newsml or whatever format is configured for that destination. How would I do it differently?

petrjasek: I thought that feature was about rendering html, not about those standard formats

tomaskikutis: The way I saw the problem is that clients want to see how would an article look like when published without actually publishing it. So I went to look at what is happening when actually publishing. I found that we format the item and post it to configured API, so I did the same only without marking it as published and using a different endpoint. It allows clients to reuse their already set up `standard format -> HTML` renderers and provides a realistic production-like preview since the code is going through the same paths.

Do you think it should be changed?

petrjasek: that makes sense, but I was expecting this to post the formatted item to configured url and return response from the external system

superdesk/io/format_document_for_preview.py
```diff
@@ -0,0 +1,43 @@
...
+
+@bp.route('/format-document-for-preview/', methods=['GET', 'OPTIONS'])
+def format_document():
+
+    document_id = request.args.get('document_id')
+    subscriber_id = request.args.get('subscriber_id')
+    formatter_qcode = request.args.get('formatter')
+
+    subscriber = get_resource_service('subscribers').find_one(req=None, _id=subscriber_id)
+    doc = get_resource_service('archive').find_one(req=None, _id=document_id)
```
petrjasek: imo we should send the item for preview in the payload, otherwise you would have to save before you can preview

superdesk/io/format_document_for_preview.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def get_mime_type(formatter_qcode):
+    if(formatter_qcode == 'newsmlg2'):
+        return 'text/xml'
+    else:
+        return 'application/json'
+
+
+@bp.route('/format-document-for-preview/', methods=['GET', 'OPTIONS'])
```
petrjasek: imo it should be using POST

superdesk/io/format_document_for_preview.py
ahilles107: Hey, 

this is exactly what we need for web preview with Publisher. It shouldn't try to return html as you don't know what is the real template. In targeted publishing preview you select website and section and with this data you can see real preview on it's final destination. 

When it can be merged?

tomaszrondio: All we need is an endpoint where I could send item guid and get back a package in same format that is sent when published. 
I get the item like that: `extensionPoints.register('authoring:publish', TargetedPublishing, {session: session, config: config}, ['item']);`

superdesk/io/format_document_for_preview.py
ahilles107: Why? Something is created on your side with it? What if we want to get it multiple times? I think that GET would be better here.

petrjasek: client sends a payload and gets it processed, imo that's what post is for, with get and sending only id it could get cached

ahilles107: POST should be used when: 
``` 
* Posting a message to a bulletin board, newsgroup, mailing list,
      blog, or similar group of articles;

* Creating a new resource that has yet to be identified by the
      origin server; and

* Appending data to a resource's existing representation(s).
```

So it depends if this endpoint creates some resource at backend. If it just process it every time based on provided parameters then GET will be better.

ahilles107: If I send a POST request twice with this same data then I should get another item in response or information about duplicated resource created. If it gives me this same (identical) response every time then it's GET.

petrjasek: imo it's more like:
```
- Providing a block of data, such as the result of submitting a
        form, to a data-handling process;
```
from https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

it depends if client will send just id or the full item, I think it should send the full item so you can get preview before saving the changes. and then it should be POST. if it's only id GET is fine

ahilles107:
Hey, what's the status of this PR?

tomaskikutis:
@ahilles107 I just got back from vacation today. I'm getting back to this PR, remembering the code and fixing a few things. I should be done with it this week unless there are significant additional requirements.

superdesk/io/format_document_for_preview.py
tomaskikutis: @tomaszrondio that's what `/format-document-for-preview/` does except that you also need to send `subscriber_id` and `formatter` in addition to guid. That's because when the item is published, it's published to multiple places in different formats, so you need to select the one you want.

You can check out the PR locally and see if the endpoint behaves like you want. Let me know if it doesn't.

superdesk/io/format_document_for_preview.py
tomaskikutis: Let's require saving https://github.com/superdesk/superdesk-client-core/pull/2429#discussion_r203343483

The publisher team needs the same API as well, so using an ID works for everyone. We can add another endpoint with different parameters if saving before previewing proves to be inconvenient.

superdesk/io/format_document_for_preview.py
tomaskikutis: As we are using the ID as explained https://github.com/superdesk/superdesk-core/pull/1316/files#r203367817 let's go with `GET`

PR #1317, opened 2018-07-02
fix(app): use older version of flask_pymongo

petrjasek:
superdesk/eve is not compatible with 2.0
@mdhaman this affects probably also 1.8


mdhaman:
@petrjasek it affects 1.8 as well.

petrjasek:
@mdhaman pushed it there

PR #1320, opened 2018-07-10
[DICTIONARIES] work around Eve behaviour which create sub-dict on eac…

jerome-poisson:
…h "." it find in keys

this works by using Flask's request object and overwritting the content
of updates (which comes after Eve) with it.

SDESK-3083

### DIFF
--- a/apps/dictionaries/service.py
+++ b/apps/dictionaries/service.py
@@ -13,7 +13,8 @@ import re
 import logging
 import collections
 
-from flask import json, current_app as app
+from flask import json, current_app as app, request
+from simplejson.errors import JSONDecodeError
 from eve.utils import config
 
 from superdesk.errors import SuperdeskApiError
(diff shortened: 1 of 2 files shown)


petrjasek:
can you add a test for it?

jerome-poisson:
@petrjasek I haven't made a test because the issue is due to behaviour with Eve during real request, and we need to use Flask to fix it, I'm not sure we can easily simulate that in a realistic way without launching client. I can try to mock, I'll check it out.

petrjasek:
imo it should be possible via behave test

jerome-poisson:
@petrjasek FYI I've added the tests. Indeed with behave it's doing the whole workflow, so tests make sense, thanks for the feedback.

tomaskikutis:
I've created an issue on eve repo to address this bug https://github.com/pyeve/eve/issues/1173

PR #1322, opened 2018-07-16
feat(login): Add support to expire the password after a given period

ioanpocol:


superdesk/default_settings.py
```diff
@@ -518,6 +518,9 @@ def local_to_utc_hour(hour):
 #: Set the timezone celery functions to UTC to avoid daylight savings issues SDESK-1057
 CELERY_TIMEZONE = 'UTC'
 
+#: The number of days after the user has to change his password
```
petrjasek: imo it should be set to 0 by default so we don't change the behaviour

apps/auth/db/db.py
```diff
@@ -12,12 +12,15 @@
...
+from apps.auth.errors import CredentialsAuthError, PasswordExpiredError
+from flask import current_app as app
+from superdesk.utc import utcnow
+import datetime
 
 
 class DbAuthService(AuthService):
 
-    def authenticate(self, credentials):
+    def authenticate(self, credentials, ignoreExpire=False):
```
petrjasek: the param should be probably `ignore_expire` by python convections

PR #1323, opened 2018-07-18
[SDESK-3145] Move aggregations to be a member of the Search endpoint class

MarkLark86:
* This gives the ability to inherit from this class and implement custom aggregation/searching functionality (as is used in analytics/source_category_report
* Use default repos if none supplied (or if repos is empty array after filtering them)

### DIFF
--- a/apps/search/__init__.py
+++ b/apps/search/__init__.py
@@ -10,12 +10,13 @@
 
 from flask import current_app as app, json, g
 from eve_elastic.elastic import set_filters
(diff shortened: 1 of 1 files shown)


MarkLark86:
@petrjasek @mugurrus 
What are your thoughts/concerns about using a class level member for controlling the aggregations in the search endpoint? As an example, please see https://github.com/superdesk/superdesk-analytics/pull/38 on this will work

PR #1327, opened 2018-07-24
feat(output): push event and coverage ids to content api

petrjasek:
it's needed in newsroom to be able to show links from wire to agenda

### DIFF
--- a/content_api/items/resource.py
+++ b/content_api/items/resource.py
@@ -98,6 +98,10 @@ schema = {
     'wordcount': metadata_schema['word_count'],
     'charcount': {'type': 'integer'},
     'readtime': {'type': 'integer'},
+    'event_id': metadata_schema['guid'],
+    'planning_id': metadata_schema['guid'],
+    'coverage_id': metadata_schema['guid'],
+    'agenda_id': metadata_schema['guid'],
 }
 
 
--- a/content_api/publish/service.py
+++ b/content_api/publish/service.py
@@ -15,7 +15,7 @@ from eve.utils import config, ParsedRequest
 
 from superdesk.utc import utcnow
 from superdesk.services import BaseService
-from superdesk.publish.formatters.ninjs_formatter import NINJSFormatter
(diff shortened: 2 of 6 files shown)


superdesk/metadata/item.py
```diff
@@ -589,6 +589,9 @@
         'type': 'string',
         'mapping': not_analyzed,
     },
+
+    # event ref
+    'event_id': {'type': 'string', 'mapping': not_analyzed},
```
mdhaman: We already have `event_id` on the archive schema (https://github.com/superdesk/superdesk-core/blob/master/apps/archive/common.py#L123). We did think of using `event_id` for pushing to Newsroom but it had the following issues:
- Users could link an item in Superdesk after it is published to Newsroom.  To achieve this options were to do correction (which business won't like) or resend to Newsroom subscriber only.
- Not all items in Superdesk belong to an event. For example, travel stories don't belong to an event.
- Users can create a Planning item from the story and retrospectively create an event as well.

We took another approach of linking `wire and agenda` via planning item. If the coverage is linked to an item in Superdesk, you would have item id in the planning JSON to Newsroom. (see here https://github.com/superdesk/superdesk-planning/blob/master/server/planning/output_formatters/json_planning.py#L97) . Please find the example of linking over here https://gist.github.com/mdhaman/73bb2081379c506542315937d5137187#file-planning-json-L34

marwoodandrew: The planning JSON sent to Newsroom should contain all the delivery information for the item. Newsroom should able to maintain an equivalent of the delivery collection. The delivery info allows bidirectional linking between planning and content items and from planning to it's associated event (admittedly a double hop).

petrjasek: ok cool, this won't be needed then

petrjasek: seems like this would be still needed if we want to link published items without any planning item update. and given we have all the metadata I would put it to the output

PR #1330, opened 2018-07-31
[VALIDATION] [MEDIA] media metadata validation

jerome-poisson:
VALIDATOR_MEDIA_METADATA is a schema indicating which metadata are
available and which fields are required, it was previously in client and
has been moved to backend. Client can retrieve this data using deployConfig.

During validation, if associations are present, those metadata are
checked, and an error message is returned if any required field is
missing.

SDFID-431

### DIFF
--- a/apps/validate/validate.py
+++ b/apps/validate/validate.py
@@ -21,6 +21,7 @@ from superdesk import get_resource_service
 from _collections_abc import MutableMapping
 
 REQUIRED_FIELD = 'is a required field'
+MAX_LENGTH = "max length is {length}"
 STRING_FIELD = 'require a string value'
 DATE_FIELD = 'require a date value'
 REQUIRED_ERROR = '{} is a required field'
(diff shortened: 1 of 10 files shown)


apps/validate/validate.py
```diff
@@ -128,6 +128,19 @@ def _validate_company_codes(self, *args):
         """Ignore company codes."""
         pass
 
+    def _validate_media_metadata(self, validate, associations_field, associations):
```
mdhaman: Current implementation checks for max_length as well (see here https://github.com/superdesk/superdesk-aap/blob/master/client/superdesk.config.js#L6) and in the client-core config only required field validation was there. 
@jerome-poisson Can you please confirm that this feature is not removed?

jerome-poisson: @mdhaman nothing has been removed, but `validatorMediaMetadata` must be moved from client to backend in `superdesk-aap`. The client will then get the whole object from backend.

Also only `required` is checked in backend so far, I'll add maxlenght if it's present (but it's still checked in client if it was before).

PR #1333, opened 2018-08-07
(feat): Executing on stage macro when content is created via scheduled template.

mdhaman:
This change is required by [SDESK-3199].

@petrjasek / @marwoodandrew other option would be to have "macro" settings on the template editor but the outcome would be the same. Please let me know your thoughts?

### DIFF
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -12,6 +12,7 @@ import flask
 import logging
 import datetime
 import superdesk
+from copy import copy, deepcopy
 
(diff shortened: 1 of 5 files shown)


petrjasek:
I think it would also make sense to be called in archive service on_created so it's consistent no matter where the item comes from

mdhaman:
I think it might be dangerous to apply marco on archive service on_created i.e. manually and fetched content as well. Idea was to apply marco on anything automatically created or routed.

PR #1335, opened 2018-08-08
(feat) Executing on stage macro when content is created via scheduled template.

mdhaman:


mdhaman:
@petrjasek same issue as @marwoodandrew's  PR in client core https://github.com/superdesk/superdesk-client-core/pull/2484

PR #1338, opened 2018-08-10
fix(images): If the image operation can't be applied ignore it(SDESK-…

ioanpocol:
…3207)

### DIFF
--- a/superdesk/media/media_editor.py
+++ b/superdesk/media/media_editor.py
@@ -116,8 +116,12 @@ class MediaEditorService(BaseService):
             for operation, param in edit.items():
                 try:
                     out = self.transform(out, operation, param)
-                except ValueError as e:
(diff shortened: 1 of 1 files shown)


superdesk/media/media_editor.py
```diff
@@ -116,8 +116,9 @@ def create(self, docs):
             for operation, param in edit.items():
                 try:
                     out = self.transform(out, operation, param)
-                except ValueError as e:
-                    raise errors.SuperdeskApiError.badRequestError('invalid edit instructions: {msg}'.format(msg=e))
+                except ValueError:
+                    # if the operation can't be applied just ignore it
+                    None
```
petrjasek: there is [pass](https://docs.python.org/3.2/reference/simple_stmts.html#the-pass-statement) for noop

PR #1340, opened 2018-08-15
[SDESK-2943](fix): Extending internal destination to publish the duplicate item.

mdhaman:


superdesk/internal_destinations.py
```diff
@@ -45,13 +51,32 @@ def handle_item_published(sender, item, **extra):
...
+            'overwrite_item': overwrite_item,
+            'duplicate': True
+        }
+
+        send_to(new_item, desk_id=dest['desk'], stage_id=dest.get('stage'))
 
         if dest.get('macro'):
             macro = macros_service.get_macro_by_name(dest['macro'])
-            macro['callback'](new_item)
+            macro['callback'](new_item, publish_args=publish_args)
```
petrjasek: I would rather call it with `**publish_args` not sure if we have to wrap it inside a dict

mdhaman: `publish_args.duplicate` value is updated by the macro. If I do `**` it will unpack the dict and `duplicate` check in the next line fails and then the macro duplicates & publishes the item as well as `handle_item_published` method duplicate the item. We end up with 2 items one published and other duplicate item.

petrjasek: maybe this would be better done via return value or raising some `StopDuplication` exception
@jerome-poisson any feedback on this?

petrjasek: I would add `class StopDuplication` to this module and probably this:
```
try:
    macro['callback'](new_item, overwire_item=overwire_item)
except StopDuplication:
    return
```
imo that's more explicit

mdhaman: `StopDuplication` sound fine to me.

PR #1349, opened 2018-08-24
SDESK-3092 IDML formatter

ride90:


apps/prepopulate/data_init/vocabularies.json
```diff
@@ -168,7 +168,8 @@
                     {"name": "NINJS", "qcode": "ninjs"},
                     {"name": "NewsML G2", "qcode": "newsmlg2"},
                     {"name": "Email", "qcode": "email"},
-                    {"name": "Newsroom NINJS", "qcode": "newsroom ninjs"}
+                    {"name": "Newsroom NINJS", "qcode": "newsroom ninjs"},
+                    {"name": "IDML", "qcode": "idml"}
```
tomaskikutis: `"Adobe InDesign"` sound like a better name :)

ride90: @tomaskikutis , yes, it sounds more familiar for a users.  Adobe indesign can also be a *.indd file. In this case we can name it as Adobe InDesign (IDML).

ride90:
Hi guys, please review.

Instance for testing: https://sdspr-1349.test.superdesk.org
I've configured `EDITOR3 TEMPLATE`, so it has all features in body_html.
There is a configured subscriber `For testing`, it has 2 destinations file and ftp configured, other transmitters don't support encoded item 😢 
If you want try to use "Breaking news" story in politic desk. 
Keep in mind that IDML does not support some features (embed for example), so it will cut off such things. 
Also I think you need inDesign to open idml :) 
P.S. If you want to see what is inside - just change file extention to zip.

PR #1357, opened 2018-09-18
[INGEST] SDNTB-512 NTB: Ritzau ingest: ingest editorial note from the webfeed

ride90:


ride90:
Hi @jerome-poisson , I've just started to play with ingest. Please, review this pr.

tests/io/feed_parsers/ritzau_test.py
```diff
@@ -50,3 +52,13 @@ def test_content(self):
                                             'kritik, fordi de har finansieret lignende forsøg.</p>')
         self.assertEqual(item['guid'], '9a6955fc-11da-46b6-9903-439ebb288f2d')
         self.assertEqual(item['firstcreated'].isoformat(), '2018-01-30T17:32:18.397000+01:00')
+        self.assertNotIn('ednote', item)
+
+    def test_ednote(self):
+        self._parse_file('ritzau_news_test_ednote.xml')
+        self.assertEqual(
+            self.item['ednote'],
+            'Det meste af artiklen er skrevet om/skrevet op:Mere info om Emmy Awards er tilføjet i de to første afsnit.'
```
petrjasek: imo would be good to add line breaks for `<br>` and after `<p>` tags,
we have a [helper for that](https://github.com/superdesk/superdesk-core/blob/master/superdesk/text_utils.py#L36)

ride90:
@jerome-poisson , I've removed `br` from `BLOCK_ELEMENTS`, since `BLOCK_ELEMENTS` is used only in `parse_html`. 
I'm not sure why, but behave test `features/places.feature` fails here, but it passes on my local.

jerome-poisson:
@ride90 you did well, if BLOCK_ELEMENT is used somewhere else in the future, it's better that it doesn't include `br`

PR #1360, opened 2018-09-19
[FORMATTER] [NINJS] Put custom media fields in "extra_items" field instead…

jerome-poisson:
… of "associations"

This patch implement a request from publisher team: instead of having `my_custom_media--1` and `my_custom_media--2` in `associations`, we have now `"extra_items": {"my_custom_media": {"items":[{…}]}}` so we know that items are linked to `my_custom_media`.

This is done only for custom media field and doesn't affect other
associations.

SDESK-2955

### DIFF
--- a/docs/superdesk-ninjs-schema.json
+++ b/docs/superdesk-ninjs-schema.json
@@ -311,6 +311,22 @@
 				}
 			}
 		},
+		"extra_items" : {
+			"description": "Custom items grouped by ids",
+			"type": "object",
+			"properties" : {
+				"type" : "object",
+				"properties" : {
+					"items" : {
+						"description": "Items linked together",
+						"type" : "array",
+						"items": {
+							"type": "object"
+						}
+					}
+				}
+			}
+		},
 		"extra" : {
 			"description" : "Extra metadata",
 			"type": "object",
@@ -415,11 +431,11 @@
 						"description" : "Author's biography",
 						"type" : "string"
 					},
-                    "avatar_url": {
-                        "description": "Author's avatar url",
-                        "type": "string",
-                        "format": "uri"
-                    }
+					"avatar_url": {
+						"description": "Author's avatar url",
+						"type": "string",
+						"format": "uri"
+					}
 				}
 			}
 		},
--- a/superdesk/publish/formatters/ninjs_formatter.py
+++ b/superdesk/publish/formatters/ninjs_formatter.py
@@ -32,6 +32,7 @@
 import json
 import superdesk
 import logging
+import re
 
 from eve.utils import config
 from superdesk.publish.formatters import Formatter
@@ -44,6 +45,11 @@ from apps.archive.common import get_utc_schedule
 from superdesk import text_utils
 
 logger = logging.getLogger(__name__)
+# this regex match the way custom media fields are put in associations (i.e. how the key
+# is generated). This is legacy, and can hardly be changed without risking to break
+# instances in production.
+MEDIA_FIELD_RE = re.compile(r"(?P<field_id>\S+)--(?P<version>\d+)$")
+EXTRA_ITEMS = "extra_items"
 
 
 def filter_empty_vals(data):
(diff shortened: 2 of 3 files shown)


docs/superdesk-ninjs-schema.json
```diff
@@ -311,6 +311,26 @@
 				}
 			}
 		},
+		"groups" : {
```
petrjasek: I was wondering about the usage, maybe it would be simple if it's more like associations:
```
{
    "groups": {
        "slideshowA": {
            "items": [...],
        "slideshowB": {
            "items": [],
         }
    }
}
```
there can be only single group with given id anyhow, so the list is not needed, and this would be more consistent.
what do you think?
cc @takeit

takeit: I think that the:

```
{
    "groups": [
        {
            "id": "slideshowA",
            "items": []
        },
        {
            "id": "slideshowB",
            "items": []
        }
    ]
}
```

should be also fine. We support it like that in the Publisher right now but your option @petrjasek also makes sense, but we would need to make some changes to support it the way you suggested now.

petrjasek: I didn't realize we already use groups for packages, then it makes sense to keep it a list.
just not sure if we should combine those 2 things (packages and multi item fields)

jerome-poisson: @petrjasek it was actually your suggestion ;) https://dev.sourcefabric.org/browse/SDESK-2955?focusedCommentId=107958&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-107958

petrjasek: I know, that doesn't mean necessary it's the right thing to do :wink:

jerome-poisson: OK so where should I put that ? It's trivial to change the name, but we need to agree on something. Something explicit like `custom_media`, or something more generic?

jerome-poisson: I'll go with `custom_fields` if nothing better arise, and use a dict instead of list.

jerome-poisson: @petrjasek @takeit are you fine with that?

```javascript
{"custom_fields":
    {"custom_media_id_1": [{"type": "picture", …}, …],
     "custom_media_id_2": […]
    }
}
```

jerome-poisson: or `custom_items` to show explicitly that the values are items.

takeit: so what will be the final format? `groups` will be renamed to `custom_fields`? because I'm lost now :)

takeit: @jerome-poisson the `custom_items` name is fine, it's easy to change, but would that be a list at the end or not?

petrjasek: @jerome-poisson I would put the items in it into `{"items": []}` to leave space for some additional metadata if needed.

petrjasek: @takeit I think `groups` will stay for packages, and `custom_items` will be for multi item custom fields

petrjasek: maybe we could call it `extra_items` - we put the other custom fields in `extra` so it would be kind of consistent

jerome-poisson: @takeit @petrjasek it would be a list. So like that:
```javascript
{
    "extra_items":
    {
        "custom_media_id_1":
        {
            "items": [{"type": "picture", …}, …]
        }
        "custom_media_id_2":
        {
            "items": […]
        }
    }
}
```

takeit: okay looks good to me 👍

jerome-poisson:
@petrjasek I've updated, is it OK now?

PR #1361, opened 2018-09-19
[INGEST] SDNTB-514 NTB: Ritzau ingest: Priority should also be ingested as urgency.

ride90:


tests/io/feed_parsers/ritzau_test.py
```diff
@@ -53,6 +53,8 @@ def test_content(self):
         self.assertEqual(item['guid'], '9a6955fc-11da-46b6-9903-439ebb288f2d')
         self.assertEqual(item['firstcreated'].isoformat(), '2018-01-30T16:32:18.397000+00:00')
         self.assertNotIn('ednote', item)
+        self.assertEqual(item['priority'], '3')
```
petrjasek: I think both should be actually integers

ride90: @petrjasek it's done.

PR #1363, opened 2018-09-24
feat(saved search): Add desk subscriptions support for saved search (…

ioanpocol:
…SDFID-495)

### DIFF
--- a/apps/saved_searches/__init__.py
+++ b/apps/saved_searches/__init__.py
@@ -95,6 +95,29 @@ def publish_report(user_id, search_data):
     send_report_email(user_id, search_data, docs)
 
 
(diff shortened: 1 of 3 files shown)


apps/saved_searches/__init__.py
```diff
@@ -119,6 +120,22 @@ def report():
                     user_data['last_report'] = now
                     user_data['next_report'] = get_next_date(scheduling)
                     do_update = True
+
+            subscribed_desks = search['subscribers'].get('desk_subscriptions', [])
+            for desk_data in subscribed_desks:
```
petrjasek: seems like there is a lot of duplicated code in this and previous `for` cycle, would it be possible to avoid that?

PR #1366, opened 2018-09-25
[IDML] SDESK-3092 Move idml convertion logic into separate class.

ride90:


ride90:
Hi @petrjasek , here is an easiest workaround for the issue we discovered with formatter lately.

PR #1367, opened 2018-09-25
[ES UTILS] implements dates filters

jerome-poisson:
Date filters are needed for saved searches report, this patch implements
it.
The date format can be specified in frontend (i.e. day first or month
first), but here we always keep day first, as the date is used for wire
format between client and backend, and we have no way to know which
format is used in client.

Also added `python-dateutil` to requirements : it is used in many
location in superdesk-core but was missing in `setup.py`.

SDESK-3425

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ install_requires = [
     'chardet<4.0',
     'pymongo>=0.4,<4.0',
     'croniter<0.4',
-    'python-dateutil',
+    'python-dateutil<2.8',
 ]
 
 package_data = {
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ install_requires = [
     'chardet<4.0',
     'pymongo>=0.4,<4.0',
     'croniter<0.4',
+    'python-dateutil',
 ]
 
 package_data = {
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,4 +1,5 @@
 [flake8]
 max-line-length=120
 exclude=env,bin,lib,include,src,docs
-ignore=F811,D200,D202,D205,D400,D401,D100,D101,D102,D103,D104,D105,D107,W503
+ignore=F811,D200,D202,D205,D400,D401,D100,D101,D102,D103,D104,D105,D107,W503,W504,W605,F401,E261,F841
+# W504, W605, F401, E261 and F841 are temporarly ignored, due to recent changes in flake8
--- a/superdesk/es_utils.py
+++ b/superdesk/es_utils.py
@@ -13,11 +13,13 @@
 
 import logging
 import json
+import pytz
+from datetime import datetime
 from superdesk import app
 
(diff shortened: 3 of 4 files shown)


ioanpocol:
@jerome-poisson  Why did you changed ' to " for strings since all over the application we use ' ?

jerome-poisson:
that's because I've been using "black" formatting tool, which is among other things replace double quotes by single ones.

jerome-poisson:
@petrjasek @ioanpocol I've split the commit in 2, it should be more easy to review now.

ioanpocol:
@jerome-poisson Until now as a code style we used ' instead of ". This is the promoted choice by python.
I think that this kind of changes should be discussed and agreed. If we decide to change that I think we have to do it in all files not only in one file.
On the other hand there is a solution for this in black, read last comment here:
https://github.com/ambv/black/issues/118

setup.py
```diff
@@ -47,6 +47,7 @@
     'chardet<4.0',
     'pymongo>=0.4,<4.0',
     'croniter<0.4',
+    'python-dateutil',
```
petrjasek: can you set some version that's tested to work here?

jerome-poisson:
@ioanpocol where is it written than we are using `'` ? I've seen both types of quote in backend, I don't think anything has been enforced in the team.

Also where do you see this is promoted by Python? I've never heard of that and PEP 8 states `This PEP does not make a recommendation for this.` (cf. https://www.python.org/dev/peps/pep-0008/#string-quotes)

setup.py
jerome-poisson: @petrjasek done

jerome-poisson:
@petrjasek @ioanpocol I've added W504 and W605 to ignore list has they are causing trouble while it was OK before. I see they have been added to pycodestyle (cf. http://pycodestyle.pycqa.org/en/latest/developer.html#id2, but W504 should be disabled by default) and flake8 has been updated today.

They should be removed from ignore list, but this should be in a separated PR IMHO.

ioanpocol:
>@ioanpocol where is it written than we are using ' ? I've seen both types of quote in backend, I don't think anything has been enforced in the team.

We have 71203 ' vs 10507 ", most of " are for exception texts and json text used for tests.
I think this is the first PR that changed all ' in a file.
I still think that we need to have a discussion and if we decide to use " instead ' we have to replace in all places.

>Also where do you see this is promoted by Python? I've never heard of that and PEP 8 states This PEP does not make a recommendation for this. (cf. https://www.python.org/dev/peps/pep-0008/#string-quotes)

>This PEP does not make a recommendation for this. Pick a rule and stick to it.
I think we had a rule and you just changed it but only in one file.
Somehow the use of ' has been promoted by python community, it is used in most of the examples, tutorials, documentation and even in python interpreter: 
>>> "string"
'string'

PR #1368, opened 2018-09-25
[SAVED SEARCHES] decode "filter" before actually updating the item

jerome-poisson:
"filter" was decoded after updating the item, to be returned decoded to
the client. But this was resulting in invalid ETAG. This patch changes
this behaviour by decoding "filter" before actually updating the item.

SDESK-3435

### DIFF
--- a/apps/saved_searches/saved_searches.py
+++ b/apps/saved_searches/saved_searches.py
@@ -227,11 +227,11 @@ class SavedSearchesService(BaseService):
         return super().get(req, lookup=None)
 
     def update(self, id, updates, original):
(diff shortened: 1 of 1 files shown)


jerome-poisson:
@petrjasek can you check this one please? cf.  https://dev.sourcefabric.org/browse/SDESK-3435?focusedCommentId=109079&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-109079  Thanks!

petrjasek:
could you add some test pls where we could see what it fixes?

PR #1375, opened 2018-10-03
FTP ingest: use LIST command as a fallback when MLSD is not supported

GyanP:
https://dev.sourcefabric.org/browse/SDESK-3261

### DIFF
--- a/superdesk/io/feeding_services/ftp.py
+++ b/superdesk/io/feeding_services/ftp.py
@@ -116,7 +116,10 @@ class FTPFeedingService(FeedingService):
         except IngestFtpError:
             raise
         except Exception as ex:
-            raise IngestFtpError.ftpError(ex, provider)
+            if '500' in str(ex):
+                ftp.nlst()
+            else:
+                raise IngestFtpError.ftpError(ex, provider)
 
     def _move(self, ftp, src, dest):
         """Move distant file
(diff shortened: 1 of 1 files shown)


superdesk/io/feeding_services/ftp.py
```diff
@@ -246,7 +256,44 @@ def _update(self, provider, update):
...
+                file_name_list = []
+                date_list = []
+                ftp.dir(file_list.append)
+                for line in file_list:
+                    col = line.split()
+                    date_string = '{} '.format(datetime.now().year) + ' '.join(col[5:8]) + ':00'
+                    date_list.append(date_string)
+                    file_name_list.append(col[8])
+
+                for filename, modify_date in zip(file_name_list, date_list):
```
petrjasek: I think you can take it a bit further and create method for getting list of filenames with last modified via mlsd or dir, and then the file processing can be there just once with checking if filename is allowed, the last updated check, and the file moving bit on error. like:
```python3
def list_items(self, ftp):
    try:
        return [(filename, facts['modify']) for filename, facts in ftp.mlsd() if facts.get('type') == 'file']
    except: Exception:
        if '500' in str(ext):
            file_list = []
            ...
            return zip(file_name_list, date_list)
```

PR #1389, opened 2018-10-10
Ability to reuse get_parent_code method.

ride90:
I would like to use `get_parent_code` method only, without creating a `SubjectIndex` instance.

### DIFF
--- a/superdesk/io/subjectcodes.py
+++ b/superdesk/io/subjectcodes.py
@@ -16,13 +16,12 @@ from datetime import datetime
 from eve.render import send_response
 
 
+NEWSCODE_PATTERN = re.compile('[0-9]{8,9}')
 bp = Blueprint('subjectcodes', __name__)
 
 
 class SubjectIndex():
     """Subjects index."""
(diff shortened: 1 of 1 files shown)


petrjasek:
imo it would be better to put it outside the class, it doesn't have to be there

superdesk/io/subjectcodes.py
```diff
@@ -45,22 +43,24 @@ def get_items(self):
...
-        if code[-3:] != '000':
-            parent_code = code[:5] + '000'
-        elif code[2:5] != '000':
-            parent_code = code[:2] + '000000'
+def get_parent_subjectcode(code):
+    """Compute parent code for iptc newscode.
+
+    :param code: iptc newscode without `subj:` prefix
+    """
+    newscode_pattern = re.compile('[0-9]{8,9}')
```
petrjasek: I would put it somewhere on top of the file like `NEWSCODE_PATTERN = re.compile(...)` so we don't compile it all the time

PR #1390, opened 2018-10-11
[SDESK-3279] Remove planning related types from 'type' vocabulary as it will be loaded from planning module

nrvikas:


petrjasek:
btw how it will be loaded?

nrvikas:
It will be loaded when the planning module initialises as per this PR:
https://github.com/superdesk/superdesk-planning/pull/1025

PR #1396, opened 2018-10-29
[FEED PARSER][STT NEWSML]implemented new fields/modifications request…

jerome-poisson:
…ed by STT

New fields have been added to STT parser.
A subject scheme can now be specified in SUBJ_QCODE_PREFIXES.
STT NewsML parser can now map location to SD place according to STT_LOCATION_MAP.

SDESK-3507 STTNHUB-14 STTNHUB-18

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -711,3 +711,11 @@ VALIDATOR_MEDIA_METADATA = {
         "required": False,
     },
 }
+
+#: Behaviour on missing vocabulary, only used in (STT)NewsML G2 for now
+#: if "reject", missing vocabulary are rejected
+#: if "create", a new vocabulary is created
+#: if "continue" (default), value is returned but not created (it will be present in the resulting item, but the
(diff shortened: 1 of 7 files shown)


jerome-poisson:
I've put the `WIP` label because the location mapping is still in discussion. Also I may create vocabulary automatically when an unknown qcode is found.

jerome-poisson:
Note: this PR is ready and can be reviewed, I'm just keeping the `WIP` label because I'm not sure what to do with `sttcontinent` metadata, I'm currently checking with @karelpetrak .

ride90:
@jerome-poisson why in some places you are using `self.__class__.missing_voc` with `__class__` statement and in some places without?

jerome-poisson:
@ride90 because once the setting is read, it wont change, so I save it in a class attribute, and I need to do it via `self.__class__` (else it would change the instance attribute, not the class one).
When reading the variable, we don't have this problem, so I use `self.missing_voc directly`.

PR #1405, opened 2018-11-07
[INGEST] [FTP] SDESK-3580 Fix FTP feeding service.

ride90:


ride90:
- added `ftp_ninjs` feeding service to the list of providers which will be created when `@provider` 
  decorator is used.
- `step_impl_run_update_ingest_command` step is created. Sometimes it's good to execute `update_ingest.UpdateIngest().run()` during testing, because it executes internal logic inside feeding_services unlike `step_impl_fetch_from_provider_ingest`

⚠️ `step_impl_run_update_ingest_command` tested only with ftp (needed objects are mocked and etc), `_retrieve_and_parse` is mocked too, so parser doesn't execute. There is a TODO, this part can be improved for sure. 

💡 I think it makes sense to pass argument to `@provider` decorator, so we can specify what feeding service we need during a test (if nothing is specified we can create all). It will save us time during writing and running tests, because posting providers to resource takes time. I've noticed a big difference when I commented rest of the providers.

PR #1409, opened 2018-11-13
[INGEST][REGISTRY] SDESK-3582 Don't store feeding service instance in registry.

ride90:
There is no sense to store feeding service instance in **registered_feeding_services** https://github.com/superdesk/superdesk-core/blob/master/superdesk/io/registry.py#L42, would be enough to store just class.
BTW. We create a new feeding service instance all the time during update_provider task https://github.com/superdesk/superdesk-core/blob/master/superdesk/io/commands/update_ingest.py#L234

⚠️ This PR should not brake any instances (I hope 🤞). Please, check if custom instances use **registered_feeding_services** somehow. For example **ntb** does not. 

⚠️ ⚠️ Planning component requires PR, because **registered_feeding_services** used in tests.

### DIFF
--- a/superdesk/tests/steps.py
+++ b/superdesk/tests/steps.py
@@ -583,8 +583,7 @@ def fetch_from_provider(context, provider_name, guid, routing_scheme=None, desk_
     else:
         rule_set = None
 
-    provider_service = registered_feeding_services[provider['feeding_service']]
-    provider_service = provider_service.__class__()
+    provider_service = registered_feeding_services[provider['feeding_service']]()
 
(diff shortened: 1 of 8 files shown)


superdesk/tests/steps.py
```diff
@@ -583,8 +583,7 @@ def fetch_from_provider(context, provider_name, guid, routing_scheme=None, desk_
     else:
         rule_set = None
 
-    provider_service = registered_feeding_services[provider['feeding_service']]
-    provider_service = provider_service.__class__()
+    provider_service = registered_feeding_services[provider['feeding_service']]()
```
petrjasek: imo would be good to have some `get_feeding_service(name)` method that would return the instance, so we don't have to bother whenever we need it

PR #1411, opened 2018-11-14
fix(related content): Define related content data type (SDESK-3600)

ioanpocol:


apps/validate/validate.py
```diff
@@ -65,6 +65,10 @@ def _validate_type_media(self, field, value):
         """Allow type media in schema."""
         pass
 
+    def _validate_type_related_content(self, field, value):
+        """Allow type media in schema."""
```
jerome-poisson: this comment looks like a copy/paste from following one which has not been modified.

ioanpocol: Fixed

PR #1412, opened 2018-11-14
[FORMATTER] SDESK-3581 Don't instantiate formatter in FormatterRegistry metaclass.

ride90:
`FormatterRegistry` metaclass creates an instance of a formatter and stores it in the `formatters` list.
We can store class in the formatters and create an instance inside `get_formatter` only when we need it. Formatter instance creation is cheap. Having the same object all the time limits using `self` inside formatter.

### DIFF
--- a/superdesk/publish/formatters/__init__.py
+++ b/superdesk/publish/formatters/__init__.py
@@ -26,7 +26,7 @@ class FormatterRegistry(type):
         """Register sub-classes of Formatter class when defined."""
         super(FormatterRegistry, cls).__init__(name, bases, attrs)
         if name != 'Formatter':
-            formatters.append(cls())
+            formatters.append(cls)
 
 
 class Formatter(metaclass=FormatterRegistry):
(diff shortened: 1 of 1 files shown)


superdesk/publish/formatters/__init__.py
```diff
@@ -147,14 +147,15 @@ def get_formatter(format_type, article):
 
     :param etree: parsed xml
     """
-    for formatter in formatters:
-        if formatter.can_format(format_type, article):
-            return formatter
+    for formatter_cls in formatters:
+        formatter_instance = formatter_cls()
+        if formatter_instance.can_format(format_type, article):
```
ride90: It would be possible to run `can_format`  without creating an `formatter_instance`, but since `can_format` is not static/class method and all core's + custom use it, it's better not to touch it.

petrjasek: yep would keep it bc compatible, but we could introduce new static method which could be implemented in formatter, and default implementation could be creating instance and calling that can_format method

ride90: @petrjasek we'll still create an instance to run `can_format`, so there is no actual benefit imo. It will make only `get_formatter` function just a little bit cleaner and at the same time that new static method can confuse somebody. I'm not sure :(

PR #1418, opened 2018-11-19
fix: /desks should return members sorted by display name

GyanP:
https://dev.sourcefabric.org/browse/SDESK-3566

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -24,6 +24,7 @@ from superdesk.notification import push_notification
 from superdesk.activity import add_activity, ACTIVITY_UPDATE
 from superdesk.metadata.item import FAMILY_ID
 from eve.utils import ParsedRequest
+from superdesk.utils import ListCursor
 
 
 class DeskTypes(SuperdeskBaseEnum):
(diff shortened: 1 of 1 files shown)


apps/desks.py
```diff
@@ -321,6 +322,35 @@ def get_desk_name(self, desk_id):
...
+
+        desk_list = []
+        members_list = []
+        db_users = app.data.mongo.pymongo('users').db['users']
+
+        # find display_name from the users document for each member in desks document
+        for item in desks_data:
+            desk_list.append(item)
+            for member in item['members']:
+                user_data = db_users.find_one({'_id': member['user']}, {'display_name': 1})
```
petrjasek: I think you could get them all at once using `get` with `{'_id': {'$in': [member['user'] for member in item['members']]}}` lookup

GyanP: Do you think it's a good idea? Considering it will return a pymongo cursor object and again I have to iterate over it to get the members_list with unique values from all the desks together.

apps/desks.py
```diff
@@ -321,6 +322,36 @@ def get_desk_name(self, desk_id):
...
+        # find display_name from the users document for each member in desks document
+        for desk in desks:
+            if 'members' in desk:
+                users = list(db_users.find({
+                    '_id': {'$in': [member['user'] for member in desk['members']]}},
+                    {'display_name': 1}
+                ))
+
+                for user in users:
+                    if not any(item == user for item in members_list):
```
petrjasek: I don't think you will get any duplicates from mongo, _id should be unique

GyanP: Actually, two desks can have the same members in that case the user data get duplicate entry in members_list, that's why I have that check for duplicate data.

petrjasek: I see, not sure why not handle it per desk then, is it to avoid extra sorting?

apps/desks.py
```diff
@@ -321,6 +322,36 @@ def get_desk_name(self, desk_id):
...
+
+        if members_list:
+            members_list.sort(key=lambda k: k['display_name'].lower())
+            ordered_dict = []
+            for member in members_list:
+                ordered_dict.append(member['_id'])
+
+            # sort the members of each desk according to ordered_dict
+            for desk in desks:
+                desk['members'].sort(key=lambda x: ordered_dict.index(x['user']))
```
petrjasek: here you should probably check again `if 'members' in desk:`

GyanP: I don't think so, as I already have this check

```suggestion
        if members_list:
```
and `members_list` will be non-empty only if there are `'members' in desk` which I have already checked

apps/desks.py
GyanP: yeah it's to avoid extra sorting.

apps/desks.py
petrjasek: but as long as there is single desk with members it will populate members_list, but there could be still other desks without any members.

GyanP: Ok, I see I have updated it.

PR #1419, opened 2018-11-20
[SDESK-3533](search): Use aggregation from query params or class instance variable

MarkLark86:


mdhaman:
@petrjasek do we want to allow aggregations from client or only controlled via server.

petrjasek:
@mdhaman I'm not sure, imo it has some dos attack potential if we let client do it, maybe if we define some whitelist on server what can be done, like eve has for mongo operations..

PR #1420, opened 2018-11-21
feat(content types): Don't return on content type the field for custo…

ioanpocol:
…m vocabulary with selection type 'do not show' (SDESK-3625)

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -16,6 +16,7 @@ from superdesk.resource import build_custom_hateoas
 
 
 CONTENT_TYPE_PRIVILEGE = 'content_type'
+DO_NOT_SHOW_SELECTION = 'do not show'
 
 # Fields that might not be in the schema but should be still available in formatter/output
 REQUIRED_FIELDS = (
@@ -194,8 +195,19 @@ class ContentTypesService(superdesk.Service):
 
 
 def clean_doc(doc):
-    clean_json(doc.get('schema', {}))
-    clean_json(doc.get('editor', {}))
+    schema = doc.get('schema', {})
+    editor = doc.get('editor', {})
+    vocabularies = get_forbiden_vocabularies()
+
+    for vocabulary in vocabularies:
+        field = vocabulary.get('schema_field', vocabulary['_id'])
+        if schema.get(field):
+            del schema[field]
+        if editor.get(field):
+            del editor[field]
+
+    clean_json(schema)
+    clean_json(editor)
 
 
 def clean_json(json):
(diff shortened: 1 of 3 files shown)


apps/content_types/content_types.py
```diff
@@ -253,12 +265,21 @@ def get_fields_map_and_names():
...
         fields_map[vocabulary.get('schema_field', vocabulary['_id'])] = vocabulary['_id']
         field_names[vocabulary['_id']] = vocabulary.get('display_name', vocabulary['_id'])
 
     return fields_map, field_names
 
 
+def get_forbiden_vocabularies():
+    vocabularies = get_resource_service('vocabularies').get_custom_vocabularies()
+
+    return (vocabulary for vocabulary in vocabularies
```
petrjasek: imo we can filter it via db, that might be quicker than fetching all vocabularies with items

tests/content_types_test.py
```diff
@@ -17,6 +17,21 @@ def find_one(self, req, **lookup):
...
 
 
+class MockVocabulariesService():
+
+    def get_custom_vocabularies(self):
+        return [
+            {
+                "_id": "test",
+                "field_type": None,
+                "selection_type": "do not show",
```
petrjasek: seems like this is bool value, would be probably more efficient to store it as such.
also the name is a bit misleading imo, it seems like there is some selection type like radio/checkboxes/select for the field, while this controls if field is displayed

ioanpocol: There are 3 values for selection_type: 'single selection', 'multi selection' and 'do not show'

petrjasek: so what does it mean actually?

ioanpocol: https://dev.sourcefabric.org/browse/SDESK-3551

if I select a "single value" type I expect the CV to be shown as a dropdown in the UI
if I select a "multiple value", I expect the CV to be shown as a multi-select UI component
Should disable the option to add to CPs the CVs that had set 'do not show' value

apps/content_types/content_types.py
ioanpocol: Done

tests/content_types_test.py
petrjasek: ok I see, if there should be single control in the cv editor for all these it makes sense. can you pls add the field to schema and list those as allowed values with some comment what it does?

ioanpocol: It is already added to vocabularies schema:
https://github.com/superdesk/superdesk-core/blob/master/superdesk/vocabularies/vocabularies.py#L81

PR #1424, opened 2018-11-23
feat(ninjs): Add support for related content on ninjs format (SDESK-3…

ioanpocol:
…631)

### DIFF
--- a/tests/publish/ninjs_formatter_test.py
+++ b/tests/publish/ninjs_formatter_test.py
@@ -946,3 +946,152 @@ class NinjsFormatterTest(TestCase):
         seq, doc = self.formatter.format(article, {"name": "Test Subscriber"})[0]
         ninjs = json.loads(doc)
         self.assertEqual(ninjs, expected)
+
+    def test_custor_related_items(self):
+        self.app.data.insert(
+            "content_types",
+            [
+                {
+                    "_id": ObjectId("5ba11fec0d6f1301ac3cbd13"),
+                    "label": "custom related content",
+                    "editor": {
+                        "slugline": {"order": 2, "sdWidth": "full"},
+                        "headline": {"order": 3, "formatOptions": ["underline", "link", "bold"]},
+                        "custom_related_content": {"order": 1},
+                    },
+                    "schema": {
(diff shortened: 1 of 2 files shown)


tests/publish/ninjs_formatter_test.py
```diff
@@ -946,3 +946,152 @@ def test_custom_media(self):
...
+                            "mimetype": "image/jpeg",
+                            "width": 550,
+                        }
+                    },
+                    "type": "picture",
+                    "version": "1",
+                },
+            },
+            "extra_items": {
+                "custom_related_content": {
```
takeit: how would I know that this is related content? Is this hardcoded value from Superdesk or user can define his own?

I ask because in `extra_items` there can be already custom media items (galleries/slideshows):

```json
{
      "extra_items":{
        "slideshow1": {
          "items":[
              ....
          ]
        }
      }
}
```

so there won't be a way to find out what is related item and what is custom media item.

takeit:
1. we don't want to hardcode anything because it will mean that we have to hardcode some values in SD config/metadata just to make this feature work with Publisher, it has to be flexible.
2. if we will have an info in `extra_items` about items which says: "Hey, this is `related_content` item" then that would work fine. The thing is there is already a `type` property in ninjs item. Maybe adding a new object `related_items` and putting related items under it (instead of extra_items) would be a better solution?

```
{
      "related_items":{
        "custom_related_content_1": {
          "items":[
              ....
          ]
        }
      }
}
```

What do you think? @ioanpocol @petrjasek

ioanpocol:
@karelpetrak Any feedback on this proposal?

petrjasek:
imo we can add `type` field to `extra_items`:
```
{
      "extra_items":{
        "slideshow1": {
          "type": "gallery", // or "related"
          "items":[
              ....
          ]
        }
      }
}
```

takeit:
that would also work 👍

ioanpocol:
👍

ioanpocol:
The PR is here: https://github.com/superdesk/superdesk-core/pull/1425

karelpetrak:
👍

PR #1427, opened 2018-11-27
fix(validation): Fix validation for custom related content fields(SDE…

ioanpocol:
…SK-3464)

### DIFF
--- a/apps/validate/validate.py
+++ b/apps/validate/validate.py
@@ -246,21 +246,23 @@ class ValidateService(superdesk.Service):
                         # fails for json fields like subject, genre
                         pass
 
-    def _get_media_field(self, field_schema, field_associations, doc,):
-        """Returns the field name in associations. For multiple valued media fields
+    def _get_media_fields(self, field_schema, field_associations, doc,):
+        """Returns the field names in associations. For multiple valued media fields
         the field asssociations name is different from the field schema name.
         :param field_schema: Field schema name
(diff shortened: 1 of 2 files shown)


apps/validate/validate.py
```diff
@@ -270,13 +272,16 @@ def _process_media(self, doc, validation_schema):
...
-                doc[field_schema] = doc['associations'][media_field]
-                if not doc.get('feature_media', None) is None and 'description_text' in doc['feature_media']:
-                    doc['media_description'] = doc['associations']['featuremedia']['description_text']
+            media_fields = self._get_media_fields(field_schema, field_associations, doc)
+            for media_field in media_fields:
+                if media_field and isinstance(doc['associations'][media_field], MutableMapping):
+                    doc[field_schema] = doc['associations'][media_field]
+                    if not doc.get('feature_media', None) is None and 'description_text' in doc['feature_media']:
+                        doc['media_description'] = doc['associations']['featuremedia']['description_text']
+                    del doc['associations'][media_field]
```
petrjasek: why is it removing the field?

petrjasek: to skip validation on it I guess..

ioanpocol: The field that is validated is copied on doc root at line 281 and this will be validated.
There is a validation for associations that treat all items as media and requests description_text, alt_text and other media related fields.

PR #1431, opened 2018-11-29
upgrade script for translations works for published items

tomaskikutis:


superdesk/data_updates/00015_20181127-105425_archive.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # processing trees
+        for root_node in get_root_nodes(tree_items):
+
+            for resource in ['archive', 'published']:
+                collection = mongodb_database[resource]
+                ids = get_ids_recursive([root_node], resource)
+
+                print(
+                    collection.update_many(
```
petrjasek: you should use probably [service.system_update](https://github.com/superdesk/superdesk-core/blob/master/superdesk/services.py#L69) in order to make the changes in both mongo and elastic

tomaskikutis: Do I update it one by one or is there a way to do something like `$in` condition in mongo?

petrjasek: one by one

superdesk/data_updates/00015_20181127-105425_archive.py
```diff
@@ -0,0 +1,95 @@
...
+        # processing trees
+        for root_node in get_root_nodes(tree_items):
+            updates = {'translation_id': root_node.id}
+
+            for resource in ['archive', 'published']:
+                service = get_resource_service(resource)
+                ids = get_ids_recursive([root_node], resource)
+
+                for item_id in ids:
+                    item = service.find_one(req=None, _id=item_id)
```
petrjasek: I think it should be using `item_id` for published

tomaskikutis: Why?

PR #1438, opened 2018-12-04
SDESK-3640 Add links to the items in superdesk instance in saved search reports.

ride90:


superdesk/templates/saved_searches_report.html
```diff
@@ -10,7 +10,10 @@ <h1>Saved Searches Report</h1>
     <ul>
         {% for doc in docs %}
         <li>
-        <strong>{{doc.headline|e}}</strong> (<em>{{doc.guid|e}}</em>)
+            <strong>{{doc.headline|e}}</strong>
+            (<a target="_blank" href="{{client_url}}/#/workspace/personal?item={{doc.guid|e}}&action=edit">
```
jerome-poisson: Are we sure that `client_url` doesn't end with a `/` ? Else a `rstrip` could help (or a path.join, or something similar)

ride90: `CLIENT_URL` exists in https://github.com/superdesk/superdesk-core/blob/master/superdesk/default_settings.py#L101. 
I will add `rstrip` just in case of broken admin.

PR #1445, opened 2018-12-07
feat(links): Add local_domains cv in order to detect local links on e…

ioanpocol:
…ditor3 (SDFID-458)

### DIFF
--- a/apps/prepopulate/data_sample/vocabularies.json
+++ b/apps/prepopulate/data_sample/vocabularies.json
@@ -3047,5 +3047,21 @@
             "qcode" : {}
         },
         "display_name" : "Countries"
-    }
+    },
+	{
+	    "_id" : "local_domains",
+	    "display_name" : "Local Domains",
+	    "type" : "manageable",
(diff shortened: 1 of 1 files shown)


apps/prepopulate/data_sample/vocabularies.json
```diff
@@ -3047,5 +3047,21 @@
...
         "display_name" : "Countries"
-    }
+    },
+	{
+	    "_id" : "local_domains",
+	    "display_name" : "Local Domains",
+	    "type" : "manageable",
+	    "unique_field" : "domain",
+	    "selection_type" : "do not show",
+	    "items" : [
```
petrjasek: shouldn't this be rather empty by default?

ioanpocol: This is prepopulation script that is used for tests so I think it is ok to have something there.

petrjasek: I see, missed that it's data sample

PR #1446, opened 2018-12-07
Display latest properties for related items

GyanP:
https://dev.sourcefabric.org/browse/SDESK-3681

### DIFF
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -941,7 +941,10 @@ class ArchiveService(BaseService):
             for item_name, item_obj in item.get(ASSOCIATIONS).items():
                 if item_obj and self._is_related_content(item_name):
                     item_id = item_obj[config.ID_FIELD]
-                    item[ASSOCIATIONS][item_name] = {'_id': item_id}
+                    if 'state' in item_obj and item_obj['state'] == 'published':
+                        item[ASSOCIATIONS][item_name] = {'_id': item_obj['item_id']}
+                    else:
+                        item[ASSOCIATIONS][item_name] = {'_id': item_id}
         return item
 
     def _is_related_content(self, item_name):
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -384,15 +384,6 @@ class ArchiveService(BaseService):
     def find_one(self, req, **lookup):
         item = super().find_one(req, **lookup)
 
-        # To fetch the related items json on the basis of ID stored in association for related items
-        if item and 'associations' in item:
-            for item_name, item_obj in item.get(ASSOCIATIONS).items():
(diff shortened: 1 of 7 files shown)


apps/archive/archive.py
```diff
@@ -303,7 +303,10 @@ def _handle_media_updates(self, updates, original, user):
             self._set_association_timestamps(item_obj, updates, new=False)
 
             stored_item.update(item_obj)
-            updates[ASSOCIATIONS][item_name] = stored_item
+            if 'test' in item_name:
```
ioanpocol: On settings you can define multiple related items fields, I think 'test' is just the name of one of defined related items on your instance.
You have to get the list of related items fields and check them against the fields on associations.
The key on associations has field_name--number format

ioanpocol:
Take a look at on_update and _handle_media_updates functions.

apps/archive/archive.py
```diff
@@ -376,18 +376,18 @@ def replace(self, id, document, original):
...
-            raise SuperdeskApiError.forbiddenError("User does not have permissions to read the item.")
-
         # To fetch the related items json on the basis of ID stored in association for related items
         if item and 'associations' in item:
             for item_name, item_obj in item.get(ASSOCIATIONS).items():
-                if item_obj and 'test' in item_name:
+                if item_obj and self._is_related_content(item_name):
                     item_id = item_obj[config.ID_FIELD]
-                    stored_item = self.find_one(req=None, _id=item_id)
+                    stored_item = super().find_one(req=None, _id=item_id)
```
ioanpocol: If the item is published a new item is created and this item should be retrieved here.
Also on every update or correction on a published item a new item is created and there is a flag last_published_version that helps you to find the last one.

apps/archive/archive.py
```diff
@@ -887,6 +894,15 @@ def _add_desk_metadata(self, updates, original):
         """
         return get_resource_service('desks').apply_desk_metadata(updates, original)
 
+    def _is_related_content(self, item_name):
```
ioanpocol: Can you retrieve the list only one time?

apps/archive/archive.py
ioanpocol: Sorry, my mistake, I think on update you don't have to retrieve the new created item, you still have to find the one with last_published_version set to true.

GyanP:
@ioanpocol, I have made a small change to fix, when we try to open a published related item by clicking on edit icon it does not open.

petrjasek:
@GyanP can you pls rebase it? it has some extra commits

GyanP:
Rebased the PR

ioanpocol:
@GyanP Please fix the failing tests

ioanpocol:
@GyanP 
Is here anything remaining to be done?

PR #1448, opened 2018-12-11
SDESK-3686 Flush elastic search with manage.py command.

ride90:


superdesk/commands/flush_elastic_index.py
```diff
@@ -0,0 +1,63 @@
...
+    def run(self, sd_index, capi_index):
+        if not (sd_index or capi_index):
+            raise SystemExit('You must specify at least one elastic index to flush. '
+                             'Options: `--sd`, `--capi`')
+        if sd_index:
+            self._flush_elastic(superdesk.app.config['ELASTICSEARCH_INDEX'])
+        if capi_index:
+            self._flush_elastic(superdesk.app.config['CONTENTAPI_ELASTICSEARCH_INDEX'])
+
+        print('- Indexing all mongo collections into elastic index(s)')
```
petrjasek: imo this part should be optional, so it would be possible to remove all content with this, otherwise it's like index from mongo just removing items first

superdesk/commands/flush_elastic_index.py
```diff
@@ -0,0 +1,63 @@
...
+            index
+        )
+        print('- Removing elastic index "{}"'.format(index))
+        resp = requests.delete(es_index_url)
+        if resp.status_code == requests.status_codes.codes.OK:
+            print('\t- "{}" elastic index was deleted'.format(index))
+        else:
+            print(
+                '\t- "{}" elastic index was not deleted. Server response: {}'.format(index, resp.text)
+            )
```
jerome-poisson: the command should exit with an error code here, or raise an exception.

ride90: @jerome-poisson this part will exit with `SystemExit` only in case if delete elastic response status not 200 or 404.

superdesk/commands/flush_elastic_index.py
```diff
@@ -0,0 +1,104 @@
...
+
+        :param str index: elastix index
+        :raise: SystemExit exception if delete elastic index response status is not 200 or 404.
+        """
+        es_index_url = urljoin(
+            superdesk.app.config['ELASTICSEARCH_URL'],
+            index
+        )
+        print('- Removing elastic index "{}"'.format(index))
+        resp = requests.delete(es_index_url)
```
petrjasek: you can get [elasticsearch instance](https://github.com/superdesk/superdesk-core/blob/master/superdesk/commands/rebuild_elastic_index.py#L41) and then you can use [indices api](https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.client.IndicesClient.delete) so you wouldn't have to deal with different errors here, it would raise an exception like index not found

ride90: @petrjasek thanks, I will use it.

ride90: @petrjasek , I used elasticsearch instance to operate with search. In this case it is kinda the same, because I want to print log when index was not found, so I still need to deal with exceptions :) But it is more solid to use a wrapper.

PR #1455, opened 2018-12-14
fix(correction): When a translated item is corrected, send superdesk …

ioanpocol:
…and email notifications to translate creator and publisher (SDFID-514)

### DIFF
--- a/superdesk/templates/translation_changed.html
+++ b/superdesk/templates/translation_changed.html
@@ -1,0 +1,5 @@
+{% extends "email_layout.html" %}
+
+{% block content %}
+<p>{{username}} changed the following translated article: <a href="{{link}}">{{link}}</a></p>
+{% endblock %}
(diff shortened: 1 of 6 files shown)


superdesk/templates/translation_changed_subject.txt
```diff
@@ -0,0 +1 @@
+The translated item {{title}} has been changed
```
petrjasek: imo we shouldn't use `title` here, either `headline` or just pass the item, to be consistent with other templates and UI

ioanpocol: I replaced title to headline. At list in this file there is no case to send the item to the template.

superdesk/templates/translation_changed.html
```diff
@@ -0,0 +1,5 @@
+{% extends "email_layout.html" %}
+
+{% block content %}
+<p>{{username}} changed the following translated article: <a href="{{link}}">{{link}}</a></p>
```
petrjasek: imo here we could use item headline for the link text

ioanpocol: This was my first intention but then I seen that on user_mentions is used in this way.
I think that the scope was to be able to copy the link.

PR #1456, opened 2018-12-14
[CONTENT TYPES] fixed subject made mandatory in schema when not prese…

jerome-poisson:
…nt in editor

If some custom vocabularies were present in `content_types` and `subject` was
not set in the editor, it was made mandatory anyway in the `schema`,
resulting in validation errors.

This patch fixes this

SDESK-3745

### DIFF
--- a/tests/content_types_test.py
+++ b/tests/content_types_test.py
@@ -87,3 +87,37 @@ class ContentTypesTestCase(TestCase):
 
         content_types.clean_doc(profile)
         self.assertEqual({'schema': {}, 'editor': {}}, profile)
+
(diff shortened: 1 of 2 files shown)


tests/content_types_test.py
```diff
@@ -4,7 +4,10 @@
 
 from superdesk.tests import TestCase
 from unittest import mock
-from apps.content_types import content_types, apply_schema
+try:
+    from apps.content_types import content_types, apply_schema
+except ImportError:
```
ride90: This is something which is makes sense to figure out why it happens..

jerome-poisson: forgot to remove it actually

PR #1457, opened 2018-12-18
[SDESK-3730] Embed featured image into email output

marwoodandrew:


superdesk/publish/formatters/email_formatter.py
```diff
@@ -63,6 +64,7 @@ def format(self, article, subscriber, codes=None):
                 doc['message_html'] = None
             doc['message_text'] = render_template('email_article_body.txt', article=formatted_article)
             doc['message_subject'] = render_template('email_article_subject.txt', article=formatted_article)
+            doc['renditions'] = formatted_article.get('associations', {}).get('featuremedia', {}).get('renditions')
```
MarkLark86: Sometimes `associations.featuremedia == None`, in this case an exception will be thrown here `AttributeError: 'NoneType' object has no attribute 'get'` when attempting to get the renditions.

superdesk/publish/transmitters/email.py
```diff
@@ -49,16 +56,70 @@ def _transmit(self, queue_item, subscriber):
...
                        text_body=text_body,
                        html_body=html_body,
-                       bcc=bcc)
+                       bcc=bcc,
+                       attachments=attachments)
 
         except Exception as ex:
             raise PublishEmailError.emailError(ex, queue_item.get('destination'))
 
+    def _get_watermark(self, image):
```
mdhaman: imo, `get_watermark` does not belong here and can be moved to `superdesk.media.media_operations` module.

superdesk/publish/formatters/email_formatter.py
marwoodandrew: OK will do!

superdesk/publish/transmitters/email.py
marwoodandrew: OK will do!

PR #1459, opened 2018-12-19
fix(notifications): Change the text on original changed notification(…

ioanpocol:
…SDFID-514)

### DIFF
--- a/superdesk/templates/translation_changed.html
+++ b/superdesk/templates/translation_changed.html
@@ -1,5 +1,6 @@
 {% extends "email_layout.html" %}
 
 {% block content %}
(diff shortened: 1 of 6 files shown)


superdesk/templates/translation_changed.html
```diff
@@ -1,5 +1,6 @@
 {% extends "email_layout.html" %}
 
 {% block content %}
-<p>{{username}} changed the following translated article: <a href="{{link}}">{{link}}</a></p>
+<p>The following original item has been modified by {{username}}.
```
petrjasek: I think it can use
```
...</p>
<p>...
```
so it will start on new line. line break doesn't do much in html

PR #1476, opened 2019-01-15
fixed: NTB "Subject required" error shown on publishing even though subject is not empty

GyanP:
https://dev.sourcefabric.org/browse/SDNTB-547

### DIFF
--- a/apps/validate/validate.py
+++ b/apps/validate/validate.py
@@ -55,6 +55,8 @@ def get_validator_schema(schema):
     validator_schema = {key: val for key, val in schema.items() if val is not None}
 
     if validator_schema.get('required') and not validator_schema.get('minlength'):
+        if 'mandatory_in_list' in validator_schema and 'subject' in validator_schema['mandatory_in_list']['scheme']:
+            validator_schema['schema']['schema']['scheme']['allowed'] = ['subject_custom', 'category']
         validator_schema.setdefault('empty', False)
 
     return validator_schema
@@ -97,9 +99,10 @@ class SchemaValidator(Validator):
 
     def _validate_mandatory_in_dictionary(self, mandatory, field, value):
         """Validates if all elements from mandatory are presented in the dictionary"""
-        for key in mandatory:
-            if not value.get(key):
-                self._error(key, REQUIRED_FIELD)
+        if field != 'subject':
+            for key in mandatory:
+                if not value.get(key):
+                    self._error(key, REQUIRED_FIELD)
 
     def _validate_empty(self, empty, field, value):
         """Original validates only strings, adding a list check."""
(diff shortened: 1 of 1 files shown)


apps/validate/validate.py
```diff
@@ -55,6 +55,8 @@ def get_validator_schema(schema):
     validator_schema = {key: val for key, val in schema.items() if val is not None}
 
     if validator_schema.get('required') and not validator_schema.get('minlength'):
+        if 'mandatory_in_list' in validator_schema and 'subject' in validator_schema['mandatory_in_list']['scheme']:
+            validator_schema['schema']['schema']['scheme']['allowed'] = ['subject_custom', 'category']
```
ioanpocol: subject_custom and category are the names of a custom fields, so they cannot be used as constants

apps/validate/validate.py
```diff
@@ -90,10 +92,11 @@ def _validate_type_text(self, field, value):
 
     def _validate_mandatory_in_list(self, mandatory, field, value):
         """Validates if all elements from mandatory are presented in the list"""
-        for key in mandatory:
-            for key_field in mandatory[key]:
-                if not check_json(value, key, mandatory[key][key_field]):
-                    self._error(key_field, REQUIRED_FIELD)
+        if field != 'subject': # ignore in case of subject and category
```
ioanpocol: I think subject can be validated too.

GyanP: This is how mandatory_in_list looks for subject schema `{'scheme': {'category': 'category', 'subject': 'subject_custom'}}`. So, if either category or subject is not in the template then it gives the error. Both must be present together.

apps/validate/validate.py
GyanP: I am unable to edit the content profiles. So, I think t depends on this https://dev.sourcefabric.org/browse/SDNTB-550.

apps/validate/validate.py
ioanpocol: I think this code is covered by the tests in ValidateMandatoryInListTest
Looks like field is ignored because only subject can have mandatory_in_list so I think field != 'subject' is always false.
What value did you receive here?

GyanP: yes, `field != 'subject'` is always false.

ioanpocol: What value did you receive here?

GyanP: you mean for this `def _validate_mandatory_in_list(self, mandatory, field, value):` ?
mandatory = `{'scheme': {'category': 'category', 'subject': 'subject_custom'}}`
field = `subject` and
value = [{'name': 'Arkeologi', 'qcode': '01001000', 'parent': '01000000', 'scheme': 'subject_custom'}]

ioanpocol: OK, so category field is missing and you will get a validation error for 'category'.

GyanP: yes, but shouldn't it be like if I select a value for `subject` then `category` field must be visible in the editor? Because I can see only `ANPA category` but not `category`.

GyanP: yes, exactly

PR #1477, opened 2019-01-15
implement async transmitting

petrjasek:
so multiple items can be pushed to single subscriber at the same time.
it must be enabled via subscriber config


petrjasek:
also changed the way how we pick queue items to publish, it used to get first 500 but if you would have subscriber with 500+ items in queue it would block other subscribers. now it gets first list of subscribers that should get content and then it gets items for each

PR #1487, opened 2019-01-21
(SDESK-3824)- fixed: After cropping image in edit image, other operat…

GyanP:
…ions bring back the original

### DIFF
--- a/apps/picture_renditions.py
+++ b/apps/picture_renditions.py
@@ -3,6 +3,7 @@ import superdesk
 from flask import current_app as app
 from superdesk.media.renditions import generate_renditions, get_renditions_spec
 from apps.picture_crop import get_file
+from superdesk import get_resource_service
 
 
 class PictureRenditionsService(superdesk.Service):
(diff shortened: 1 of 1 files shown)


apps/picture_renditions.py
```diff
@@ -25,8 +29,15 @@ def create(self, docs, **kwargs):
...
             renditions = generate_renditions(orig_file, orig['media'], inserted, media_type,
                                              mimetype, rendition_spec, app.media.url_for_media)
+
             doc['renditions'] = renditions
             ids.append(item['_id'])
+
+            updates = {'renditions': item['renditions']}
+            try:
+                get_resource_service('archive').update(item['_id'], updates, item)
+            except Exception as ex:
```
petrjasek: any more specific exception we could use here?

GyanP: No, I think.

ioanpocol: @petrjasek Any feedback here?

petrjasek: no

PR #1496, opened 2019-01-22
Ingest afp newsml

idsvn:
SDBELGA-14 Ingest of AFP content
Create a new parser file for Belga(belga_newsml_1_0.py) to avoid effect to process of origin process.
-Support parser multi newsitems in xml file. Because Belga xml have many newsitem in a xml file.
-Check and set default value to avoid the error, when element is missing.
-Create and set value for the properties, is missing for Belga.
fix format datetime '%Y%m%dT%H%M%S+0000'(only get utc if not raise error) to '%Y%m%dT%H%M%S%z'(newsml_1_2.py)
fix convert localize to utc timezone (afp_newsml_1_2.py)


superdesk/io/feed_parsers/belga/belga_newsml_1_2.py
```diff
@@ -0,0 +1,452 @@
...
+            l_item = []
+            self.root = xml
+
+            # parser the NewsEnvelope element
+            item_envelop = self.parser_newsenvelop(xml.find('NewsEnvelope'))
+
+            # parser the NewsItem element
+            l_newsitem_el = xml.findall('NewsItem')
+            for newsitem_el in l_newsitem_el:
+                item = item_envelop
```
petrjasek: I think it should use `item_envelop.copy()` or something, otherwise it would be updating same dict all the time

idsvn: Yeah, that right. Thanks.

ride90:
Hi @idsvn , I saw your pr to belga repo, but now it's disappeared. Did you close it?

ride90:
Closing this one, duplicate of https://github.com/superdesk/superdesk-belga/pull/15

PR #1508, opened 2019-01-30
SDESK-3880 Timer to measure program execution time and logs for FTP ingest.

ride90:
FTP ingest logs example:
screenshot 2019-01-31 at 14 19 29

### DIFF
--- a/superdesk/io/feeding_services/ftp.py
+++ b/superdesk/io/feeding_services/ftp.py
@@ -184,6 +184,7 @@ class FTPFeedingService(FeedingService):
         return ext.lower() in allowed_ext
 
     def _list_files(self, ftp, provider):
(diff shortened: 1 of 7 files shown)


superdesk/io/feeding_services/ftp.py
```diff
@@ -277,7 +309,15 @@ def _update(self, provider, update):
...
 
+                self._log_msg(
+                    "Got {} file for processing. Exec time: {:.4f} secs.".format(
+                        len(files_to_process), self._timer.stop('files_to_process')
+                    )
+                )
+
                 # process files
+                self._timer.start('start_processing')
+                self._timer.start('retrieve_parse')
```
petrjasek: imo this should go into `_retrieve_and_parse` method, it's only used there

PR #1511, opened 2019-01-31
fix(upload): Set 'in-progress' status for uploaded images (SDESK-3898)

ioanpocol:


apps/archive/archive_media.py
```diff
@@ -45,6 +45,7 @@ def on_create(self, docs):
 
             try:
                 doc[ITEM_TYPE] = self.type_av.get(file_type)
+                doc[ITEM_STATE] = 'in_progress'
```
petrjasek: btw I think we must have some constant for this

PR #1519, opened 2019-02-06
feat(signals): add item_fetched and item_moved signals

petrjasek:
SDANSA-220


ride90:
@petrjasek are handlers for this signals in the ansa codebase?

petrjasek:
@ride90 yes it will be used there

PR #1520, opened 2019-02-07
(SDESK-3910) fixed: Errors when saving an image after cropping and no…

GyanP:
… further actions available

### DIFF
--- a/apps/picture_renditions.py
+++ b/apps/picture_renditions.py
@@ -33,9 +33,12 @@ class PictureRenditionsService(superdesk.Service):
             doc['renditions'] = renditions
             ids.append(item['_id'])
 
+            is_article_edit_media = doc.get('is_article_edit_media', False)
+
             updates = {'renditions': item['renditions']}
             try:
-                get_resource_service('archive').update(item['_id'], updates, item)
+                if not is_article_edit_media:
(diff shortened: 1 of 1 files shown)


apps/picture_renditions.py
```diff
@@ -33,9 +33,12 @@ def create(self, docs, **kwargs):
             doc['renditions'] = renditions
             ids.append(item['_id'])
 
+            is_article_edit_media = doc.get('is_article_edit_media', False)
+
             updates = {'renditions': item['renditions']}
             try:
-                get_resource_service('archive').update(item['_id'], updates, item)
+                if not is_article_edit_media:
+                    get_resource_service('archive').update(item['_id'], updates, item)
```
mdhaman: This is going to change the `_etag`. Is the client aware about the change in `_etag`?

GyanP: This is to avoid the change in `_etag`. If the media is in the body of the article then only this update method gets called otherwise not.

mdhaman: @GyanP instead having another parameter in the payload and update the resource over here can you issue another request from the client to update resource `item` ?

GyanP: @mdhaman moved the resource update request to the client.

PR #1523, opened 2019-02-07
[INGEST] fix ingestion crash on unknown subject qcode:

jerome-poisson:
ingestion was failing (after parsing) when an unknown qcode was found.
This patch fixes it by ignoring unknown qcodes

SDESK-3804

### DIFF
--- a/superdesk/io/commands/update_ingest.py
+++ b/superdesk/io/commands/update_ingest.py
@@ -356,11 +356,19 @@ def process_iptc_codes(item, provider):
             if 'qcode' in subject and len(subject['qcode']) == 8 and subject['qcode'].isdigit():
                 top_qcode = subject['qcode'][:2] + '000000'
                 if not iptc_already_exists(top_qcode):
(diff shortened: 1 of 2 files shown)


superdesk/io/commands/update_ingest.py
```diff
@@ -356,11 +356,19 @@ def iptc_already_exists(code):
             if 'qcode' in subject and len(subject['qcode']) == 8 and subject['qcode'].isdigit():
                 top_qcode = subject['qcode'][:2] + '000000'
                 if not iptc_already_exists(top_qcode):
-                    item['subject'].append({'qcode': top_qcode, 'name': subject_codes[top_qcode]})
+                    try:
+                        item['subject'].append({'qcode': top_qcode, 'name': subject_codes[top_qcode]})
+                    except KeyError:
+                        logger.warning("missing qcode un subject_codes: {qcode}".format(qcode=top_qcode))
```
petrjasek: `un` looks :fr:

jerome-poisson: @petrjasek it's fixed ;)

PR #1525, opened 2019-02-11
Ignore etag during update when IF_MATCH is switched off.

ride90:


mdhaman:
@ride90 , @petrjasek what is problem here?

ride90:
@mdhaman when you have `IF_MATCH` switched off (for development), PATCH request raises an exception here `updates[config.ETAG] = updated[config.ETAG]`

PR #1529, opened 2019-02-13
set subject scheme

takeit:


superdesk/io/feed_parsers/ninjs.py
```diff
@@ -126,7 +126,12 @@ def _transform_from_ninjs(self, ninjs):
         return item
 
     def _format_qcodes(self, items):
-        return [{'name': item.get('name'), 'qcode': item.get('code')} for item in items]
+        subject = {'name': item.get('name'), 'qcode': item.get('code')}
+        
+        if item.get('scheme'):
+            subject['scheme'] = item.get('scheme')
+        
+        return [subject for item in items]
```
petrjasek: it should be more like:
```
subjects = []
for item in items:
    subject = {'name': item.get('name'), 'qcode': item.get('code')}
    if item.get('scheme'):
        subject['scheme'] = item.get('scheme')
    subjects.append(subject)
return subjects
```

takeit: ah right, fixed

PR #1533, opened 2019-02-14
[FILTER CONDITION] fixed exceptions on some missing vocabularies

jerome-poisson:
when some vocabularies were missing (namely "urgency", "priority",
"type" or "place"/"location"), the filter_conditions/parameters service was not working
correctly, resulting in empty `Field` dropdown list in client.
This patch fixes this by using empty list when those vocabularies don't
exist.

fixes SDBELGA-61

### DIFF
--- a/apps/content_filters/filter_condition/filter_condition_parameters.py
+++ b/apps/content_filters/filter_condition/filter_condition_parameters.py
@@ -135,9 +135,11 @@ class FilterConditionParametersService(BaseService):
         genre = vocabularies_resource.get(req=req, lookup=None)
         if genre.count():
             values['genre'] = genre[0]['items']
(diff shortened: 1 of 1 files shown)


jerome-poisson:
**note:** this patch fixes the issue with missing vocabularies, but I suspect that those missing vocabularies may cause troubles in others places. For instance `priority` may be checked by some formatters (but in my opinion they should use default values when not available).

In the case of Belga, the issue was caused by missing `priority`, `type` and `location` vocabularies.
I wonder if we should create those missing vocabularies all the time, even if we use empty items list. I guess not (superdesk should be able to deal with missing vocabularies, whatever it is).

@karelpetrak @petrjasek any thought on this?

PR #1539, opened 2019-02-21
SDESK-4001 Collation support. Add case insensitive indexes for concept items.

ride90:


ride90:
Depends on https://github.com/superdesk/fireq/pull/38 and https://dev.sourcefabric.org/browse/IN-3526

PR #1540, opened 2019-02-25
v1.28.1

pablopunk:


apps/picture_renditions.py
```diff
@@ -33,9 +33,12 @@ def create(self, docs, **kwargs):
             doc['renditions'] = renditions
             ids.append(item['_id'])
 
+            is_article_edit_media = doc.get('is_article_edit_media', False)
```
mdhaman: @pablopunk this was latter on remove by @GyanP. Can you please check with  @GyanP

pablopunk: @mdhaman the PR you are mentioned doesn't remove that line

pablopunk: okay I think I know what you mean, I might have cherry-picked only one commit from that PR

PR #1542, opened 2019-03-04
fix(macro): Add replace_type for macros (SDANSA-222)

ioanpocol:


petrjasek:
I think it can use some docs

PR #1545, opened 2019-03-07
SDFID-535 Rename 'code' to 'group_id' in concept items schema

ride90:


petrjasek:
any reason for that?

ride90:
@petrjasek `code` is not obvious name, it does grouping, so it's better to name it `group_id`.

PR #1546, opened 2019-03-07
feat(localisation): Add support for localisation on backend (SDESK-3834)

ioanpocol:


petrjasek:
might be easier with https://pythonhosted.org/Flask-Babel/

ioanpocol:
Ok, seams that it is better integrated.
I will still use 'format' for parameters in order to keep the same text format as on frontend:
_('Invalid version {old_version}').format(old_version=old_version)

apps/content_filters/filter_condition/filter_condition_service.py
```diff
@@ -45,13 +46,13 @@ def _check_parameters(self, docs):
         for doc in docs:
             parameter = [p for p in parameters if p['field'] == doc['field']]
             if not parameter or len(parameter) == 0:
-                raise SuperdeskApiError.badRequestError(
-                    'Filter condition:{} has unidentified field: {}'
-                    .format(doc['name'], doc['field']))
+                raise SuperdeskApiError.badRequestError(_(
+                    'Filter condition:{condition} has unidentified field: {field}')
+                    .format(consition=doc['name'], field=doc['field']))
```
mdhaman: typo here `condition`

superdesk/factory/app.py
```diff
@@ -115,6 +117,13 @@ def get_app(config=None, media_storage=None, config_object=None, init_elastic=No
     app.jinja_loader = custom_loader
     app.mail = Mail(app)
     app.sentry = SuperdeskSentry(app)
+    babel = Babel(app, configure_jinja=False)
+    app.config['BABEL_TRANSLATION_DIRECTORIES'] = '../translations'
+
+    @babel.localeselector
+    def get_locale():
+        user = getattr(g, 'user', {})
+        return user.get('language', 'en')
```
petrjasek: we should probably add some config for default language

translations/messages.pot
```diff
@@ -0,0 +1,816 @@
+# Translations template for PROJECT.
```
petrjasek: we probably don't need this file in repo

PR #1551, opened 2019-03-12
fix(archive): keep remove associations as null

petrjasek:
otherwise it won't be removed from elastic and is visible in lists

SDBELGA-74


mdhaman:
@petrjasek can we push this PR to 1.28 as well?

PR #1558, opened 2019-03-18
SDESK-4001 Add case insensitive index for concept items.

ride90:
Add case insensitive index for concept items.


apps/concept_items/tests.py
```diff
@@ -0,0 +1,137 @@
...
+    def setUp(self):
+        with self.app.app_context():
+            self.app.data.insert('concept_items', self.concept_items)
+
+    def test_query_all_items(self):
+        service = get_resource_service('concept_items')
+
+        self.assertTrue(
+            len(list(service.get_from_mongo(req=None, lookup={}))),
+            len(self.concept_items)
```
jerome-poisson: the second argument of assertTrue is a debug message, is it what is expected here? Doesn't seem so.
Maybe it should be `assertEqual` here?

ride90: @jerome-poisson , right, it should be assertEqual

PR #1561, opened 2019-03-20
SDFID-538 Implement storing HTML in knowledge base

ride90:


apps/concept_items/resource.py
```diff
@@ -30,9 +30,14 @@ class ConceptItemsResource(Resource):
             required=False,
             embeddable=False
         ),
-        'definition': {
+        'definition_plain': {
```
petrjasek: I think it could be like `definition_text` so it's consistent with item, we have there `body_text` and `body_html`

ride90: Makes sense. Thanks.

PR #1563, opened 2019-03-22
feat(signal): add signal for update action

petrjasek:
it's called `item_rewrite` to match endpoint name

SDANSA-239


mdhaman:
Hi @petrjasek, can we have this commit in 1.28 branch as well?

petrjasek:
sure @mdhaman

PR #1576, opened 2019-04-09
ingest embargoed in newsml g2 and publish it in ninjs

petrjasek:
STTNHUB-56

### DIFF
--- a/superdesk/publish/formatters/ninjs_formatter.py
+++ b/superdesk/publish/formatters/ninjs_formatter.py
@@ -173,7 +173,13 @@ class NINJSFormatter(Formatter):
         if extra_items:
             ninjs.setdefault(EXTRA_ITEMS, {}).update(extra_items)
 
-        if article.get(EMBARGO):
+        if article.get('embargoed'):
+            try:
+                ninjs['embargoed'] = article['embargoed'].isoformat()
+            except AttributeError:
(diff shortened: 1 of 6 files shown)


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -173,7 +173,13 @@ def _transform_to_ninjs(self, article, subscriber, recursive=True):
         if extra_items:
             ninjs.setdefault(EXTRA_ITEMS, {}).update(extra_items)
 
-        if article.get(EMBARGO):
+        if article.get('embargoed'):
+            try:
+                ninjs['embargoed'] = article['embargoed'].isoformat()
+            except AttributeError:
```
jerome-poisson: In which case the `isoformat` attribute may be missing? If it's a datetime, should not be always present?

petrjasek: just in case when it gets ingested and it might not parse the format properly

jerome-poisson: @petrjasek in this case maybe something should notify the user, a `logger.warning` for instance? That would indicate that something is wrong with either the file format or the parser.

petrjasek: good point, will update

PR #1578, opened 2019-04-16
Fixed: Cannot deschedule scheduled articles. (SDESK-4129)

GyanP:
New item created from internal destinations, is always set to published state as we were not sending `schedule_settings` and `publish_schdule` fields in `updates`. So, irrespective of what is the state of the original item, a newly created item is always set to be published. But now the state of new item will be the same as the original item.

### DIFF
--- a/apps/publish/enqueue/__init__.py
+++ b/apps/publish/enqueue/__init__.py
@@ -11,6 +11,7 @@
 import cProfile
 import logging
 import superdesk
+import superdesk.signals as signals
 from flask import current_app as app
 
 from superdesk import get_resource_service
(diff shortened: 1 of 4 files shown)


MarkLark86:
@GyanP It would be good to have feature tests along with this.

mdhaman:
@GyanP user can `deschedule` a scheduled item. In this case, the user has to manually `deschedule` an item on the internal destination and the user might forget to `deschedule` the item on the internal destination. Also user can `deschedule` item and again `schedule` and this would publish item twice for the internal destinations (this is an existing bug).

imo, we should delay the creation of the item on the internal destination for `scheduled` state i.e. only create the item for the internal destination after the schedule time as lapsed.  We could change logic of enqueue such that if the `scheduled` item is published then apply internal destination. (see https://github.com/superdesk/superdesk-core/blob/master/apps/publish/enqueue/__init__.py#L111)

@petrjasek @karelpetrak ?

karelpetrak:
Hi @mdhaman, 
in case of NTB, the internal destination is used to send stories for translation to another desk. They actually need to receive the story before it is transmitted to subscribers and keep the scheduled time of the original also for the routed copy (to prevent accidental publishing before the original is scheduled). 
In case you need to delay the creation of scheduled items, I think we need to have it configurable. Ideally as a switch on the internal destination settings page.

GyanP:
Requires https://github.com/superdesk/superdesk-client-core/pull/2965

PR #1579, opened 2019-04-17
[STTNHUB-58] - Auto published ingested items should preserve id

akintolga:


apps/duplication/archive_fetch.py
```diff
@@ -98,7 +98,8 @@ def fetch(self, docs, id=None, **kwargs):
             dest_doc['versioncreated'] = archived
             send_to(doc=dest_doc, desk_id=desk_id, stage_id=stage_id)
             dest_doc[ITEM_STATE] = doc.get(ITEM_STATE, CONTENT_STATE.FETCHED)
-            dest_doc[INGEST_ID] = dest_doc[FAMILY_ID] = ingest_doc[config.ID_FIELD]
+            dest_doc[FAMILY_ID] = ingest_doc[config.ID_FIELD]
+            dest_doc[INGEST_ID] = self.__strip_version_from_guid(ingest_doc[GUID_FIELD], ingest_doc.get('version'))
```
mdhaman: Here relationship is getting change between `ingest` and `archive` collection.  Can we handle this via newsroom formatter for STT?

akintolga: I see, thanks for the example, I see how you search ingest in the formatter.

At the moment this issue is limited to Newsroom, I think it is an issue for publishing in general. 

It seemed to me instead of doing this for formatters separately doing in superdesk for once would be better. But I don't want to break any existing dependency. 

When I checked I didn't see where else we used `INGEST_ID` field, which I double check again. 

An alternative would be to have a new field for keeping the original id somewhere.

mdhaman: If the issue is related to generic auto publishing and also ingest is transient (3 days) then we can break this dependency.

akintolga: No problem, I'll double check with @petrjasek see what he thinks,I think either way would work

PR #1582, opened 2019-04-29
[SWP-1525] - Setting the default category per environment variable

akintolga:


superdesk/default_settings.py
```diff
@@ -588,6 +588,9 @@ def local_to_utc_hour(hour):
 DEFAULT_GENRE_VALUE_FOR_MANUAL_ARTICLES = env('DEFAULT_GENRE_VALUE_FOR_MANUAL_ARTICLES',
                                               [{'qcode': 'Article', 'name': 'Article (news)'}])
 
+#: Defines default value for category for ingested and auto published articles
+DEFAULT_CATEGORY_VALUE_FOR_AUTO_PUBLISHED_ARTICLES = env('DEFAULT_CATEGORY_VALUE_FOR_AUTO_PUBLISHED_ARTICLES', None)
```
petrjasek: I wouldn't try to read it from env given that it should be list of objects, also it would make sense to have that hardcoded value as default here

MarkLark86: What about using a comma separated list of qcodes. When assigning the value retrieve the list of actual CV objects (otherwise it may not get the correct name if the CV is changed).

apps/rules/routing_rules_tests.py
```diff
@@ -344,3 +347,50 @@ def test_schedule_routing_rules_at_end_time(self):
...
+        instance = RoutingRuleSchemeService()
+        category = [{'qcode': 'a', 'name': 'Australian General News'}]
+        result = instance._assign_default_values({'anpa_category': category}, None)
+        self.assertEqual(result['anpa_category'], [{'qcode': 'a', 'name': 'Australian General News'}])
+
+        with self.app.app_context():
+            result = instance._assign_default_values({}, None)
+            self.assertIsNone(result['anpa_category'])
+
+            # self.app.config['DEFAULT_CATEGORY_VALUE_FOR_AUTO_PUBLISHED_ARTICLES'] = \
```
petrjasek: I guess we don't need this

PR #1587, opened 2019-05-07
setup: updated celery minimal version to 4.3.0

jerome-poisson:
Celery starts to be compatible with Python 3.7 from 4.3.0, and with this
version minimal accepted redis version is 3.2.0, so this patch update
those modules.


setup.py
```diff
@@ -27,8 +27,8 @@
     'asyncio>=3.4,<3.5',
     'bcrypt>=3.1.1,<3.2',
     'blinker>=1.3,<1.5',
-    'celery[redis]>=4.1.1,<4.2',
-    'redis==2.10.6',
+    'celery[redis]>=4.3.0,<4.4',
+    'redis>=3.2.0,<3.3',
     'kombu>=4.2,<4.3',
```
petrjasek: I guess kombu also needs an update

jerome-poisson: right, will update the PR

petrjasek:
@jerome-poisson any idea what's wrong here?

ride90:
@jerome-poisson what is the status of this pr? I've started to use 3.7.4 and it wants a new celery :)

jerome-poisson:
@petrjasek @ride90 not sure what's wrong, I'll have a look at it.

ride90:
Tested manually with python 3.7.4 and it works 👍

PR #1589, opened 2019-05-10
[SDESK-4218] - Extra fields for STT

akintolga:


petrjasek:
I think this should go to superdesk-stt repo, it's not useful for others using stt ingest right?

akintolga:
https://github.com/superdesk/superdesk-stt/pull/10

PR #1591, opened 2019-05-13
fixed: fetched image item belonging to package can't be saved in media gallery. (SDESK-4229)

GyanP:


apps/archive/archive.py
```diff
@@ -780,7 +780,9 @@ def _validate_updates(self, original, updates, user):
             raise SuperdeskApiError.badRequestError(_('Cannot change the genre for broadcast content.'))
 
         if PUBLISH_SCHEDULE in updates or "schedule_settings" in updates:
-            if is_item_in_package(original) and not force_unlock:
+            if (updates.get(PUBLISH_SCHEDULE, None) or
+                    any(v is not None for v in updates.get('schedule_settings', {}).values())) and\
```
jerome-poisson: small style note: the backslash is not needed at the end thanks to the parenthesis and the boolean operator should be at the beginning of next line.

PR #1600, opened 2019-05-23
SDESK-4200 concept items: implicitly apply case insensitive collation for name and definition fields.

ride90:
- Apply case insensitive collation only if collation was not provided explicitly and sorting by `name`, `definition_html` or `definition_text` was used
`http://localhost:5000/api/concept_items?sort=name`
 equal to 
`http://localhost:5000/api/concept_items?sort=name&collation={"locale": "en", "strength":"1"}`
- Sort using `definition_text` field if `definition_html` was provided as sort argument.
 `http://localhost:5000/api/concept_items?sort=definition_html`
equal to
`http://localhost:5000/api/concept_items?sort=definition_text&collation={"locale": "en", "strength":"1"}`

### DIFF
--- a/apps/concept_items/service.py
+++ b/apps/concept_items/service.py
@@ -12,6 +12,7 @@ import json
 from uuid import uuid4
 
 import superdesk
+from werkzeug.datastructures import ImmutableMultiDict
 from eve.utils import ParsedRequest
 from superdesk.text_utils import get_text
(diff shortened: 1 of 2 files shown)


tomaskikutis:
I've tested the client against this PR and case insensitive sorting as well as filtering by `definition_html` work now :+1:

PR #1601, opened 2019-05-28
[SPELLCHECKERS] Spellcheckers integration + Grammelecte implemention

jerome-poisson:
This patch introduce a generic way to register spellcheckers or tools to check
grammar/style.

2 new endpoints are available:

- spellcheckers_list, to retrieve list of registered spellcheckers
- spellchecker, to check a text

The later expects a spellchecker name to be specified using `spellchecker` key,
and a text to be specified using `text` key.

The returned dict will contain an "errors" key, with a list of dict containing
the following keys:
- `startOffset`: offset of the offending text
- `suggestions` (list of dicts)
- `text`: offending text
- `type`: one of "grammar" or "spelling"

alternatively, `suggestions` boolean can be set, in which case only
suggestions are returned (in `suggestions` key).

A first module implement this API for `Grammalecte` grammar + spelling
checker for French.

SDBELGA-109

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,7 +25,7 @@ For technical documentation on the Superdesk client web application, have a look
 
     We've got you covered. Check out the `User Manual <https://www.superdesk.org/uploads/superdeskmastermanual.pdf>`_.
 
-First steps 
+First steps
 -----------
 
 New to Superdesk? This is the best place to start. Here you will get the big picture of
@@ -47,6 +47,7 @@ works and can be developed further.
     architecture
     publish
     ingest
+    spellchecking
     settings
     schema
     content-types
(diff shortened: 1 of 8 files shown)


jerome-poisson:
Please check commit message + docstrings for documentation. Here is an example of payload for corrections:
```json
{
    "spellchecker": "grammalecte",
    "text": "Il nous reste à vérifié votre maquette."
}
```

Here is an example payload for suggestions:
```json
{
    "spellchecker": "grammalecte",
    "suggestions": true,
    "text": "fote"
}
```

jerome-poisson:
@ride90 it's risky to change the endpoint without knowing if and who is using it. I'll let as it is for now, but this should be checked later.

PR #1605, opened 2019-06-03
log time for media upload

petrjasek:
SDANSA-269


ride90:
@petrjasek can we reuse this one https://github.com/superdesk/superdesk-core/blob/master/superdesk/utils.py#L246 ?

PR #1614, opened 2019-06-14
[SDBELGA-125] Upload video from Superdesk to video server. (need review)

idsvn:
-Setting a switch for enable video server. if disable use upload normal in Superdesk.
-Add video editor service for integrating with Superdesk.
-Upload video to video server and record data to Superdesk.


apps/archive/archive_media.py
```diff
@@ -37,33 +40,44 @@ def on_create(self, docs):
...
-                    delete_file_on_error(doc, file_id)
-                abort(500)
+            if file_type == 'video' and app.config.get("VIDEO_SERVER_ENABLE"):
+                if not self.videoEditor.check_video_server():
+                    raise SuperdeskApiError(message="Cannot connect to videoserver", status_code=500)
+                # upload media to video server
+                self.upload_file_to_video_server(doc, content_type)
+                # get thumbnails for timeline bar
+                self.videoEditor.get_timeline_thumbnails(doc.get('media'), 40)
+            else:
```
petrjasek: I think it should be structured a bit differently to avoid duplication, there are some common parts which are here and inside `upload_file_to_video_server`, imo it would be better to only do the upload there and create a method for that original upload handling with generating renditions, and then use the same code for both again.

superdesk/media/video_editor.py
```diff
@@ -0,0 +1,180 @@
...
+            return False
+
+    def _url(self, _id, resource=''):
+        url = urljoin(self.get_base_url(), _id)
+        if resource:
+            return urljoin(url + '/', resource)
+        return url
+
+    def _get_response(self, resp, *status):
+        json = json_util.loads(resp.text)
```
petrjasek: I guess this might fail for some network errors, maybe it should only do it for 400 errors or good responses? or use try/catch

petrjasek:
pls make the pr on develop branch

PR #1618, opened 2019-07-01
[SPELLCHECKERS] implemented Leuven University Dutch Spellchecker

jerome-poisson:
text checking + suggestions have been implemented.

SDBELGA-122

### DIFF
--- a/superdesk/text_checkers/spellcheckers/leuven_dutch.py
+++ b/superdesk/text_checkers/spellcheckers/leuven_dutch.py
@@ -1,0 +1,90 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013-2019 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+import os
+import logging
+import requests
+from superdesk.errors import SuperdeskApiError
+from superdesk.text_checkers.spellcheckers.base import SpellcheckerBase
+
+logger = logging.getLogger(__name__)
+OPT_API_KEY = "LEUVEN_DUTCH_API_KEY"
+API_URL = "https://dev.schrijfhulp.be/v1/{method}/"
+START_MARKER = "##__ERR_START__##"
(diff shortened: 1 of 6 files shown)


superdesk/text_checkers/spellcheckers/leuven_dutch.py
```diff
@@ -0,0 +1,87 @@
...
+        r = requests.post(check_url, data=data)
+        if r.status_code != 200:
+            raise SuperdeskApiError.internalError("Unexpected return code from {}".format(self.name))
+
+        data = r.json()
+
+        err_list = []
+        check_data = {'errors': err_list}
+        for mistake in data['spellingchecker']['output'].get('mistakes', []):
+            idx = text.find(mistake)
```
petrjasek: what about case when you have same mistake multiple times in text?

jerome-poisson: only the first one will be underlined. I've actually thought about this one too, and I was thinking it was an acceptable compromise as it's an edge case and their output format is not that nice (marking and mistakes are not directly linked).

But after thinking about it again it should not be such a problem to parse and associate to spelling mistake, and it's more clean indeed, so I'll update the PR to work with the marked text.

jerome-poisson: @petrjasek I've changed the code it's not using `mistakes` anymore, only the marker. Now the same mistake done several times is underlined every time as expected.

superdesk/text_checkers/spellcheckers/leuven_dutch.py
```diff
@@ -0,0 +1,90 @@
...
+            raise SuperdeskApiError.internalError("Unexpected return code from {}".format(self.name))
+
+        data = r.json()
+
+        err_list = []
+        check_data = {'errors': err_list}
+        len_end_marker = len(END_MARKER)
+        output = data['spellingchecker']['output']
+        marked = output['marked'].split(START_MARKER)
+        # the first item in marded is unmarked text, we start our index there
```
petrjasek: typo `s/marded/marked/`

jerome-poisson: fixed, thanks

PR #1626, opened 2019-07-11
SPELLCHECKERS] default spellchecker implementation + discard spelling…

jerome-poisson:
… mistakes

Default spellchecker is a basic spellchecker working with personal
dictionaries.

Spelling mistakes for words in personal dictionary are discarded by
default (this is controlled by the new `use_internal_dict` key while
checking a text).

Also improved spellcheckers documentation.

SDBELGA-132

### DIFF
--- a/superdesk/text_checkers/spellcheckers/__init__.py
+++ b/superdesk/text_checkers/spellcheckers/__init__.py
@@ -22,6 +22,12 @@ logger = logging.getLogger(__name__)
 # can be set to False if importSpellcheckers need to be called manually
 # (e.g. in unit tests)
 AUTO_IMPORT = True
+# default spellchecker is the basic one, using internal dictionary
+SPELLCHECKER_DEFAULT = "default"
+CAP_SPELLING = "spelling"
+CAP_GRAMMAR = "grammar"
+# any language is supported
+LANG_ANY = "*"
 
 
 class SpellcheckerResource(Resource):
@@ -34,9 +40,18 @@ class SpellcheckerResource(Resource):
             'type': 'string',
             'required': True,
         },
+        'language': {
+            'type': 'string',
+            'nullable': True,
+        },
         'suggestions': {
             'type': 'boolean',
             'default': False,
+        },
+        # if True, spelling errors which are in internal dictionary will be ignored
+        'use_internal_dict': {
+            'type': 'boolean',
+            'default': True,
         },
     }
     internal_resource = False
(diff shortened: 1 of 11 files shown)


superdesk/text_checkers/spellcheckers/default.py
```diff
@@ -0,0 +1,57 @@
...
+
+class Default(SpellcheckerBase):
+    """Default spellchecker of Superdesk
+
+    This spellchecker works with internal dictionaries, in any language.
+    """
+
+    name = SPELLCHECKER_DEFAULT
+    label = "Superdesk default spellchecker"
+    capacities = ["spelling"]
```
petrjasek: imo it could use some constants to avoid spelling errors :wink: 
eg. `SpellcheckerBase.SPELLING`

superdesk/text_checkers/spellcheckers/default.py
```diff
@@ -0,0 +1,57 @@
...
+    languages = ['*']
```
petrjasek: also this could be a constant `SpellcheckerBase.ANY_LANGUAGE``

superdesk/text_checkers/spellcheckers/default.py
jerome-poisson: indeed, I'll change that, thanks

superdesk/text_checkers/spellcheckers/default.py
jerome-poisson: right

superdesk/text_checkers/spellcheckers/__init__.py
```diff
@@ -34,33 +40,105 @@ class SpellcheckerResource(Resource):
             'type': 'string',
             'required': True,
         },
+        'language': {
```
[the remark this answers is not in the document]
jerome-poisson: the error is in the payload but it would make sense to make it a requirement and return error code indeed, I'll do it.

superdesk/text_checkers/spellcheckers/__init__.py
```diff
@@ -34,33 +40,105 @@ class SpellcheckerResource(Resource):
...
+        to_remove = []
+        for error in errors:
+            if error.get('type', 'spelling') != 'spelling':
+                continue
+            if error['text'].lower() in model:
+                to_remove.append(error)
+        for error in to_remove:
+            errors.remove(error)
 
     def create(self, docs, **kwargs):
```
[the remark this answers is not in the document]
jerome-poisson: I'm not sure if I can change that easily with eve, I'll check it.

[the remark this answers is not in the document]
jerome-poisson: As I was expecting, it's not trivial to change in eve. The 201 status code is hardcoded in a `post` method that I can't override without some kind of dirty monkey patching.

We are not in REST anymore with the spellchecker, so this make sense. Anyway I don't think a `201` code is such a big deal for the moment.

PR #1630, opened 2019-07-17
SDESK-4495 Content API sends randomly inconsistent responses

ride90:


tests/content_api_test.py
```diff
@@ -574,3 +575,89 @@ def test_publish_item_with_attachments(self):
...
+
+    def test_items_default_sorting(self):
+        subscriber = {'_id': 'sub1'}
+        headers = self._auth_headers(subscriber)
+
+        self.content_api.publish(
+            {'_id': 'aaaa', 'guid': 'aaa', 'urgency': '1', 'type': 'text'},
+            [subscriber]
+        )
+        time.sleep(1)
```
jerome-poisson: why doing the sleep? I'm a bit afraid this will make testing too long.

tests/content_api_test.py
```diff
@@ -574,3 +575,89 @@ def test_publish_item_with_attachments(self):
...
+            response = c.get('items', headers=headers)
+            self.assertEqual(200, response.status_code)
+            data = json.loads(response.data)
+            self.assertListEqual([i['urgency'] for i in data['_items']], ['5', '4', '3', '2', '1'])
+
+            # with filtering
+            response = c.get('items?type=["text"]', headers=headers)
+            self.assertEqual(200, response.status_code)
+            data = json.loads(response.data)
+            self.assertListEqual([i['urgency'] for i in data['_items']], ['5', '4', '3', '2', '1'])
```
jerome-poisson: we can't really be sure the filtering worked, because all items are `text`. Would it make sense to add some other types?

tests/content_api_test.py
```diff
@@ -574,3 +575,89 @@ def test_publish_item_with_attachments(self):
...
+
+            # with filtering and custom sorting
+            response = c.get('items?type=["text"]&sort=[("versioncreated",-1)]', headers=headers)
+            self.assertEqual(200, response.status_code)
+            data = json.loads(response.data)
+            self.assertListEqual([i['urgency'] for i in data['_items']], ['5', '4', '3', '2', '1'])
```
jerome-poisson: same here, the ordering doesn't change, it would be better to change order of initial items so we can check sorting actually worked.

tests/content_api_test.py
ride90: @jerome-poisson I agree that it is a little bit weird :) The trick is that we want to test a default sorting and default sort field is `versioncreated`, so if I `self.content_api.publish` without delay `versioncreated` will be almost the same (it filled automatically in publish service). The lowest dimension is a `second` in datetime for versioncreated. That's why I use delay with 1 second. Does it make any sense?

tests/content_api_test.py
ride90: Hi Jerome, this test is not about testing a filtering, there are other tests for it, current test is only to test a default sorting applied (`versioncreated` field). Previous behavior was removing a default sorting if query filter was added, in our case it is `type=["text"]`. But for better test I will change one type to something else.

tests/content_api_test.py
ride90: I will change it here to ascending.

tests/content_api_test.py
jerome-poisson: OK it makes sense. Can you add a comment to explain that?

ride90: sure

PR #1632, opened 2019-07-23
Fix ordering of media-gallry and related items in ninjs. (SDESK-4423) & (SDESK-4463).

GyanP:
order of related items can be decided using their key ex: related--1, related--2
order of media-gallery items can be decided using `order` field ex: 'order': 0, 'order': 1


ride90:
it would be nice to have a test for it

PR #1634, opened 2019-07-25
As a user I want to be notified when a story is marked for me. (SDBELGA-124)

GyanP:


tomaskikutis:
By the way, I need a way to know when to refresh the list of items "marked for me". Can you also send a websocket message every time an item is marked for user, unmarked from user, or marked user changes?

GyanP:
@tomaskikutis that's what this PR will do. Whenever mark_for_user action will take place, it will send a WebSocket notification to all the participant user, i.e. to the user for which item is marked, from which item is unmarked and to both the users previous and the new user when item marked from one user to another.

tomaskikutis:
Can you add an additional notification not related to activity? I'd like the following event format to be used:

```
{
    "event": "item:marked",
    "item_id": "urn:newsml:localhost:5000:2018-12-20T11:16:39.471231:aea87c32-e680-4006-940e-092fbdb79b44",
    "extension": "markForUser"
}
```

There are 2 reasons I want this:
* I don't want to allow client extensions to see all websocket messages. The point of extensions is to isolate features, so extensions should only be allowed to see messages addressed to them explicitly
* I don't want markForUser client extension to depend on activities/notifications which is another feature on its own. If we can avoid unnecessary coupling, which we can in this case, then we should.

GyanP:
@tomaskikutis this is how new notification looks like

mark_for_user_notification

I am sending this along with activity notification. 
So, If I should keep both?

tomaskikutis:
:+1: the new format works for me. Yes, keep both of them so the one for the extension is not used by code outside the extension.

PR #1635, opened 2019-07-26
fix saved search report not using doc_type

petrjasek:
so it would query all always and not respect selected repos

SDANSA-309 SDANSA-307

### DIFF
--- a/apps/search/__init__.py
+++ b/apps/search/__init__.py
@@ -28,6 +28,7 @@ class SearchService(superdesk.Service):
     It can search against different collections like Ingest, Production, Archived etc.. at the same time.
     """
 
+    repos = None
     aggregations = deepcopy(common_aggregations)
 
     @property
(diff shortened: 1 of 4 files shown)


apps/search/__init__.py
```diff
@@ -149,15 +148,7 @@ def _get_types(self, req):
...
-
-        # If not repos were supplied, return the default repos
-        if repos is None:
-            return self.repos.copy()
-        else:
-            repos = repos.split(',')
-
-            # If the repos array is still empty after filtering, then return the default repos
-            return [repo for repo in repos if repo in self.repos] or self.repos.copy()
+        return es_utils.get_doc_types(repos)
```
MarkLark86: Can we still use `self.repos` here please and pass that as the second parameter to `es_utils.get_doc_types` as a default value?

Reason being is that Analytics inherits from this class and makes use of self.repos (defaulting to the one that this Search class implements). Modifying `get_doc_types` for a default value will also make it usable elsewhere (if required).

PR #1636, opened 2019-07-26
return related items for given item

petrjasek:
- searches production api
- can include or exclude the given item

SDFID-555


pablopunk:
> can include or exclude the given item

By looking at the code I don't see any option to include the item

petrjasek:
@pablopunk right, I've removed that at the end, because there was no need for it. it always excludes now

PR #1639, opened 2019-07-30
(fix): Use hachoir instead of hachoir3 package

MarkLark86:


petrjasek:
pushed also to 1.29 and 1.30 branch

Magdosoft:
trying to install using :
curl -s https://raw.githubusercontent.com/superdesk/fireq/files/superdesk/install | sudo bash
but the same error

PR #1641, opened 2019-08-01
fix related items reference in associations

petrjasek:
we should only store _id and type for superdesk items,
not the full item.

including upgrade script to fix archive/published items.

SDESK-4574

### DIFF
--- a/superdesk/data_updates/00018_20190801-145234_archive.py
+++ b/superdesk/data_updates/00018_20190801-145234_archive.py
@@ -16,24 +16,20 @@ class DataUpdate(DataUpdate):
     resource = 'archive'
 
     def forwards(self, mongodb_collection, mongodb_database):
+        related = get_resource_service('vocabularies').get(req=None, lookup={'field_type': 'related_content'})
+        archive_service = get_resource_service('archive')
         for resource in ('archive', 'published'):
             service = get_resource_service(resource)
             for item in mongodb_database[resource].find({'type': 'text', 'associations': {'$gt': {}}}):
                 update = False
                 associations = {}
                 for key, val in item['associations'].items():
-                    if val and val.get('type') == 'text' and len(val.keys()) > 2:
+                    if val and archive_service._is_related_content(key, related) and len(val.keys()) > 2:
                         update = True
                         associations[key] = {
                             '_id': val.get('item_id') or val.get('_id'),
                             'type': val['type'],
                         }
-                    elif val and val.get('_id') and len(val.keys()) == 1:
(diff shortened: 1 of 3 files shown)


superdesk/data_updates/00018_20190801-145234_archive.py
```diff
@@ -0,0 +1,43 @@
...
+    resource = 'archive'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        for resource in ('archive', 'published'):
+            service = get_resource_service(resource)
+            for item in mongodb_database[resource].find({'type': 'text', 'associations': {'$gt': {}}}):
+                update = False
+                associations = {}
+                for key, val in item['associations'].items():
+                    if val and val.get('type') == 'text' and len(val.keys()) > 2:
```
GyanP: Why are you checking only the item of type `text`? We can add `picture`, `audio` or `video` as well inside related items of type `related_content`.

petrjasek: good point, I'm not sure to be honest :wink: will fix

petrjasek: actually it's safer like that, because otherwise it would also affect featured media etc, while text items can be only used as related and they were causing the performance issue this is trying to solve

GyanP: Yep, I do agree that text can only be used as related but a picture, audio or video can also be used as related. So, why not check if the association key is a related key of type `realted_content` not `media` then. If that makes sense :)

petrjasek: makes sense, will check that :+1:

superdesk/data_updates/00018_20190801-145234_archive.py
```diff
@@ -0,0 +1,39 @@
...
+            for item in mongodb_database[resource].find({'type': 'text', 'associations': {'$gt': {}}}):
+                update = False
+                associations = {}
+                for key, val in item['associations'].items():
+                    if val and archive_service._is_related_content(key, related) and len(val.keys()) > 2:
+                        update = True
+                        associations[key] = {
+                            '_id': val.get('item_id') or val.get('_id'),
+                            'type': val['type'],
+                        }
```
GyanP: I think if you are trying to add `type` in the related items, then you should add that for those items also which have `_id` only. You added it in first commit but removed in second.

`elif val and archive_service._is_related_content(key, related) and len(val.keys()) == 1:
    type_ = mongodb_database[resource].find_one({'_id': val['_id']}, {'type': 1})
    if type_:
        update = True
        associations[key] = val
        associations[key]['type'] = type_['type']`

PR #1643, opened 2019-08-07
[SDESK-4571] Do not expire items from production if they reference an…

marwoodandrew:
… assignment


MarkLark86:
@petrjasek Being a small change we would like this backported to 1.29 please so that this can be included in our next production release.

PR #1647, opened 2019-08-15
[SDESK-4613] Remove multibyte and characters to avoid from filenames/…

marwoodandrew:
…id's used in Amazon s3 keys


MarkLark86:
@petrjasek This fixes a bug when using S3 and uploading files with filenames that are not UTF-8 compliant. This also causes an issue when receiving these files from Newsroom.

Being an important bug fix, this will also be required for 1.29 & 1.30

PR #1648, opened 2019-08-15
[FORMATTER][NINJS] Added `id` to authors

jerome-poisson:
Author id was not available in ninjs output, making it hard to
impossible to distinguish authors with same name.

This patch introduce it and change ninjs schema accordingly.

SDFID-569

### DIFF
--- a/docs/superdesk-ninjs-schema.json
+++ b/docs/superdesk-ninjs-schema.json
@@ -407,6 +407,10 @@
 			"items": {
 				"type" : "object",
 				"properties" : {
+					"code" : {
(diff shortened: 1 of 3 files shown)


docs/superdesk-ninjs-schema.json
```diff
@@ -407,6 +407,10 @@
 			"items": {
 				"type" : "object",
 				"properties" : {
+					"id" : {
```
petrjasek: what about using `code` here? that's more inline with other ninjs properties like place/person/subject/etc

PR #1649, opened 2019-08-19
fix(ap anpa wordcount) Handle missing word count

marwoodandrew:


MarkLark86:
@petrjasek This is a bug affecting 1.29 and 1.30 as well. I will cherry-pick into those branches

PR #1656, opened 2019-08-28
Dictionary suggestions should not be case sensitive. (SDESK-4636)

GyanP:
suggestions

### DIFF
--- a/apps/spellcheck/spellcheck.py
+++ b/apps/spellcheck/spellcheck.py
@@ -9,6 +9,7 @@ def norvig_suggest(word, model):
     """
     NWORDS = model
     alphabet = 'abcdefghijklmnopqrstuvwxyz'
+    suggestions = []
 
     def edits1(word):
         splits = [(word[:i], word[i:]) for i in range(len(word) + 1)]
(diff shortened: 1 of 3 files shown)


petrjasek:
actually I'm not sure it's a valid requirement, it won't suggest fixes for capital letters in names

GyanP:
> actually I'm not sure it's a valid requirement, it won't suggest fixes for capital letters in names

Ahh, I thought we should match spelling, not the case. But the suggestions are always in lowercase.

suggestions

and that's why I thought we have to match spelling only not the case

GyanP:
The use case was like if I type add `Nareg` to the dictionary and then type `Naareg`, the spellchecker will show the error but when I right-click to see the suggestions, it shows no suggestions found. And that's because we find the suggestions by converting them the word to lowercase, see [here](https://github.com/superdesk/superdesk-core/blob/master/apps/spellcheck/spellcheck.py#L31).
We added `Nareg` to the dictionary but we are finding the suggestion for `naareg` not for `Naareg`

karelpetrak:
We can assume that words which are stored in a dictionary with a capital letter are always supposed to be written like that (names in English, nouns in German etc.). While words which are in the dictionary in lowercase only can be written both with a capital letter (beginning of sentence) or lowercase.

Use case 1)
in dictionary: Obama
I type: Obana
suggestion: Obama (and not obama)

Use case 2)
in dictionary: Obama
I type: obama OR obana
suggestion: Obama (and not obama)

use case 3)
in dictionary: banana
I type: Babana
suggestion: Banana AND banana

use case 4)
in dictionary: banana
I type: babana
suggestion: Banana AND banana

PR #1657, opened 2019-08-30
[SDBELGA-125] Upload video from Superdesk to video server.

idsvn:
-Setting a switch for enable video server. if disable use upload normal in Superdesk.
-Add video editor service for integrating with Superdesk.
-Upload video to video server and record data to Superdesk.


superdesk/media/video_editor.py
```diff
@@ -0,0 +1,216 @@
...
+        raise NotImplementedError
+
+    def post(self, file_storage):
+        """Create new project.
+
+        :param file_storage:
+        """
+        raise NotImplementedError
+
+    def duplicate(self, project_id, updates):
```
ride90: do we need `updates` here?

apps/archive/archive_media.py
```diff
@@ -102,3 +114,24 @@ def get_file_from_document(self, doc):
...
+        :return:
+        """
+        file = doc.get('media')
+        # upload video to video server
+        res = self.videoEditor.post(doc.get('media'))
+        doc['media'] = res['_id']
+        metadata = res.get('metadata')
+        # create renditions
+        rend = {
+            'href': res.get('url'),
```
ride90: We should not use direct reference to a video server here, because a browser will not have a direct access to it. 
All requests to `raw` resources (video, preview thumbnail and timeline thumbnails) should go through superdesk.
As an option we can do some nginx magic for this `raw` urls. 

UPD: This will be solved with nginx config, so let's keep it as it is.

apps/archive/archive_media.py
```diff
@@ -102,3 +114,24 @@ def get_file_from_document(self, doc):
...
+            'media': res.get('_id'),
```
ride90: It's better to use `video_editor_id` not `media` to store `_id` of project. It will give us an ability to know if this rendition belongs to video editor or not.

superdesk/media/video_editor.py
```diff
@@ -0,0 +1,216 @@
...
+                raise SuperdeskApiError.conflictError(message=json)
+            if resp.status_code == 404:
+                raise SuperdeskApiError.notFoundError(message=json)
+            if resp.status_code == 500:
+                raise SuperdeskApiError.internalError(message=json)
+        return json
+
+    def get(self, project_id):
+        try:
+            resp = self.session.get(self._url(str(project_id)))
```
ride90: just a small notice: it's better to move `str` into `_url` method.

PR #1660, opened 2019-09-04
[TESTING][BEHAVE] Network calls mocking

jerome-poisson:
This patch introduce a way to mock network calls in behave, and uses it
in some steps. The mocking uses a special JSON file, and the method
parsing it is separated so it can be used easily in others steps.

Documentation explaining how to use it has also been written.

SDESK-4586


petrjasek:
nice docs :+1:

PR #1662, opened 2019-09-05
fix(ap category mapping) Numbers in sluglines failed to parse and fix…

marwoodandrew:
… mapping


marwoodandrew:
Will remove AAP specific dependency

marwoodandrew:
Removes hard coded dependency on AAP's categories.

PR #1665, opened 2019-09-06
(fix-requirements) Add responses lib in setup.py

MarkLark86:


dev-requirements.txt
```diff
@@ -11,6 +11,6 @@ sphinx-autobuild
 sphinxcontrib-plantuml
 docutils==0.12
 requests-mock
-responses
+responses>=0.10.6,<0.11
```
petrjasek: I guess you can remove it from here, it should be installed via setup.py

MarkLark86:
Closing this. After review, responses lib should be development only. Therefor other repos, such as planning, require responses lib in their dev-requirements.txt file

PR #1667, opened 2019-09-06
make sure we don't send multiline email subjects

petrjasek:
which is not possible, so the lib raises an error

### DIFF
--- a/superdesk/emails/__init__.py
+++ b/superdesk/emails/__init__.py
@@ -50,7 +50,7 @@ def send_email(self, subject, sender, recipients, text_body, html_body, cc=None,
         return
 
     try:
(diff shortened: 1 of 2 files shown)


superdesk/emails/__init__.py
```diff
@@ -50,7 +50,7 @@ def send_email(self, subject, sender, recipients, text_body, html_body, cc=None,
         return
 
     try:
-        msg = SuperdeskMessage(subject, sender=sender, recipients=recipients, cc=cc, bcc=bcc,
+        msg = SuperdeskMessage(subject.split('\n')[0], sender=sender, recipients=recipients, cc=cc, bcc=bcc,
```
ride90: I think we need `except` here

petrjasek: yep just added that

PR #1668, opened 2019-09-09
Video editor

idsvn:
Function edit:
- Capture and upload a thumbnail.
- Edit video (crop, cut, rotate, ..) video.


ride90:
@idsvn why do we generate a timeline thumbnails in advance https://github.com/superdesk/superdesk-core/blob/develop/apps/archive/archive_media.py#L53 ? Just to improve a user feeling when editor is opened, right?

ride90:
@idsvn I think this is not a factory actually https://github.com/superdesk/superdesk-core/blob/develop/superdesk/media/video_editor.py#L14 more like an interface/base-class
Any reason to call it factory?

idsvn:
https://github.com/superdesk/superdesk-core/pull/1668#pullrequestreview-289111442
Hi @ride90,  Maybe your superdesk-ui-framework lib is old version, so it cannot load `css` files. Please check and update it to newest version.

apps/archive/archive_media.py
```diff
@@ -45,12 +45,12 @@ def on_create(self, docs):
             doc['media'].seek(0)
             file_type = content_type.split('/')[0]
             if file_type == 'video' and app.config.get("VIDEO_SERVER_ENABLE"):
-                if not self.videoEditor.check_video_server():
+                if not self.video_editor.check_video_server():
```
ride90: Do you think it's necessary to use this method? 
If the server is unavailable (`self.video_editor.check_video_server`) we `raise SuperdeskApiError`, but we do the same in let's say in `self.video_editor.create_timeline_thumbnails`. We raise `SuperdeskApiError` in all `VideoEditorWrapper` methods which do network calls, so it's the same. The only benefit I see in the current approach is a better log message, but we send 2 requests instead of 1. It's not a big deal actually.

apps/archive/archive_media.py
```diff
@@ -76,7 +76,7 @@ def on_create(self, docs):
                 for file_id in inserted:
                     delete_file_on_error(doc, file_id)
                 if res:
-                    self.videoEditor.delete(res.get('_id'))
+                    self.video_editor.delete(res.get('_id'))
```
ride90: For example, here we don't do `self.video_editor.check_video_server()`.

ride90:
@idsvn please, try to fix a build and if you can add some tests 👍

apps/archive/archive_media.py
idsvn: Hi @ride90, we do this method, because we want if the video server is disconnect it will auto upload to `Superdesk`. But now we think it not necessary, we should raise error when video server enable but its disconnect.
We will remove checking method.

apps/video_edit/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        res = super().find_one(req, **lookup)
+        if req is None:
+            return res
+
+        video_id = res['media']
+        if req.args.get('action') == 'timeline':
+            response = self.video_editor.create_timeline_thumbnails(video_id, req.args.get('amount', 40))
+            return {
+                config.ID_FIELD: video_id,
+                **response
```
ride90: This will raise an error. In this case, `response` is a list.

ride90:
Hi @idsvn , please, don't use force push. It overwrites a history and it's impossible to follow what was done.

apps/video_edit/__init__.py
idsvn: Hi, Oleg,
Should we only update response of timeline from video server ([change](https://github.com/superdesk/video-server/pull/19/files))?
It make format of response is homogeneous, easier to handling.

ride90: @idsvn I'm not sure If I understood. In case when action is `timeline` it raises an error.

apps/video_edit/__init__.py
```diff
@@ -0,0 +1,186 @@
...
+        ids = []
+        for doc in docs:
+            item = doc.get('item')
+            item_id = item[config.ID_FIELD]
+            renditions = item['renditions']
+            video_id = renditions['original'].get('video_editor_id')
+            if not video_id:
+                raise SuperdeskApiError.badRequestError(message='Missing video_editor_id')
+            if 'capture' not in doc and 'edit' not in doc:
+                raise SuperdeskApiError.badRequestError(message='Must include either capture or edit param')
```
petrjasek, line 26: can you make those messages translatable?

apps/archive/archive_media.py
```diff
@@ -45,12 +45,10 @@ def on_create(self, docs):
             doc['media'].seek(0)
             file_type = content_type.split('/')[0]
             if file_type == 'video' and app.config.get("VIDEO_SERVER_ENABLE"):
-                if not self.videoEditor.check_video_server():
-                    raise SuperdeskApiError(message="Cannot connect to videoserver", status_code=500)
                 # upload media to video server
                 res, renditions, metadata = self.upload_file_to_video_server(doc)
                 # get thumbnails for timeline bar
-                self.videoEditor.get_timeline_thumbnails(doc.get('media'), 40)
+                self.video_editor.create_timeline_thumbnails(doc.get('media'), 60)
```
petrjasek, line 51: I wonder if that 60 should be rather a constant or a config option

ride90: I will make it constant for now

apps/archive/archive_media.py
```diff
@@ -44,13 +43,11 @@ def on_create(self, docs):
             content_type = magic.from_buffer(doc['media'].read(1024), mime=True)
             doc['media'].seek(0)
             file_type = content_type.split('/')[0]
-            if file_type == 'video' and app.config.get("VIDEO_SERVER_ENABLE"):
-                if not self.videoEditor.check_video_server():
-                    raise SuperdeskApiError(message="Cannot connect to videoserver", status_code=500)
+            if file_type == 'video' and app.config.get("VIDEO_SERVER_ENABLED"):
```
idsvn, line 46: Should update this on `default_settings` too

ride90: hi @idsvn , did not add all files to a commit

PR #1672, opened 2019-09-19
Introduce configuration for the main left hand side toolbar

tomaskikutis:
SDESK-4668


apps/archive/__init__.py
```diff
@@ -89,9 +89,11 @@ def init_app(app):
     register_component(ItemAutosave(app))
     register_model(ItemAutosaveModel(EveProxy(superdesk.get_backend())))
 
+    superdesk.privilege(name='monitoring_view', label='Can access monitoring view', description='')
     superdesk.privilege(name='archive', label='Archive', description='User can view the published content.')
     superdesk.privilege(name='ingest', label='Ingest', description='User can view content in ingest and fetch it.')
-    superdesk.privilege(name='spike', label='Spike', description='User can spike content.')
+    superdesk.privilege(name='spike_read', label='Can see spiked items', description='User can see spiked content.')
+    superdesk.privilege(name='spike', label='Can spike items', description='User can spike content.')
```
MarkLark86: These labels are more of a description then a label. This would mean the `Spike` role would now sorted with other roles starting with `C`.

We should be displaying the description of privileges in the UI and leave the names short, i.e. `Spike` and show `User can spike content` on hover (or another column in the table, as we have the sapce).

tomaskikutis: We don't currently display a description in the UI and we want to ship this feature fast so I'll leave it as is for now. We are thinking about revamping the entire privileges section in the future so privileges are grouped. That would be a good opportunity to take a look at naming as well. I don't see it being clear if we name privileges spike1, spike2 and explain it in the description. If we can describe a privilege clearly in a few words, say 6 for example, I'd put it in the name a leave a description blank if there's not anything else to add.

apps/archive/__init__.py
```diff
@@ -89,9 +89,11 @@ def init_app(app):
     register_component(ItemAutosave(app))
     register_model(ItemAutosaveModel(EveProxy(superdesk.get_backend())))
 
+    superdesk.privilege(name='monitoring_view', label='Can access monitoring view', description='')
     superdesk.privilege(name='archive', label='Archive', description='User can view the published content.')
     superdesk.privilege(name='ingest', label='Ingest', description='User can view content in ingest and fetch it.')
-    superdesk.privilege(name='spike', label='Spike', description='User can spike content.')
+    superdesk.privilege(name='spike_read', label='Can see spiked items', description='')
```
petrjasek: imo `spike_view` would be more consistent

apps/highlights/__init__.py
```diff
@@ -20,6 +20,8 @@ def init_app(app):
     service = GenerateHighlightsService(endpoint_name, backend=get_backend())
     GenerateHighlightsResource(endpoint_name, app=app, service=service)
 
+    superdesk.privilege(name='highlights_read', label='Can see a list of highlights',
```
petrjasek: this could be `highlights_view`

apps/saved_searches/__init__.py
```diff
@@ -40,6 +40,9 @@ def init_app(app):
     service = SavedSearchItemsService(endpoint_name, backend=superdesk.get_backend())
     SavedSearchItemsResource(endpoint_name, app=app, service=service)
 
+    superdesk.privilege(name='use_global_saved_searches',
```
petrjasek: and this `global_search_view`

apps/workspace/__init__.py
```diff
@@ -15,7 +15,12 @@
 def init_app(app):
     superdesk.register_resource('workspaces', WorkspaceResource, WorkspaceService,
                                 privilege=['POST', 'PATCH', 'DELETE'])
+
     superdesk.register_default_user_preference('workspace:active', {
         'type': 'string',
         'workspace': ''
     })
+
+    superdesk.privilege(name='dashboard',
```
petrjasek: `dashboard_view`

apps/archive/__init__.py
MarkLark86: In Analytics and Planning we use short labels and descriptive descriptions.
i.e.
label: Planning - Spike Event Items
description: Ability to spike an Event

When alphabetically sorted, this privilege is placed with other Planning privileges starting with 'S'.

Adding one more column in the privileges table would be a very quick and easy way to show the description of a Privilege, while still have a short label/name of the privilege.

The label should be short and the description should describe what it is for. Such as:
label: Spike
description: User can spike items

label: Spike - View
description: Use can see spiked items

Later on when refactoring we can add the Superdesk module or group by any other means. But simply adding in a column in the UI to show the description should be enough for now.

tomaskikutis:
I think it's better to organize things by functionality, not views where the functionality is in. If the user doesn't have `highlights_read` privilege, I expect him to not be able to read highlights regardless which page or view the component is rendered in. Let's say we build a dashboard widget to show the latest highlights. It should not work for people without a privilege even if it's in a **dashboard view** and not **highlights view**.

apps/archive/__init__.py
tomaskikutis: I mostly agree with you, except regarding the quickness to add it to the UI. It's quick technically to make the text appear, but it [doesn't look good](https://user-images.githubusercontent.com/2076953/65593523-f0380a80-df90-11e9-8562-0678fab0339e.png). There are also duplicate data that needs to be cleaned up.

petrjasek: what about using [tooltips](https://ui-framework.superdesk.org/#/components/tooltips) ?

tomaskikutis: I've added tooltips in the [client PR](https://github.com/superdesk/superdesk-client-core/pull/3118)

screenshot_9

MarkLark86: The tooltips are truncated if the description is too long. But I guess this will do for now

tomaskikutis:
@petrjasek do you want me to change anything else?

PR #1674, opened 2019-09-24
[AUTH SERVER] Authorization Server

jerome-poisson:
This is the initial implementation of the Authorization Server. It is
used as part of the OAuth2 authorization of third party client, used
with Superdesk Production API.

The server use `Client Credentials` grant.

Check /docs/auth_server for more details.

SDBELGA-167

### DIFF
--- a/docs/auth_server.rst
+++ b/docs/auth_server.rst
@@ -1,0 +1,179 @@
+Authentication Server
+=====================
+
+Superdesk provides an OAuth2 authentication server for the production API. The server is
+using the `Client Credentials Grant`_ with JWT access token.
+
+The endpoint used in Superdesk is ``/api/auth_server/token``.
+
+.. _Client Credentials Grant: https://tools.ietf.org/html/rfc6749#section-4.4
+
+Overview
+--------
+
+Here is the sequence of the authentication, which must entirely be done in HTTPS:
+
+.. uml::
+
+    title Authentication server flow in Superdesk (for third party clients)
+
+    boundary client
+    control "authentication server" as auth_server
+    boundary "resource server\n(//Production API//)" as res_server
+
+    client -> auth_server: request token
+    note right
+        using **client credentials** grant
+        and **HTTP Basic Authentication**
+    end note
+
+    alt <color: green>success</color>
+        client <- auth_server : **JWT** access token
+        note right
+            JWT payload contains the **client id**,
+            the **expiration date**, and the **scope**
+        end note
+        loop
+            client -> res_server: request with access token
+            res_server -> res_server: validate **JWT** token signature, \nclient_id, expiration date and scope
+            alt <color:green>valid token</color>
+                client <- res_server: request answer
+            else <color:DarkRed>invalid token (bad signature, expired token)</color>
+                client <- res_server: HTTP 401 (Unauthorized) error
+            else <color:DarkRed>invalid scope</color>
+                client <- res_server: HTTP 403 (Forbidden) error
+            end
+        end
+
+    else <color:DarkRed>failure</color>
+        auth_server -> client: HTTP error\n(400 or something else)
+        note right
+            a //JSON// payload with ""error"" key should be present,
+            the value being an explanation of the issue
+        end note
+
+    end
+
+Settings
+--------
+
+The variables explained in the table below can be set, either as environment variable or
+in ``settings.py``.
+
+Please note that for ``AUTH_SERVER_SHARED_SECRET`` we recommend to use environment
+variable instead of settings to limit attack surface if a malicious user gains access to
+your settings.
+
+
+============================  ========================================================
+name                          explanation
+============================  ========================================================
+AUTH_SERVER_EXPIRATION_DELAY  delay in seconds when the access token is valid,
+                              after this delay, a new token is needed.
+                              Default to 86400 (1 day)
+AUTH_SERVER_SHARED_SECRET     a secret shared with resource server to sign/validate
+                              the access token. We recommend to set this value using
+                              environment variable.
+============================  ========================================================
+
+Allowing a client
+-----------------
+
+A client needs to be registered with Superdesk authentication server.
+
+You need to specify a *client id*, a *password* and the *scopes* allowed (see below).
+
+This is done using ``python manage.py auth_server:register_client``, please check the
+appropriate documentation in :ref:`cli` section.
+
+
+Scope
+-----
+
+To interact with the production API, clients need to be explicitly allowed to perform
+actions. This is done using the ``--scope`` argument when registering the client. The
+possible values are:
+
+=============  ======================================================
+permission     explanation
+=============  ======================================================
+ARCHIVE_READ   The client can read items from archive collection
+DESK_READ      The client can get desks metadata
+PLANNING_READ  The client can access planning, events and assignments
+CONTACT_READ   The client can get contacts metadata
+USERS_READ     The client can retrieve users
+=============  ======================================================
+
+Access token
+------------
+
+If client is allowed, an access token is generated and can be used to access the desired
+resources. The token is a `JSON Web Token`_ (JWT), which is signed, allowing the resource
+server to verify it.
+
+The payload of the token will contain the following keys:
+
+=========  ====================================================================
+key        explanation
+=========  ====================================================================
+client_id  id of the allowed client
+iss        principal that issued the JWT, it's always ``Superdesk Auth Server``
+iat        time when the JWT was issued (Unix time)
+exp        expiration time (Unix time)
+scope      list of allowed scopes (see above)
+=========  ====================================================================
+
+.. _JSON Web Token: https://tools.ietf.org/html/rfc7519
+
+Security
+--------
+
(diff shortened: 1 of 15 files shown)


jerome-poisson:
**note**: I haven't pushed yet the testing and JWT part.

docs/auth_server.rst
```diff
@@ -0,0 +1,108 @@
...
+USERS_READ     The client can retrieve users
+=============  ======================================================
+
+Access token
+------------
+
+When created, the access token is stored in database in ``auth_server_token`` collection.
+You can use the service of the same name to access it.
+
+Please note that ``JWT`` does sign its payload but doesn't encrypt it, so you must be
```
MarkLark86: Why is the web token not encrypted? Let me know if I am incorrect, but isn't this token sent to the client so they can send this token with every request? The client only uses the token to send back to the server with every request to authorise the request. Therefor it would be best if this token was encrypted (as the token leaves your known network)

jerome-poisson: The JWT token itself is used only between client and authorisation server, and there is not much sensitive data inside (client id mostly). The point of using it is stronger auth, and not passing the client secret on the wire. It's done via HTTPS anyway, so there is an encryption layer. It must mostly not be leaked to avoid replay attack, but there is protection against that (JTI).
The authorisation server then give an access token (again via HTTPS) which is a random series of bytes, this is the one which will then be used between the client and the resource owner. This one can be revoked.

In short: as long as HTTPS is OK, the token is OK. If HTTPS is broken, we have bigger problems anyway.

petrjasek:
I wonder why we store the tokens?

jerome-poisson:
> I wonder why we store the tokens?

@petrjasek For the production API to get it. Do you have an other idea to share the access token with Production API?

petrjasek:
so that's not token for clients?

docs/auth_server.rst
```diff
@@ -0,0 +1,146 @@
...
+======  ===============  ======================================================
+
+Here is an example of client configuration:
+
+.. code:: python
+
+    AUTH_SERVER_CLIENTS = [
+        {
+            "id": "123456",
+            "secret": "789101112",
```
petrjasek: imo we shouldn't store the secret in plain text on server, and ideally this would be in db (not plain text again)

jerome-poisson:
> so that's not token for clients?

that's the token the client get after authentication, and it will use it with production API. But Production API need to know it to check that it's valid and what is the scope of this token. You can check the sequence diagram (it's not rendered here but quite understandable in text) to see the exact flow.

docs/auth_server.rst
jerome-poisson: I was thinking that this could be a first step, and later we can add a client UI for that with a proper API and storage in db.

An other option would be to create a `manage.py` command for that, should I go this way?

jerome-poisson: Also I don't see how this could be stored encrypted in db without a masterkey or something like that. We can't store only a hash, the secret is need by authlib to validate the signature.

petrjasek: isn't it only used here? https://github.com/superdesk/superdesk-core/pull/1674/files#diff-bcd8f2afa533368cd035ae88d6b012c6R41

petrjasek:
I thought we will use jwt tokens for clients so that production api would only check if jwt is valid without asking auth server

jerome-poisson:
note: following @petrjasek comment we've had a discussion at https://sourcefabric.slack.com/archives/CN0683NF7/p1570021331002000 .
We'll wait for more input to see if we keep this workflow or change it.

docs/auth_server.rst
jerome-poisson: no, it's also used by JWT for signing auth token.

ride90:
Hi @jerome-poisson , 

I've talked with @petrjasek and we decided not to merge a PR for now and wait for you to provide some changes.
We can discuss details in the chat channel. Here is a main updates to deliver:
- send `client id` + `client secret` not JWT
- don't store JWT token + scopes in the DB. Auth server + prod API will share `JWT_SECRET` to verify access token and scopes will go to the JWT payload. In this case no need to hit DB with every request on prod API.

Talk to you when you are back 🍻

jerome-poisson:
I've updated with new workflow, rewrote the doc and tests to fit, and added some commands to manage clients (it's not done in `settings.py` anymore).

I would recommand to render the docs locally if you wanna see it, the plantuml schema is far more readable and will help for the global understanding.

Also note that it would not really be hard to use a public/private key which could improve security in case of separate resource owner, but the current shared secret solution is probably sufficient for now.

superdesk/auth_server/clients.py
```diff
@@ -0,0 +1,182 @@
...
+
+    If the client already exists, it will be updated. You may use ``--password`` to update
+    the passsword, otherwise the former password is kept. Keep empty to prompt password,
+    this is better for security as the password won't appear on your screen, or in your
+    shell history or in your process list.
+
+    Examples:
+    Register a new client with id ``123456`` and allowed to reach items in archive and get users::
+
+        $ python manage.py auth_server:register_client 123456 -s ARCHIVE_READ -s USERS_READ
```
petrjasek: imo would be better (for future UI) to register clients with only name and generate id and secret by superdesk

jerome-poisson: hey. Well the `client_id` and `password` must be known by the client, so if SD generates them that means showing them, and that's not great to show the password in clear on the screen (that's the reason why I've implemented a prompt).

For future UI I was expecting an admin panel where SD client calls this API, in this case the ID and password can be generated by SD client it's not a problem.

On a side node there is currently no client name stored, but `client_id` is a free string, so we could use it to put a name.

petrjasek: yep they must be known by the client, so that command must return it. but I don't know why the client would set a password, it's an app after all. here is how it's done by google, same is azure and onelogin at least, you set a name and get this:
oauth

jerome-poisson: so the password is shown in clear on the screen. I can do that, it's not a big change. Should I add a client name?

petrjasek: it's shown clear on the screen. and the name you put in the form before this popup shows up.

jerome-poisson: `password` and `client id` are now generated and `name` has been added.

docs/auth_server.rst
```diff
@@ -0,0 +1,179 @@
...
+
+.. code:: sh
+
+    #!/bin/sh
+
+    CLIENT_ID=123456
+    CLIENT_SECRET=789101112
+    URL=http://127.0.0.1:5000/api/auth_server/token
+
+    curl -u ${CLIENT_ID}:${CLIENT_SECRET} -XPOST ${URL} -F grant_type=client_credential
```
ride90: Should it be grant_type=client_credential**s**?

superdesk/default_settings.py
```diff
@@ -739,3 +740,8 @@ def local_to_utc_hour(hour):
 
 #: Whether to generate custom crops on image upload by default or not, if false means generate else don't generate
 NO_CUSTOM_CROPS = strtobool(env('NO_CUSTOM_CROPS', 'false'))
+
+#: Authorisation server
+
+#: how long an authorisation token is valid (in seconds)
+AUTH_SERVER_EXPIRATION_DELAY = env('AUTH_SERVER_EXPIRATION_TIME', 60 * 60 * 24)  # 1 day by default
```
ride90: I think `AUTH_SERVER_SHARED_SECRET = env('AUTH_SERVER_SHARED_SECRET' , '')` is missed here

superdesk/auth_server/clients.py
```diff
@@ -0,0 +1,257 @@
...
+from superdesk.auth_server.scopes import allowed_scopes
+
+
+logger = logging.getLogger(__name__)
+
+
+class AuthServerClientsResource(superdesk.Resource):
+
+    schema = {
+        'client_id': {
```
petrjasek: it could store it as `_id` to avoid additional id and it would have unique index out of the box

jerome-poisson: yes I've thought about that, but `_id` has a restricted format, and I would to have a free form for client id. It easier for some use cases, like it is the case now with testing (I can specify any id easily in the feature).

docs/auth_server.rst
jerome-poisson: indeed, good catch.

superdesk/auth_server/clients.py
petrjasek: :ok_hand:

petrjasek: but would be good to add unique index for it then

superdesk/default_settings.py
jerome-poisson: I think I didn't put it here because I didn't wanted a default value, but I could check against empty string, so I'll update.

superdesk/auth_server/clients.py
jerome-poisson: but thinking about it again, there is not real benefit of free form id, for the tests I'll just use an id which is valid for Mongo.

PR #1676, opened 2019-09-25
Internal attachments

pablopunk:
SDFID-574

- [x] Remove internal attachments on http push
- [x] Remove internal attachments on content api

Client part: https://github.com/superdesk/superdesk-client-core/pull/3120


superdesk/metadata/item.py
```diff
@@ -642,6 +643,19 @@
 }
 
 
+def remove_metadata_for_publish(item):
```
jerome-poisson: a docstring would be needed here to explain what this method is doing

PR #1689, opened 2019-10-10
Add ability to specify custom S3 endpoint

niklasetc:
This pull-request adds the ability to specify a custom S3-compatible server by setting the environment variable AMAZON_ENDPOINT_URL.


superdesk/storage/amazon_media_storage.py
```diff
@@ -68,13 +68,23 @@ class AmazonMediaStorage(MediaStorage):
...
     def __init__(self, app=None):
         super().__init__(app)
-        self.client = boto3.client(
-            's3',
-            aws_access_key_id=self.app.config['AMAZON_ACCESS_KEY_ID'],
-            aws_secret_access_key=self.app.config['AMAZON_SECRET_ACCESS_KEY'],
-            region_name=self.app.config.get('AMAZON_REGION'),
-            config=Config(signature_version='s3v4'),
-        )
+        if self.app.config['AMAZON_ENDPOINT_URL']:
```
petrjasek: I think we don't need that if, there could be:
```
endpoint_url=self.app.config['AMAZON_ENDPOINT_URL'] or None,
```

PR #1690, opened 2019-10-11
Generate custom renditions on image upload by default.(SDESK-4742)

GyanP:


apps/archive/archive_media.py
```diff
@@ -54,7 +54,9 @@ def on_create(self, docs):
             else:
                 file, content_type, metadata = self.get_file_from_document(doc)
                 inserted = [doc['media']]
-                rendition_spec = get_renditions_spec(no_custom_crops=True)
+                # set no_cutom_crops to False so that custom crops are generated automatically on media upload
+                # see (SDESK-4742)
+                rendition_spec = get_renditions_spec(no_custom_crops=False)
```
petrjasek: imo we should make this configurable. and there is typo in comment `no_cutom_crops`

GyanP: Its default value should be false or true?

petrjasek: I think it should stay as it is

PR #1691, opened 2019-10-13
[SDESK-4708] Create filter condition on on existence of featuremedia

marwoodandrew:


MarkLark86:

apps/content_filters/filter_condition/filter_condition_field.py
```diff
@@ -34,7 +34,8 @@ class FilterConditionFieldsEnum(Enum):
     ednote = 15,
     place = 16,
     ingest_provider = 17
-    embargo = 18
+    embargo = 18,
```
petrjasek: I think there is no need for commas

apps/content_filters/filter_condition/filter_condition_field.py
```diff
@@ -242,3 +245,22 @@ def is_in_article(self, article):
 
     def get_value(self, article):
         return article['extra'].get(self.field.name, '')
+
+
+class FilterConditionFeatueMediaField(FilterConditionField):
```
petrjasek: typo: `Featue`

apps/content_filters/filter_condition/filter_condition_field.py
```diff
@@ -242,3 +245,22 @@ def is_in_article(self, article):
...
+    def __init__(self, field):
+        self.field = FilterConditionFieldsEnum.featuremedia
+        self.entity_name = 'associations.featuremedia._id'
+        self.field_type = bool
+
+    def get_value(self, article):
+        return article.get('associations', {}).get('featuremedia', {}).get('_id')
```
petrjasek: I think `featuremedia` can be `None` (when you remove association) and this would raise exception

apps/content_filters/filter_condition/filter_condition_operator.py
```diff
@@ -180,3 +183,24 @@ def does_match(self, article_value, filter_value):
...
+class ExistsOperator(FilterConditionOperator):
+    def __init__(self, operator):
+        self.operator = FilterConditionOperatorsEnum[operator]
+        self.get_mongo_operator = '$exists'
+        self.get_elastic_operator = '{{"exists": {{"field": "{}"}} }}'
+
+    def does_match(self, article_value, filter_value):
+        # exists
+        if filter_value:
+            if article_value:
```
petrjasek: seems like `return article_value`

apps/content_filters/filter_condition/filter_condition_field.py
marwoodandrew: Yep

apps/content_filters/filter_condition/filter_condition_field.py
marwoodandrew: Fixed

apps/content_filters/filter_condition/filter_condition_field.py
marwoodandrew: fixed

apps/content_filters/filter_condition/filter_condition_operator.py
marwoodandrew: simplified

PR #1694, opened 2019-10-15
fix routing ingested items on every ingest update

petrjasek:
when versioncreated is missing don't populate it with current time
in ingest so it's not used for new item checking.

it will be populated later when item is stored to ingest collection.

SDESK-4650

### DIFF
--- a/apps/io/feeding_services/reuters.py
+++ b/apps/io/feeding_services/reuters.py
@@ -258,7 +258,7 @@ class ReutersHTTPFeedingService(HTTPFeedingService):
         result_items = []
         while items:
             item = items.pop()
-            self.add_timestamps(item)
+            self.localize_timestamps(item)
             try:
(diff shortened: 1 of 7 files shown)


petrjasek:
`pubDate` is mapped to both `published` and `updated` by feedparser (if updated is not specified via specific field)

MarkLark86:
> `pubDate` is mapped to both `published` and `updated` by feedparser (if updated is not specified via specific field)

Understood.

So the cause of the symptom is that the feed is not providing a date of any kind on some items, which is a problem with the feed provider not our system. In that case, with your solution, only the first item will be ingested and future updates will be discarded. Without looking at the content itself, there is no other way around the problem with these incorrect feeds. We don't see this as a problem.

petrjasek:
right, it handles missing timestamps.
future updates will check the data, if there are any changes it will route it again.

PR #1700, opened 2019-10-23
Send notification when an article marked for user is spiked or published so that list can be updated. [SDBELGA-192]

GyanP:
SDBELGA-200


apps/publish/published_item.py
```diff
@@ -170,8 +172,11 @@ def on_update(self, updates, original):
 
     def on_updated(self, updates, original):
         if 'marked_for_user' in updates:
+            updated = copy(original)
```
petrjasek: that's like `updated = original.copy()` - or did you mean to use `copy.deepcopy()` ?

GyanP: Yup, it's a shallow copy.

PR #1702, opened 2019-10-23
Remove internal attachments on ninjs formatter

pablopunk:
SDFID-574

This is going to branch `release/1.32`

### DIFF
--- a/superdesk/metadata/item.py
+++ b/superdesk/metadata/item.py
@@ -646,6 +646,15 @@ crop_schema = {
     'CropBottom': {'type': 'integer'}
 }
 
+def is_attachment_public(attachment):
+    """Retuns true if attachment is public. False if it's internal.
+
+    :param attachment: Attachment object or id inside attachment attribute
+    :return: boolean
+    """
+    if attachment.get('attachment'): # retrieve object reference
+        attachment = superdesk.get_resource_service('attachments').find_one(req=None, _id=attachment['attachment'])
+    return not attachment.get('internal')
 
 def remove_metadata_for_publish(item):
     """Remove metadata from item that should not be public.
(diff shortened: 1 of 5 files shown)


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -426,16 +427,17 @@ def _format_attachments(self, article):
...
+            if is_attachment_public(attachment): # don't save internal attachments
+              output.append({
+                  'id': str(attachment['_id']),
+                  'title': attachment['title'],
+                  'description': attachment['description'],
+                  'filename': attachment['filename'],
+                  'mimetype': attachment['mimetype'],
+                  'length': attachment.get('length'),
+                  'media': str(attachment['media']),
+                  'href': '/assets/{}'.format(str(attachment['media'])),
```
petrjasek: it's better to use `app.media.url_for_media`

superdesk/metadata/item.py
```diff
@@ -646,6 +646,15 @@
     'CropBottom': {'type': 'integer'}
 }
 
+def is_attachment_public(attachment):
```
petrjasek: imo this should be in `superdesk.attachments` module

superdesk/publish/formatters/ninjs_formatter.py
pablopunk: you mean instead `href`? cause I didn't change any of this code, I just indented it under the `if`

superdesk/metadata/item.py
pablopunk: agree

superdesk/metadata/item.py
```diff
@@ -654,14 +663,8 @@ def remove_metadata_for_publish(item):
...
 
-    def filter_internal_attachment(attachment):
-        if attachment.get('attachment'): # retrieve object reference
-            attachment = superdesk.get_resource_service('attachments').find_one(req=None, _id=attachment['attachment'])
-        return not attachment.get('internal')
-
-    # Remove attachments that are internal
     if len(item.get('attachments', [])) > 0:
-        item['attachments'] = filter(filter_internal_attachment, item['attachments'])
+        item['attachments'] = list(filter(is_attachment_public, item['attachments']))
```
jerome-poisson: the pythonic way is to use list comprehension here:
```python
[attachment for attachment in item['attachments'] if is_attachment_public(attachment)]
```

superdesk/publish/formatters/ninjs_formatter.py
pablopunk: should I change it anyway then?

petrjasek: yes pls

pablopunk: is this ok?

```
'href': app.media.url_for_media(attachment['media']),
```

PR #1706, opened 2019-10-28
Use field label instead of ID in content filters

tomaskikutis:
SDESK-4538

### DIFF
--- a/apps/content_filters/filter_condition/filter_condition_parameters.py
+++ b/apps/content_filters/filter_condition/filter_condition_parameters.py
@@ -9,6 +9,7 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 import json
+from flask_babel import _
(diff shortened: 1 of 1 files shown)


tomaskikutis:
@petrjasek can you point me to an example on how to translate the labels?

petrjasek:
```py
from flask_babel import _
...
'label': _('ANPA Category'),
```

tomaskikutis:
Thanks. And how do I verify that it works? Will the same translations be shared between client/server?

petrjasek:
I don't think we have it working now e2e

tomaskikutis:
what do you mean by e2e? That sharing translations between client/server doesn't work, or that translating itself doesn't work? :smile:

petrjasek:
I mean sharing translations

PR #1707, opened 2019-10-29
Integrate OAuth with production API

ride90:
There is no fireq config for prod api (only for belga).
I will add it later to run pytest during ci and run prod api server.
UPD:
- PR for fireq https://github.com/superdesk/fireq/pull/51
- Fake PR to show how actually fireq works https://github.com/superdesk/superdesk-belga/pull/89


prod_api/app/settings.py
```diff
@@ -54,3 +54,9 @@
 
 # response in json
 XML = False
+
+# authorisation server
+AUTH_SERVER_SHARED_SECRET = env('AUTH_SERVER_SHARED_SECRET', '')
+
+# authentication
+PRODAPI_AUTH_ENABLED = env('PRODAPI_AUTH_ENABLED', True)
```
petrjasek: I guess it should use `strtobool` or something if you want to use env variable

apps/contacts/resource.py
```diff
@@ -44,7 +45,8 @@ class ContactsResource(Resource):
         },
         'last_name': {
             'type': 'string',
-            'required': False
+            'required': False,
+            'mapping': not_analyzed
```
petrjasek: not sure this is correct, why not analyzed?

ride90: changed to `{'type': 'string'}`

docs/production_api.rst
```diff
@@ -0,0 +1,225 @@
...
+
+.. _nose-exclude: https://pypi.org/project/nose-exclude/
+
+To run tests for production API, execute ``pytest`` command from ``prod_api`` folder.
+
+All fixtures for production API tests are defined in the ``conftest.py`` file.
+Tests for authentication and authorization are in ``test_auth.py`` file and they are more like e2e.
+To test an entire auth process close to real interaction between client, auth server and production API,
+2 flask apps are required respectively.
+**To avoid spinning 2 flask servers (superdesk and prod api) in separate processes and send real requests via local network to test things,
```
jerome-poisson: small details, but you can use `.. note::` instead of putting many sentences in bold, this would be more readable.

jerome-poisson:
https://github.com/superdesk/superdesk-core/pull/1707/files#diff-c6795de3b403789a774c297c1064710aL4  ==> it would be nice to add a link to production API doc here

superdesk/auth_server/models.py
```diff
@@ -52,7 +53,11 @@ def get_allowed_scope(self, scope):
 
 def query_client(client_id):
     clients_service = superdesk.get_resource_service('auth_server_clients')
-    client_data = clients_service.find_one(req=None, _id=ObjectId(client_id))
+    try:
+        client_data = clients_service.find_one(req=None, _id=ObjectId(client_id))
+    except InvalidId:
```
jerome-poisson: Why is the error silently ignored? IMHO, at least a `logger.error` should be done here.

ride90: Log message was added.

docs/production_api.rst
ride90: done

ride90:
@jerome-poisson link in docs from auth server to prod api was added.

docs/production_api.rst
```diff
@@ -0,0 +1,228 @@
...
+    | path: **/prodapi/v1/assets/MEDIA_ID.jpg**
+    | example: http://hostname/prodapi/v1/assets/5d22f47e5589a98f90775752.jpg
+
+
+Authentication
+--------------
+
+Production API implements JWT token authentication.
+Third-party apps must retrieve token using AuthServer_ and provide it with every request.
+
```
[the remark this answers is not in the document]
jerome-poisson: something like:
```shell
export PRODAPI=http://127.0.0.1:5500/prodapi/v1
export TOKEN=blabla.bla
export FILTER='?source={"query":{"filtered":{"filter":{"terms":{"type":["composite"]}}}}}'
curl -g -i $PRODAPI/items$FILTER -H "Authorization: Bearer $TOKEN"
```

prod_api/tests/test_auth.py
```diff
@@ -0,0 +1,336 @@
...
+        for resource in RESOURCES:
+            # we send a requests with 'empty scopes' token
+            resp = prodapi_client.get(
+                url_for('{}|resource'.format(resource)),
+                headers={
+                    'Authorization': 'Bearer {}'.format(access_token_no_scopes)
+                }
+            )
+            resp_data = json.loads(resp.data.decode('utf-8'))
+            # we get a 401 response
```
jerome-poisson: The sequence diagram at https://superdesk.readthedocs.io/en/latest/auth_server.html#overview says that a `403` must be returned in case of invalid scope (`401` is for invalid credentials), so this must be either changed here or in the diagram.

prod_api/tests/test_auth.py
```diff
@@ -0,0 +1,336 @@
...
+            assert resp.status_code == 401
+            assert resp_data == {
+                '_status': 'ERR',
+                '_error': {
+                    'code': 401,
+                    'message': 'Please provide proper credentials'
```
jerome-poisson: would not it better to have a message explicitly saying that the issue is with scope and not credentials? That would help admin to set the right values.

prod_api/tests/test_auth.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # we send a request with 'DESKS_READ scope' token to items endpoint
+        resp = prodapi_client.get(
+            url_for('archive|resource'),
+            headers={
+                'Authorization': 'Bearer {}'.format(access_token)
+            }
+        )
+        resp_data = json.loads(resp.data.decode('utf-8'))
+        # we get a 401 response
```
jerome-poisson: same remark as above (either change to `403` or change the sequence diagram of auth server)

jerome-poisson:
I have an internal error when I do this:
```shell
export PRODAPI=http://127.0.0.1:5500/prodapi/v1
curl -g -i $PRODAPI/items -H "Authorization: Bearer BAD_TOKEN"
```

ride90:
@jerome-poisson I've provided updates according to your comments.

PR #1711, opened 2019-11-04
save new keywords when publishing item

petrjasek:
SDBELGA-212

### DIFF
--- a/superdesk/vocabularies/__init__.py
+++ b/superdesk/vocabularies/__init__.py
@@ -10,8 +10,10 @@
 
 import superdesk
 
+from superdesk.signals import item_published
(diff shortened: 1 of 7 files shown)


superdesk/vocabularies/__init__.py
```diff
@@ -24,3 +26,5 @@ def init_app(app):
         'category': 'cvs',
         'label': 'Prefered CV items',
     })
+
+    item_published.connect(add_missing_keywords)
```
MarkLark86: @petrjasek Can we please have a system setting for this (we want our keywords vocabulary to be managed by admins).

PR #1715, opened 2019-11-06
Fdl 548 - modified date when updating a published article

mehdignu:


petrjasek:
how is it different from `versioncreated` ?

mehdignu:
the article's modified date should be changed according to the type of change:
- if grammar correction the modified date should not be updated.
- if republishing modified date should be updated.

petrjasek:
what is republishing?

mehdignu:
change in the content of the article

petrjasek:
why not use correction or update for these?

mehdignu:
We have two versions for updating a published article.
1) only some grammar corrections --> no republishing/ every time stamp stays the same (=modified date will not update)
2) content adjustments or changes --> republishing / time stamp for modified date changes because of the new content inside the article also in FE there should be the modified date displayed*

mehdignu:
We have two versions for updating a published article.
1) only some grammar corrections --> no republishing/ every timestamp stays the same (modified date will not be updated) no changes for date!
2) content adjustments or changes --> republishing / time stamp for modified date changes because of the new content inside the article --> also in FE there should be the modified date displayed

PR #1717, opened 2019-11-06
[METADATA] Fixed data type for custom date fields

jerome-poisson:
This patch fixes types for fields in extra (which is used for custom fields) when a date
is expected. So far, custom date fields were stored as plain text, which is causing
trouble when anything related to date is needed (like a comparison).

This patch also provides test + a data migration script to fix types of items already in
database. The data migration script provides the downgrade part, to avoid problems if
anything is going wrong.

fixes SDESK-4795

### DIFF
--- a/superdesk/data_updates/00020_20191106-180807_archive.py
+++ b/superdesk/data_updates/00020_20191106-180807_archive.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8; -*-
+# This file is part of Superdesk.
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+#
+# Author  : Jérôme
+# Creation: 2019-11-06 18:08
+
+from superdesk.commands.data_updates import DataUpdate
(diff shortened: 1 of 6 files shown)


superdesk/data_updates/00020_20191106-180807_archive.py
```diff
@@ -0,0 +1,47 @@
...
+from apps.archive.common import transtype_metadata
+from datetime import datetime
+
+
+class DataUpdate(DataUpdate):
+
+    resource = 'archive'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        for resource in ['archive', 'archive_autosave', 'published']:
```
petrjasek: I think it would be good to check if there is any extra date field defined on instance. I think usually there won't be any but there might be lots of items with extra data like on stt where we migrate their archives and this upgrade would take there forever

jerome-poisson: :+1:

jerome-poisson: :+1:

PR #1722, opened 2019-11-12
check ctime when detecting if file is old for ingest

petrjasek:
it was using only mtime for checking if file is old but
mtime is not updated when file is moved or pushed via ftp
so it would never try to ingest such file again.

now it will use both mtime and ctime to check how old file is.

SDANSA-349


superdesk/io/feeding_services/file_service.py
```diff
@@ -167,5 +166,16 @@ def is_old_content(self, last_updated):
         """
         return last_updated < utcnow() - timedelta(minutes=10)
 
+    def get_last_updated(self, file_path):
```
ride90: I would put it superdesk.utils

PR #1723, opened 2019-11-12
fix date parsing in picture iptc ingest

petrjasek:


superdesk/io/commands/update_ingest.py
```diff
@@ -616,7 +616,10 @@ def is_new_version(item, old_item):
             try:
                 return int(item[field], 10) > int(old_item[field], 10)
             except (ValueError, TypeError):
-                return item[field] > old_item[field]
+                try:
+                    return item[field] > old_item[field]
+                except TypeError:
+                    return True  # no idea
```
jerome-poisson: what this `# no idea` means?

jerome-poisson: would it be possible to have a comment more epxlicit? And would not it be better to log the error?

PR #1724, opened 2019-11-13
[ARCHIVE][FETCHING] add destination desk and stage ids to macros kwargs

jerome-poisson:
Macros had no way to know where an item is fetched or routed, this
patches fixes this by specifying `dest_desk_id` and `dest_stage_id` as
macros keyword arguments.

needed for SDBELGA-142 and SDBELGA-143

### DIFF
--- a/superdesk/internal_destinations.py
+++ b/superdesk/internal_destinations.py
@@ -73,7 +73,11 @@ def handle_item_published(sender, item, **extra):
         if dest.get('macro'):
             macro = macros_service.get_macro_by_name(dest['macro'])
             try:
-                macro['callback'](new_item)
+                macro['callback'](
+                    new_item,
+                    dest_desk_id=dest.get('desk'),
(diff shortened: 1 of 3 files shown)


jerome-poisson:
actually the same thing is needed for internal destination, so I'll rework this PR instead of make a new one, adding the `WIP` label now

superdesk/internal_destinations.py
```diff
@@ -73,7 +73,7 @@ def handle_item_published(sender, item, **extra):
         if dest.get('macro'):
             macro = macros_service.get_macro_by_name(dest['macro'])
             try:
-                macro['callback'](new_item)
+                macro['callback'](new_item, dest_desk_id=dest.get('desk'))
```
petrjasek: you could also add stage to be consistent

jerome-poisson: oh yes good catch, I've just overlooked it, adding it now.

PR #1725, opened 2019-11-14
[SDESK-4814] fix the feed query, request all versions, allow point in…

marwoodandrew:
… time recovery


MarkLark86:

PR #1727, opened 2019-11-15
[SDESK-4766] Add 'contact_type' to Contacts resource

MarkLark86:
* Allow linking vocabulary field to another vocabulary item (via link_vocab and link_field attributes in the schema)

### DIFF
--- a/apps/contacts/service.py
+++ b/apps/contacts/service.py
@@ -83,7 +83,7 @@ class ContactsService(Service):
         if not contact_type or not contact_type.get('assignable'):
             return
 
-        if not len(contact.get('contact_email') or []):
+        if not contact.get('contact_email'):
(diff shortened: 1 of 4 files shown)


MarkLark86:

apps/contacts/service.py
```diff
@@ -50,6 +63,33 @@ def on_deleted(self, doc):
...
+        contact_type = next((
+            item
+            for item in (types.get('items') or [])
+            if item.get('qcode') == contact.get('contact_type')
+        ), None)
+
+        if not contact_type or not contact_type.get('assignable'):
+            return
+
+        if not len(contact.get('contact_email') or []):
```
petrjasek: I guess `if not contact.get('contact_email'):` would work the same

PR #1729, opened 2019-11-19
Add schema_field to Terms of Use

pablopunk:
SDESK-4815

Needed for https://github.com/superdesk/superdesk-client-core/pull/3220


pablopunk:
@petrjasek this issue is on 1.32 btw, I'll merge it to develop but take that in mind

PR #1730, opened 2019-11-20
[ERRORS]Avoid RuntimeError when SuperdeskValidationError is thrown ou…

jerome-poisson:
…tside of request context

SuperdeskValidationError was assuming that the exception was raised in a
Flask request context, but this is not necessary the case (e.g. it can
be raised with a CLI command), and this result in a RuntimeError masking
the initial one.

This patch avoid it by catching the RuntimeError and showing a simple
warning, thus the initial SuperdeskValidationError will still be raised
normally.

SDFID-589

### DIFF
--- a/superdesk/errors.py
+++ b/superdesk/errors.py
@@ -733,20 +733,26 @@ class SuperdeskValidationError(HTTPException):
         Exception.__init__(self)
         self.errors = errors
         self.fields = fields
-        self.response = send_response(
-            None, (
-                {
-                    app.config['STATUS']: app.config['STATUS_ERR'],
-                    app.config['ISSUES']: {
-                        'validator exception': str([self.errors]),  # BC
-                        'fields': self.fields,
+        try:
+            self.response = send_response(
+                None, (
+                    {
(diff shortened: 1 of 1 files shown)


superdesk/errors.py
```diff
@@ -733,20 +733,26 @@ def __init__(self, errors, fields, message=None):
...
+                    None,
+                    400,
+                )
             )
-        )
+        except RuntimeError as e:
+            # the exception is run outside of request context
+            # it may be the case with CLI commands
+            # we log the error to not loose the initial error
+            logger.warning(e)
```
petrjasek: given that it's an error - why not use `logger.error()` ?

jerome-poisson: because in the case of CLI command, it's not an error. There error is actually in `SuperdeskValidationError` assuming that we are in an HTTP Request. I was initially wanting to pass it and not log anything, but a warning is still useful, as something may be wrong. So IMHO a warning is adapted here, not an error.

petrjasek: ok makes sense

PR #1732, opened 2019-11-22
Validate special characters. SDCP-83

GyanP:
Required for https://github.com/superdesk/superdesk-client-core/pull/3217

### DIFF
--- a/apps/client_config.py
+++ b/apps/client_config.py
@@ -34,4 +34,5 @@ def init_app(app):
         'publish_content_expiry_minutes': app.config.get('PUBLISHED_CONTENT_EXPIRY_MINUTES', 0),
         'high_priority_queue_enabled': app.config.get('HIGH_PRIORITY_QUEUE_ENABLED'),
         'default_language': app.config.get('DEFAULT_LANGUAGE'),
+        'disallowed_characters': app.config.get('DISSALLOWED_CHARACTERS'),
     })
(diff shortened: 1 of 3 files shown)


apps/validate/validate.py
```diff
@@ -149,6 +150,15 @@ def _validate_company_codes(self, *args):
         """Ignore company codes."""
         pass
 
+    def _validate_validate_characters(self, validate, field, value):
```
petrjasek: can you add some test to see it in action? how does it know what field to check?

GyanP: sure, I will try to add some.
> how does it know what field to check?

In the content-profile added an option to enable and disable the character validation as we do for `required` validation.

apps/client_config.py
```diff
@@ -34,4 +34,5 @@ def init_app(app):
         'publish_content_expiry_minutes': app.config.get('PUBLISHED_CONTENT_EXPIRY_MINUTES', 0),
         'high_priority_queue_enabled': app.config.get('HIGH_PRIORITY_QUEUE_ENABLED'),
         'default_language': app.config.get('DEFAULT_LANGUAGE'),
+        'disallowed_characters': app.config.get('DISSALLOWED_CHARACTERS'),
```
petrjasek: there is spelling mistake in the config name `s/DISSALLOWED/DISALLOWED/`

PR #1734, opened 2019-11-27
Remove 'marked_for_user' when article is killed. [SDBELGA-205]

GyanP:


apps/publish/content/kill.py
```diff
@@ -184,3 +189,28 @@ def apply_kill_override(self, item, updates):
...
+            return
+
+        updates = {'marked_for_user': None}
+        published_service = get_resource_service(PUBLISHED)
+
+        published_items = list(published_service.get_from_mongo(req=None, lookup={'item_id': item_id}))
+        if not published_items:
+            return
+
+        for item in published_items:
```
petrjasek: you could use [update_published_items](https://github.com/superdesk/superdesk-core/blob/develop/apps/publish/published_item.py#L33)

GyanP: Actually, I tried this one first, but it never returns any data, maybe those articles not present in the elastic. I tried with elastic versions 1.7 and 2.4. Not sure whats's wrong. Fetching directly from mongo does work well so I decided to do the way it is currently.

PR #1738, opened 2019-12-04
elastic7 support

petrjasek:


docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+            - "6379:6379"
+
+    mongo:
+        image: mongo:3.4
+        ports:
+            - "27017:27017"
+        tmpfs:
+            - /data/db
+
+    rabbitmq:
```
ride90: Why do we need a rabbitmq ?

petrjasek: we don't, I just wanted to run tests in environment close to production

petrjasek: but it's probably not used anyhow due to eager celery, so will remove it

superdesk/metadata/item.py
```diff
@@ -231,11 +231,13 @@
         'type': 'string',
         'mapping': {
             'type': 'string',
+            'fielddata': True,
```
ride90: Why do we need `fielddata` here? For sorting?

superdesk/publish/publish_content.py
```diff
@@ -157,7 +157,11 @@ def transmit_item(queue_item_id, is_async=False):
         logger.info('Transmitting queue item {}'.format(log_msg))
 
         destination = queue_item['destination']
-        transmitter = superdesk.publish.registered_transmitters[destination.get('delivery_type')]
+        try:
+            transmitter = superdesk.publish.registered_transmitters[destination.get('delivery_type')]
+        except KeyError:
+            print(destination['delivery_type'] not in superdesk.publish.registered_transmitters)
```
ride90: Forgot to remove a print statement.

ride90: I think having a `logger.exception/error` might be useful here.

petrjasek:
it's changed [there in this PR](https://github.com/superdesk/superdesk-core/pull/1738/files#diff-5d4eecebe8590ab62714759eb3959404R52)

PR #1739, opened 2019-12-04
Remove translations when article is spiked. [SDBELGA-236]

GyanP:
1. If the translated item is spiked remove link to original (translated_from) from the translated item and also remove translated item from the list of translations in the original item.

2. If the translated item is spiked remove link to original (translated_from) from the translated item and if the original item is in published state remove translated item from the list of translations of all the versions of the original item.

### DIFF
--- a/apps/archive/archive_spike.py
+++ b/apps/archive/archive_spike.py
@@ -16,7 +16,7 @@ from flask import current_app as app
 import superdesk
 from superdesk import get_resource_service, config
 from superdesk.errors import SuperdeskApiError, InvalidStateTransitionError
-from superdesk.metadata.item import ITEM_STATE, CONTENT_TYPE, ITEM_TYPE, GUID_FIELD, GUID_TAG
+from superdesk.metadata.item import ITEM_STATE, CONTENT_TYPE, ITEM_TYPE, GUID_FIELD, GUID_TAG, PUBLISH_STATES
 from superdesk.notification import push_notification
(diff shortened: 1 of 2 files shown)


petrjasek:
pls add some tests

apps/archive/archive_spike.py
```diff
@@ -211,6 +237,22 @@ def on_updated(self, updates, original):
             # send notification so that marked for me list can be updated
             get_resource_service('archive').handle_mark_user_notifications(updates, original, False)
 
+    def _update_translations(self, service, article, id_to_remove, article_id):
```
ride90: I would call it `def _remove_translation` imo it's more clear what it does.

apps/archive/archive_spike.py
```diff
@@ -162,6 +164,30 @@ def update(self, id, updates, original):
             updates['previous_marked_user'] = original['marked_for_user']
             updates['marked_for_user'] = None
 
+        if original.get('translation_id') and original.get('translated_from'):
+            # remove translations info from the translated item on spike
+            updates['translated_from'] = None
+            updates['translation_id'] = None
+
+            id_to_remove = original.get('_id')
```
ride90: I might be wrong, but I think it is `guid`.

GyanP: _id field goes to `translations`, so it to remove should be _id as well

PR #1740, opened 2019-12-05
Don't store replaced associated item's metadata in the newly added as associated item. [SDESK-4764]

GyanP:
https://dev.sourcefabric.org/browse/SDESK-4764

### DIFF
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -154,6 +154,32 @@ def flush_renditions(updates, original):
                         updates[ASSOCIATIONS][key]['renditions'][old_rendition] = None
 
 
+def flush_previous_item_keys_from_associations(updates, original):
+    """Remove original association items keys from updated association item
+        if the association key is same and associated items _id is diffrent.
+
+    for ex:
+        1. Add an image to a articles body and save the article where let's say key is(association.edior_0)
+        2. Remove the associated image don't save artilce this time
+        3. And add a new media item wehere key is still (association.edior_0)
+        4. save the article and notice that new associated item contains all those keys from previous associated
+            items that were not present in this one.
+        Which is wrong, this function identifies those keys and sets them to None
+
+    :param dict updates: updates for the document
+    :param original: original is document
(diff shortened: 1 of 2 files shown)


GyanP:
@ride90 @petrjasek is it possible to completely remove the extra keys instead of setting them to None on patch request?

ride90:
Hi @GyanP I think delete is fine.

petrjasek:
I think this won't work for extra:
```py
original['associations']['foo']['extra'] = {'foo': 1}
updates['associations']['foo']['extra'] = {'bar': 1}
```
will produce `{'foo': 1', 'bar': 1}`

GyanP:
> I think this won't work for extra:

good catch, I think this is getting a bit complicated. Isn't there a way around so that we can avoid setting all those keys to None in updates? Simply set `original['associations'][key] = None` and merge updates and original?

petrjasek:
there is a way - set [MERGE_NESTED_DOCUMENTS](https://docs.python-eve.org/en/stable/config.html) to `False`. problem is it's only supported in eve 0.8 so it would require an update. which we should do anyhow..

ride90:
@petrjasek @GyanP since it may take us some time to update to Eve 0.8 (http://docs.python-eve.org/en/stable/changelog.html#id85), we should somehow safety prevent the issue described in the ticket (don't overwrite nested docs), I was able to repeat it quite often playing with Belga images, it results in a wrong data in DB and in output.
@GyanP it's good that only one method handles it, so after an update, it will be easy to erase this logic. Just add a comment to that method, so we know how and why we should erase it. 
It would be good to have a ticket for Eve update with reference to this pr/ticket there.

petrjasek:
maybe we could also try to backport that config into older eve - imo it's not a big change - to avoid all those workarounds, and upgrade eve later

ride90:
@GyanP @petrjasek I've created a ticket to update an eve https://dev.sourcefabric.org/browse/SDESK-4968

PR #1743, opened 2019-12-09
 Add notification for content_templates. [SDESK-4828]

GyanP:
Required for https://github.com/superdesk/superdesk-client-core/pull/3240

### DIFF
--- a/apps/templates/content_templates.py
+++ b/apps/templates/content_templates.py
@@ -31,6 +31,7 @@ from superdesk.celery_task_utils import get_lock_id
 from croniter import croniter
 from datetime import datetime
 from flask_babel import _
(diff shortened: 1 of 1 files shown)


apps/templates/content_templates.py
```diff
@@ -106,6 +107,18 @@ def get_next_run(schedule, now_utc=None):
...
 
+def push_template_notification(docs, event='template:update'):
+    user = get_user()
+    template_desks = set()
+
+    for doc in docs:
+        if doc.get('template_desks'):
+            template_desks.update([str(template) for template in doc.get('template_desks')])
+
+    print(event, template_desks)
```
ride90: ⚠️ I think you forgot to remove `print`

PR #1744, opened 2019-12-10
[SDESK-4766] Data update for ContentType/CoverageProvider CVs

MarkLark86:


MarkLark86:
@petrjasek I have created another PR targeting 1.30 and 1.31. This data update is for 1.32+ (as there were 2 newer data updates that are not in 1.30 nor 1.31)

PR #1747, opened 2019-12-10
implement archive autocomplete resource

petrjasek:
SDCP-81

### DIFF
--- a/apps/archive/autocomplete.py
+++ b/apps/archive/autocomplete.py
@@ -1,12 +1,19 @@
 
 import locale
+import warnings
 import superdesk
 
 from flask import current_app as app, request
+from flask_babel import _
 from datetime import timedelta
 from superdesk.utils import ListCursor
 from superdesk.utc import utcnow
 from superdesk.errors import SuperdeskApiError
+
+
+SETTING_ENABLED = 'ARCHIVE_AUTOCOMPLETE'
+SETTING_DAYS = 'ARCHIVE_AUTOCOMPLETE_DAYS'
+SETTING_HOURS = 'ARCHIVE_AUTOCOMPLETE_HOURS'
 
 
 class AutocompleteResource(superdesk.Resource):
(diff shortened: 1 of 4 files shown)


apps/archive/autocomplete.py
```diff
@@ -0,0 +1,53 @@
...
+
+    allowed_fields = {
+        'slugline',
+    }
+
+    def get(self, req, lookup):
+        field = request.args.get('field', 'slugline')
+        language = request.args.get('language', app.config.get('DEFAULT_LANGUAGE', 'en'))
+
+        if not app.config.get('ARCHIVE_AUTOCOMPLETE'):
```
MarkLark86: It might be good to also check `ARCHIVE_AUTOCOMPLETE_DAYS == 0 and ARCHIVE_AUTOCOMPLETE_HOURS == 0`, as those conditions equates to this endpoint returning empty values all the time.

Also text translation for API errors are missing, i.e. `from flask_babel import _`

apps/archive/autocomplete.py
```diff
@@ -0,0 +1,53 @@
...
+
+class AutocompleteResource(superdesk.Resource):
+    item_methods = []
+    resource_methods = ['GET']
+
+
+class AutocompleteService(superdesk.Service):
+
+    allowed_fields = {
+        'slugline',
```
MarkLark86: Being as this endpoint is quite generic, what if we were to allow any non-system level fields and string only fields.

For example:
slugline
headline
abstract
ednote
description_text
archive_description
alt_text

superdesk/default_settings.py
```diff
@@ -756,3 +757,24 @@ def local_to_utc_hour(hour):
...
+
+#: Enable archive autocomplete API
+#:
+#: .. versionadded:: 2.0
+#:
+#: It will return suggestions for field from archive api,
+#: only from published items for time defined via
+#: :data:`ARCHIVE_AUTOCOMPLETE_DAYS` and :data:`ARCHIVE_AUTOCOMPLETE_HOURS`.
+#:
+ARCHIVE_AUTOCOMPLETE = False
```
MarkLark86: Should anything in `default_settings.py` use `env` utility method when defining settings?

petrjasek: I don't think so. It make sense for eg db connection which would be different in different environments where you deploy an instance (testing/production), but a feature you either use and have it on everywhere or you don't use it.

apps/archive/autocomplete.py
petrjasek: I think for those non single line fields it could be a lot of data, also not sure how the autocomplete would work there in the editor. but I'm open to add more if needed.

PR #1748, opened 2019-12-10
allow creating update before publishing previous one

petrjasek:
SDBELGA-229

### DIFF
--- a/apps/archive/archive_rewrite.py
+++ b/apps/archive/archive_rewrite.py
@@ -113,6 +113,14 @@ class ArchiveRewriteService(Service):
         if not is_workflow_state_transition_valid('rewrite', original[ITEM_STATE]):
             raise InvalidStateTransitionError()
 
+        if (
+            original.get('rewrite_of')
+            and not (original.get(ITEM_STATE) in PUBLISH_STATES)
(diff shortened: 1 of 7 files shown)


MarkLark86:
This is a hard requirement for our installation. We do not allow our users to create a rewrite of a rewrite without the first being published.

Can we please allow this to be configurable, with behave tests for both scenarios to ensure this functionality will continue to work for both Belga and AAP?

petrjasek:
ok, it's a bit confusing because you allow rewrite of the original item to be created, but not rewrite of a rewrite. is that how it's supposed to work?

MarkLark86:
> ok, it's a bit confusing because you allow rewrite of the original item to be created, but not rewrite of a rewrite. is that how it's supposed to work?

In our installation, you can create a rewrite of a rewrite, as long as the rewrite_of is published. So at any one time you can only have 2 linked items that are not yet published.

i.e.
1. Create draft item A
2. Update A to create B
3. Update B to create C -- fails as item A is not yet published
4. Publish item A
5. Update B to create C -- succeeds as item A is now published
6. Update C to create D -- fails as item B is not yet published
7. Publish item B
8. Update C to create D -- succeeds as item B is now published

PR #1749, opened 2019-12-11
Fix: POI of associated images is not updated in NIJS on correction. [SDESK-4871]

GyanP:
Handle media updates on correction so that multiple crops can be generated for associated media as well.

### DIFF
--- a/features/content_publish.feature
+++ b/features/content_publish.feature
@@ -3180,6 +3180,17 @@ Feature: Content Publishing
         """
         [{"name": "Sports", "members":[{"user":"#CONTEXT_USER_ID#"}]}]
         """
+      And "vocabularies"
+      """
+      [{
+        "_id": "crop_sizes",
+        "unique_field": "name",
+        "items": [
+          {"is_active": true, "name": "600x800", "width": 800, "height": 600},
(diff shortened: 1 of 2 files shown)


features/content_publish.feature
```diff
@@ -3544,7 +3555,7 @@ Feature: Content Publishing
                             "media": "600x800_new",
                             "mimetype": "image/jpeg"
                         },
-                        "1280x720": null
+                        "1280x720": "__none__"
```
petrjasek: shouldn't it be actually populated now?

GyanP: Not actually, because we do flush the renditions.

petrjasek: ok, so how does it test the issue?

GyanP: When we do send a correction with updated poi the updated custom crops weren't generated earlier. But now we do generate the updated custom crops and if there are any corps that aren't defined in vocabulary if will throw an error. That's why this test was failing after current changes. So, it covers both the scenarios the one it was originally written for and also the current one. And that's why I thought there is no need to write a separate test for the current fix.

petrjasek: I see, ok then

PR #1755, opened 2019-12-17
Fix back-end validation for required custom fields. [SDESK-4886]

GyanP:
Custom fields might contain `string` (Belga coverage) as well as `datetime` (due date) fields.

### DIFF
--- a/apps/validate/validate.py
+++ b/apps/validate/validate.py
@@ -96,6 +96,10 @@ class SchemaValidator(Validator):
         if value and not isinstance(value, str):
             self._error(field, STRING_FIELD)
 
+    def _validate_type_any(self, field, value):
+        """Allow type any, ex: for CV of type 'custom'."""
+        pass
+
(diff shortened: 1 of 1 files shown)


apps/validate/validate.py
```diff
@@ -207,7 +207,8 @@ def create(self, docs, fields=False, **kwargs):
 
     def _get_profile_schema(self, schema, doc):
         doc['validate'].setdefault('extra', {})  # make sure extra is there so it will validate its fields
-        extra_field_types = {'text': 'string', 'embed': 'dict', 'date': 'date', 'urls': 'list'}
+        extra_field_types = {'text': 'string', 'embed': 'dict', 'date': 'date',
+                             'urls': 'list', 'custom': ['datetime', 'string']}
```
petrjasek: not sure this would work in general, I think it would be better to introduce new type `any` without type validation

PR #1758, opened 2019-12-17
fix error when moving to archived and item is there already

petrjasek:
it shouldn't happen, but handle it gracefuly when it does.


MarkLark86:
@petrjasek How is this error replicated? When an item is moved to the `archived` collection, the item is meant to be removed from both `published` and `archive` collections.

`347: published_service.delete_by_article_id(item_id)`
and
`350: archive_service.delete_by_article_ids([item_id])`

petrjasek:
seems like there were multiple processes running for expiry, trying to handle that now via lock touch

apps/archive/commands.py
```diff
@@ -105,6 +106,10 @@ def _remove_expired_items(self, expiry_datetime):
                 logger.info('{} No items found to expire.'.format(self.log_msg))
                 return
 
+            if not touch(lock_name, expire=300):
```
ride90: It means that for every iteration in the current loop we will reset expiry for the lock on 300 seconds, right?

apps/archive/commands.py
```diff
@@ -334,8 +339,11 @@ def _move_to_archived(self, item, filter_conditions):
                 logger.info('{} Found {} published items for item: {}'.format(self.log_msg,
                                                                               len(published_items), item_id))
                 if moved_to_archived:
-                    archived_service.post(published_items)
-                    logger.info('{} Moved item to text archive for item {}.'.format(self.log_msg, item_id))
+                    try:
+                        archived_service.post(published_items)
+                        logger.info('{} Moved item to text archive for item {}.'.format(self.log_msg, item_id))
+                    except Conflict:
+                        logger.info('{} Moved item already to text archive for item {}.'.format(self.log_msg, item_id))
```
ride90: Not clear what does it mean, IMO, it's better to rephrase it.

apps/archive/commands.py
petrjasek: yep it would add 300 more every time, which might be actually a lot, will decrease it

apps/archive/commands.py
petrjasek: will remove catch after all

apps/archive/commands.py
petrjasek: actually it doesn't add it, but rather sets expiry to now + 300, so it should be a bit more

PR #1762, opened 2019-12-25
Add url in the search provider resource. [SDESK-4905]

GyanP:
required for https://github.com/superdesk/superdesk-client-core/pull/3266

### DIFF
--- a/apps/search_providers/resource.py
+++ b/apps/search_providers/resource.py
@@ -22,6 +22,7 @@ class SearchProviderResource(Resource):
         },
         'source': required_string,
         'name': {'type': 'string'},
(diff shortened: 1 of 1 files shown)


apps/search_providers/resource.py
```diff
@@ -22,6 +22,7 @@ class SearchProviderResource(Resource):
         },
         'source': required_string,
         'name': {'type': 'string'},
+        'url': {'type': 'string'},
```
petrjasek: shouldn't this be part of config dict like username/password?

ride90:
@GyanP what is the status of this pr?

PR #1765, opened 2020-01-07
avoid testing config when saving ingest provider config

petrjasek:
it was not working when celery and rest server were on different hosts

SDESK-4853

### DIFF
--- a/features/ingest_provider.feature
+++ b/features/ingest_provider.feature
@@ -50,7 +50,7 @@ Feature: Ingest Provider
         """
         {"feeding_service": "rss", "config": {"url": "http://localhost:1234"}}
         """
-        Then we get error 400
+        Then we get OK response
 
         When we post to "ingest_providers"
         """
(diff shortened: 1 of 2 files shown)


MarkLark86:
@petrjasek 
As per the original ticket, https://dev.sourcefabric.org/browse/SDESK-638, the reason behind checking the config is `As a user, I want to get a warning if I try to save a misconfigured ingest source`.

Most probably due to human error, if the ingest setting is incorrectly entered the administrator will not know until the first ingest iteration. If the FTP, API or File location is incorrect, this should be notified when saving the ingest source. In cases for Belga, can we add an attribute in the schema to not check the config for that specific ingest source?

i.e.
Backend:
Add the following attribute to the `ingest_provider_model` schema
`'validate_config': {'type': 'boolean', 'default': True}`

Frontend:
Add a ToggleBox in the ingest source form

Then when creating or updating an Ingest source, if `validate_config` is false, then don't check the config for errors, otherwise check the config as normal.

petrjasek:
re-implemented the check via:
https://github.com/superdesk/superdesk-core/pull/1768
https://github.com/superdesk/superdesk-client-core/pull/3272

PR #1766, opened 2020-01-07
SDESK-4928 Do not try to publish/unpublish not fetchable associated item.

ride90:
SDESK-4928

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -469,7 +469,7 @@ class BasePublishService(BaseService):
         else:
             return [], []
 
-    def _validate_associated_items(self, original_item, validation_errors=[]):
+    def _validate_associated_items(self, original_item, validation_errors=None):
         """Validates associated items.
 
         This function will ensure that the unpublished content validates and none of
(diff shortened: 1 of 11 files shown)


apps/publish/content/common.py
```diff
@@ -576,7 +576,9 @@ def _publish_associated_items(self, original, updates={}):
 
         associations = original.get(ASSOCIATIONS) or {}
         for associations_key, associated_item in associations.items():
-            if type(associated_item) == dict and associated_item.get(config.ID_FIELD):
+            if (type(associated_item) == dict
+                    and associated_item.get(config.ID_FIELD)
+                    and associated_item.get('_fetchable')):
```
petrjasek: imo would be good to use some default here so it won't change behaviour when `_fetchable` is missing:
```py
and associated_item.get('_fetchable', True) != False
```

superdesk/tests/steps.py
```diff
@@ -989,7 +993,9 @@ def when_upload_patch_dictionary(context):
     assert_ok(context.response)
 
 
-def upload_file(context, dest, filename, file_field, extra_data=None, method='post', user_headers=[]):
+def upload_file(context, dest, filename, file_field, extra_data=None, method='post', user_headers=None):
+    if user_headers is None:
+        user_headers = []
```
petrjasek: any idea if we can enable some flake8 check for that pattern to avoid it in future?

ride90: I found it only for pylint. There is a plugin for flake8 which may help with it. I will check it and if it works I'll create a new pr.

PR #1767, opened 2020-01-08
[SDBELGA-152] Implement OpenID Connect authentication

idsvn:
For installation and configuration instruction, please check our comment in the ticket.
There is some small changes compare to the implementation in Belga (superdesk/superdesk-belga#92)
- Use `flask-oidc-ex` instead of `flask-oidc`: Because the `flask-oidc` has no release since 2018, which contains a commit support load configuration from dict
- Change configure environment variable **name**, this is the completed list of ODIC env config:
`OIDC_ENABLE`, `OIDC_ISSUER`, `OIDC_SERVER_CLIENT`, `OIDC_SERVER_CLIENT_SECRET`, `OIDC_WEB_CLIENT`, `OIDC_BROWSER_REDIRECT_URI`

SDBELGA-152


apps/auth/oidc/auth.py
```diff
@@ -0,0 +1,73 @@
...
+            'email': g.oidc_token_info.get('email'),
+            'first_name': g.oidc_token_info.get('given_name'),
+            'last_name': g.oidc_token_info.get('family_name'),
+            'display_name': g.oidc_token_info.get('name'),
+            'user_type': g.oidc_token_info.get('user_type'),
+            'role': role.get('_id') if role else None,
+            'needs_activation': False,
+        }
+        if not user:
+            sync_data['password'] = get_hash(
```
idsvn: `auth.db` will raise error if password is None, but we wonder if we should care about the case when Keycloak is turning off after it has turned on?

apps/auth/oidc/auth.py
```diff
@@ -0,0 +1,73 @@
...
+                get_random_string(), current_app.config.get('BCRYPT_GENSALT_WORK_FACTOR', 12))
+            user_id = users_service.post([sync_data])[0]
+            user = auth_service.find_one(req=None, _id=user_id)
+        else:
+            users_service.put(user.get('_id'), sync_data)
```
idsvn: We will **override user info** with info from Keycloak, it also means changing info from user page will no longer work

apps/auth/oidc/auth.py
```diff
@@ -0,0 +1,73 @@
...
+        user = auth_service.find_one(req=None, username=g.oidc_token_info.get('username'))
+        role = get_resource_service('roles').find_one(req=None, name=g.oidc_token_info.get('role'))
+        sync_data = {
+            'username': g.oidc_token_info.get('username'),
+            'email': g.oidc_token_info.get('email'),
+            'first_name': g.oidc_token_info.get('given_name'),
+            'last_name': g.oidc_token_info.get('family_name'),
+            'display_name': g.oidc_token_info.get('name'),
+            'user_type': g.oidc_token_info.get('user_type'),
+            'role': role.get('_id') if role else None,
```
idsvn: How should we handle `role` and `user_type`? We are using [user attribute](https://user-images.githubusercontent.com/45254285/71950353-765b2280-3209-11ea-9063-e917c3fb2e3d.png), but as you can see, admin has to enter role and user_type manually.
We think that **Roles** in `Clients` can be used for `role` and ignore  fields [Roles Attribute](https://user-images.githubusercontent.com/45254285/71950603-760f5700-320a-11ea-9e8e-cf37295a7cc8.png), what do you think?

petrjasek: I think you can create those users as `user_type=user` and if needed it can be changed to administrator in superdesk.
you could probably check those roles from keycloak and look for matching role in superdesk if there is any, if multiple use first one. it would be also editable in superdesk later on

apps/auth/oidc/auth.py
petrjasek: could you set there empty string?

apps/auth/oidc/auth.py
petrjasek: it makes sense for some common data like email/username/first name/last name, but not sure about superdesk specific data like user role/user type, that should be app specific, shouldn't it?

idsvn: Make sense, we will update to sync only username, email, first/last name and display name.

apps/auth/oidc/auth.py
```diff
@@ -0,0 +1,83 @@
...
+        token = auth_header[1]
+        is_valid = self.oidc.validate_token(token, ['openid', 'email', 'profile'])
+        if not is_valid:
+            raise CredentialsAuthError(credentials)
+
+        auth_service = get_resource_service('auth_users')
+        users_service = get_resource_service('users')
+        username = g.oidc_token_info.get('username')
+        user = auth_service.find_one(req=None, username=username)
+        user_info = users_service.find_one(req=None, username=username) or {}
```
idsvn: Can we only use `users` service here or it has to be `auth_users` service for checking existing user and return value for `authenticate` method?

petrjasek: you should use `users` imo, `auth_users` is only for authentication via password which `users` won't return, but you don't need that, right?

ride90:
@idsvn what is the status of this pr?

idsvn:
> @idsvn what is the status of this pr?

Sorry for late replay, it's ready for review
We don't write test for this because it requires exchanging token with keycloak server (Mock?) which is handled internally by `flask_oidc_ex`.

ride90:
@idsvn please, resolve conflicts.

apps/auth/oidc/auth.py
```diff
@@ -54,16 +55,21 @@ def authenticate(self, credentials):
...
-            'first_name': g.oidc_token_info.get('given_name'),
-            'last_name': g.oidc_token_info.get('family_name'),
             'display_name': g.oidc_token_info.get('name'),
         }
+
+        # first name and last name is optional in Keycloak
+        if 'given_name' in g.oidc_token_info:
+            sync_data['first_name'] = g.oidc_token_info.get('given_name')
+        if 'family_name' in g.oidc_token_info:
+            sync_data['last_name'] = g.oidc_token_info.get('family_name')
```
idsvn, line 65: avoid raising error in https://github.com/superdesk/superdesk-core/blob/3e7ad403ea04b674009dbef04df7dda03f0aae22/superdesk/users/services.py#L111

idsvn:
@ride90 @petrjasek I have resolved all merge conflict. Summary changes:
- Change `OIDC_ENABLE` to `OIDC_ENABLED` (similar to `VIDEO_SERVER_ENABLED`)
- Ignore case when querying `role`
- Handle empty `first_name` and `last_name`

ride90:
Hi @idsvn, please, make all ci checks passing.

idsvn:
@ride90 Not sure why `fire:build` does not restart when reopen PR but Travis build has passed so I guess it's fine?
I added a simple warning for missing `SECRET_KEY`, feel free to update the message as you see fit.

apps/auth/oidc/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+import superdesk
+
+from .auth import OIDCAuthResource, OIDCAuthService
+
+logger = logging.getLogger(__name__)
+
+
+def init_app(app):
+    endpoint_name = 'auth_oidc'
+    if bool(app.config['OIDC_ENABLED']) and not bool(app.config['SECRET_KEY'] != ''):
```
ride90, line 22: why do we need `bool` here?
`app.config['OIDC_ENABLED']` is bool anyway and `!=` returns a bool

apps/auth/oidc/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+        logger.warn('SECRET_KEY is not set')
+
+    app.client_config['oidc_auth'] = bool(app.config['OIDC_ENABLED']) and bool(app.config['SECRET_KEY'] != '')
```
ride90, line 25: same here

superdesk/default_settings.py
```diff
@@ -907,4 +908,18 @@ def local_to_utc_hour(hour):
...
+#: .. versionadded:: 2.0
+#:
 DEFAULT_LANGUAGE = 'en'
+
+#: OIDC config
+#:
+#: .. versionadded:: 2.1
+#:
+OIDC_ENABLED = strtobool(env('OIDC_ENABLED', 'false'))
+OIDC_ISSUER = env('OIDC_ISSUER', 'http://localhost:8080/auth/realms/SUPERDESK_REALM')
```
ride90, line 921: I would rename it to `OIDC_ISSUER_URL`

apps/auth/oidc/auth.py
```diff
@@ -0,0 +1,88 @@
...
+        auth_header = request.headers.get('Authorization', '').split(' ', 1)
+        if auth_header[0] != 'Bearer' and len(auth_header) != 2:
+            raise CredentialsAuthError(credentials)
+        token = auth_header[1]
+        is_valid = self.oidc.validate_token(token, ['openid', 'email', 'profile'])
+        if not is_valid:
+            raise CredentialsAuthError(credentials)
+
+        users_service = get_resource_service('users')
+        username = g.oidc_token_info.get('username')
```
ride90, line 53: is `username` required here?

apps/auth/oidc/auth.py
```diff
@@ -0,0 +1,88 @@
...
+        user = users_service.find_one(req=None, username=username) or {}
+
+        sync_data = {
+            'username': username,
+            'email': g.oidc_token_info.get('email'),
```
ride90, line 58: same for email, is it required?

ride90: if email and username are required and we expect them to be in `oidc_token_info`, I would use `g.oidc_token_info['email/username']` instead of `get` to raise a KeyError. It will help to find and fix issue much faster unlike when it fails silently.

apps/auth/oidc/auth.py
```diff
@@ -0,0 +1,88 @@
...
+                    break
+            sync_data.update({
+                'password': '',
+                'user_type': 'user',
+                'role': user_role,
+                'needs_activation': False,
+            })
+            users_service.post([sync_data])
+        else:
+            users_service.patch(user[config.ID_FIELD], sync_data)
```
ride90, line 85: does it make sense to update a `role` if it was changed on the keycloak side?

superdesk/default_settings.py
```diff
@@ -907,4 +908,18 @@ def local_to_utc_hour(hour):
...
+OIDC_SERVER_CLIENT = env('OIDC_SERVER_CLIENT')
+OIDC_SERVER_CLIENT_SECRET = env('OIDC_SERVER_CLIENT_SECRET')
+OIDC_WEB_CLIENT = env('OIDC_WEB_CLIENT')
+OIDC_BROWSER_REDIRECT_URI = env('OIDC_BROWSER_REDIRECT_URI', 'http://localhost:9000')
```
ride90, line 925: I would rename it to `OIDC_BROWSER_REDIRECT_URL` to be more precise

ride90:
@idsvn it would be great to add docs how to configure keycloak into the repo itself. Here https://github.com/superdesk/superdesk-belga/blob/master/README.md or better to a separate file.

apps/auth/oidc/auth.py
idsvn: Yes, we need to check if user is exist and sync user data from `keycloak`, so we could use either `username` or `email` here to find user

apps/auth/oidc/auth.py
idsvn: How about fallback to superdesk user email? Keycloak `username` is required and unchangeable so there is no problem

apps/auth/oidc/auth.py
idsvn: I think it's easier to manage roles on superdesk side, as we also create role on superdesk.
Also superdesk only allow 1 role per user, but you could assign multiple roles on single user in Keycloak, which is easy to misuse.

apps/auth/oidc/auth.py
ride90: yes, let's raise an error

apps/auth/oidc/auth.py
ride90: yep, let's keep it as it is for now

PR #1770, opened 2020-01-10
implement editor utils text replace

petrjasek:
SDBELGA-201

### DIFF
--- a/superdesk/editor_utils.py
+++ b/superdesk/editor_utils.py
@@ -479,13 +479,14 @@ class Editor3Content(EditorContent):
         if not value:
             return
 
-        def create_entity(offset, length):
-            entity = {
-                'key': len(self.content_state['entityMap'].keys()),
-                'offset': offset,
-                'length': length,
+        def create_entity(type, data, mutability='MUTABLE'):
+            key = len(self.content_state['entityMap'].keys())
+            self.content_state['entityMap'][str(key)] = {
+                'type': type,
+                'mutability': mutability,
+                'data': data,
             }
-            return entity
+            return key
 
         if html:
             root = parse_html(value, 'html')
@@ -496,20 +497,15 @@ class Editor3Content(EditorContent):
                 except StopIteration:
                     block_type = None
                     depth = 0
-                    if elem.tag == 'figure' and i > 0 and root[i-1].text:
-                        m = re.search(r'<!-- EMBED START Image {id: "(.*)"}', str(root[i-1]).strip())
+                    if elem.tag == 'figure' and i > 0 and root[i - 1].text:
+                        m = re.search(r'<!-- EMBED START Image {id: "(.*)"}', str(root[i - 1]).strip())
                         if m and self.item.get('associations') and self.item['associations'].get(m.group(1)):
                             media = self.item['associations'][m.group(1)]
                             block = self.create_block('atomic', text=" ").data
-                            entity = create_entity(0, 1)
+                            entity_key = create_entity('MEDIA', {'media': media})
+                            block['entityRanges'] = [{'offset': 0, 'length': 1, 'key': entity_key}]
                             block['inlineStyleRanges'] = []
-                            block['entityRanges'] = [entity]
                             self.content_state['blocks'].append(block)
-                            self.content_state['entityMap'][str(entity['key'])] = {
-                                'type': 'MEDIA',
-                                'mutability': 'MUTABLE',
-                                'data': {'media': media},
-                            }
                     elif elem.tag in ('ul', 'ol'):
                         pass
                     elif elem.text and 'EMBED END' in elem.text:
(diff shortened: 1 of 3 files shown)


petrjasek:
I've simplified the api to work like:
```py
from superdesk.editor_utils import text_replace

text_replace(item, 'body_html', 'from', 'to')
```
it handles both editor3 and editor2 fields via converting editor2 content into editor3 format, so we would implement all required utils just once for editor3 content state

what do you think? @jerome-poisson @ride90

superdesk/editor_utils.py
```diff
@@ -392,7 +454,7 @@ class Editor3Content(EditorContent):
     editor_version = 3
     HTML_EXPORTER = DraftJSHTMLExporter
 
-    def __init__(self, item, field='body_html'):
+    def __init__(self, item, field='body_html', html=True):
```
jerome-poisson: we can guess content type from the suffix (body_**html**), so maybe we can detect with suffix by default with a `html=None` value, and use `True`/`False` only when we want to force or not HTML.

superdesk/editor_utils.py
```diff
@@ -401,29 +463,154 @@ def __init__(self, item, field='body_html'):
...
+                        if m and self.item.get('associations') and self.item['associations'].get(m.group(1)):
+                            media = self.item['associations'][m.group(1)]
+                            create_atomic_block('MEDIA', {'media': media})
+                        continue
+                    elif elem.tag in ('ul', 'ol'):
+                        pass
+                    elif elem.text and '<!-- EMBED' in str(elem):
+                        continue
+                    elif elem.tag == 'table':
+                        data = {'numCols': 0, 'numRows': 0, 'withHeader': False, 'cells': {}}
```
jerome-poisson: Initially I was thinking about creating specialised block classes with high level method, for instance there would be here a `TableBlock` with methods like `addRow` or `addCol`, that would avoid the manipulation of data inside `Editor3Content`. The goal is to let the dev modify internal data if necessary, but to have high level methods to make life easy as much as possible.

@petrjasek @ride90 what do you think?

superdesk/editor_utils.py
```diff
@@ -439,3 +626,20 @@ def prepend_block(self, block):
         else:
             index = 0
         self.blocks.insert(index, block)
+
+
+def replace_text(item, field, old, new, html=True):
```
jerome-poisson: here too, `html` can be guessed from `field` name

superdesk/editor_utils.py
```diff
@@ -439,3 +626,20 @@ def prepend_block(self, block):
...
+    editor = Editor3Content(item, field, html)
+    for block in editor.blocks:
+        block.replace_text(old, new)
+    editor.update_item(html)
+
+
+def filter_blocks(item, field, filter, html=True):
```
jerome-poisson: same comment for `html`

superdesk/editor_utils.py
```diff
@@ -401,29 +465,170 @@ def __init__(self, item, field='body_html'):
...
+    def text(self):
+        return '\n'.join([block.text for block in self.blocks])
+
     def get_next_entity_key(self):
         """Return a non existing key for entityMap"""
         return max((int(k) for k in self.content_state['entityMap'].keys()), default=-1) + 1
 
-    def update_item(self):
-        self.item[self.field] = self.html
+    def update_item(self, html=True):
```
jerome-poisson: why do we have to specify again HTML when we already did in `__init__`? We may want to convert a text field to HTML or vice versa, but in this case the `self.field` should change too, right?

superdesk/editor_utils.py
```diff
@@ -401,29 +465,170 @@ def __init__(self, item, field='body_html'):
         self.field = field
         data = item.get('fields_meta', {}).get(field, {})
```
jerome-poisson: I don't remember why exactly it was done this way, but I think that it should be a `setdefault` here instead of `get`, otherwise `fields_meta` won't be generated if it was not already present in the item (which can happen even for editor 3 on just ingested or unmodified articles)

superdesk/editor_utils.py
```diff
@@ -439,3 +644,20 @@ def prepend_block(self, block):
         else:
             index = 0
         self.blocks.insert(index, block)
+
+
+def replace_text(item, field, old, new, html=True):
+    editor = Editor3Content(item, field, html)
```
jerome-poisson: It's a good idea to manage both editor2 and editor3, but I think we should have a dedicated `Editor2Content` which may simply inheritate from `Editor3Content` and remove `fields_meta` after updating.
There is already an `EditorContent` factory, you just have to replace the `NotImplementedError` by the new `Editor2Content`: https://github.com/superdesk/superdesk-core/blob/39dfe04397b99f95d88cb5e3437d1e4c3e734784/superdesk/editor_utils.py#L254

superdesk/editor_utils.py
petrjasek: we could guess it for core fields, but not sure about extra fields so would rather make it explicit

superdesk/editor_utils.py
petrjasek: my bad, didn't have it in init at first, only added it later

superdesk/editor_utils.py
```diff
@@ -439,3 +615,70 @@ def prepend_block(self, block):
...
+            continue
+        start = 0
+        while True:
+            try:
+                index = block['text'].index(old, start)
+                start = index + len(old)
+                block['text'] = new.join([block['text'][:index], block['text'][start:]])
+                for range_field in (ENTITY_RANGES, INLINE_STYLE_RANGES):
+                    if block.get(range_field):
+                        ranges = []
```
ride90: we might lose all style ranges at this point.
Example:
```
item = {
    'body_html': '<h1>head</h1>\n<p>lorem <b>this is bold</b> and <b>bold</b> end</p>',
    'fields_meta': {}
}
replace_text(item, 'body_html', 'bold', 'UL')
pprint(item)
```

superdesk/editor_utils.py
ride90: Sure, that makes sense, but I would not spend time with it until manipulations with tables are required.

superdesk/editor_utils.py
petrjasek: I would actually keep fields_meta once it's parsed so we don't have to parse it again if doing multiple edits

superdesk/editor_utils.py
jerome-poisson: indeed, let's keep it for a later time

superdesk/editor_utils.py
jerome-poisson: Editor instance is not returned, but still I find it more clean to use the dedicated factory, even if `Editor2Content` would only be a subclass of `Editor3Content`. This way, if one day we have `Editor4Content`, we would not have to change this method.

jerome-poisson:
@petrjasek @ride90 RST documentation should also be updated.

superdesk/editor_utils.py
petrjasek: I've fixed it to keep styles

superdesk/editor_utils.py
```diff
@@ -12,12 +12,15 @@
 import re
 import logging
 import uuid
+import json
```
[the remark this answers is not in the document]
petrjasek: it is used now

PR #1774, opened 2020-01-16
Fix: Removed associated item gets corrected on story correction. [SDESK-4944]

GyanP:


apps/publish/content/common.py
```diff
@@ -594,6 +594,12 @@ def _publish_associated_items(self, original, updates=None):
                     original[ASSOCIATIONS][associations_key]['operation'] = self.publish_type
                     continue
 
+                # Do not publish removed association items on correction.
```
ride90: I think you can simplify it with:
```
if updates.get(ASSOCIATIONS, {}).get(associations_key) is None:
    continue
```

apps/publish/content/common.py
```diff
@@ -594,6 +594,12 @@ def _publish_associated_items(self, original, updates=None):
...
+                if (updates.get(ASSOCIATIONS, None)
+                        and associations_key in updates[ASSOCIATIONS]
+                        and updates[ASSOCIATIONS][associations_key] is None):
+                    continue
```
ride90: I think you can simplify it with:
```
if updates.get(ASSOCIATIONS, {}).get(associations_key) is None:
    continue
```

GyanP: Actually, association_key must be present, so check for updates[ASSOCIATIONS][association_key] is None only if association_key is present in updates.

ride90: makes sense

PR #1775, opened 2020-01-16
[FEEDING SERVICES][FTP] Don't move file on failed ingestion before ba…

jerome-poisson:
…ckstop delay

If a file can't be ingested with FTP, it will be tried again until
`INGEST_OLD_CONTENT_MINUTES` delay expires. The last modification date
is used to check when backstop delay. The behaviour is the same as for
file feeding service.

The `FILE_INGEST_OLD_CONTENT_MINUTES` setting has been renamed to
`INGEST_OLD_CONTENT_MINUTES` has it is now used both for `file` and
`ftp` feeding services (and maybe others in the future).

SDESK-4834

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -645,7 +645,7 @@ FTP_INGEST_FILES_LIST_LIMIT = 100
 #:
 #: .. versionadded:: 1.32.2
 #:
-FILE_INGEST_OLD_CONTENT_MINUTES = 10
+INGEST_OLD_CONTENT_MINUTES = 10
 
 #: default timeout for email connections
 EMAIL_TIMEOUT = 10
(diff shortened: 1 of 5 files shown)


superdesk/default_settings.py
```diff
@@ -638,7 +638,7 @@ def local_to_utc_hour(hour):
 #:
 #: .. versionadded:: 1.32.2
 #:
-FILE_INGEST_OLD_CONTENT_MINUTES = 10
```
petrjasek: would be good to raise some error if old config is set and not the new one

petrjasek: actually new one will be always set, so just when old one is set and is different from new config value

ride90:
@jerome-poisson it would be good to have some tests on it.

jerome-poisson:
@petrjasek @ride90 I've updated code, adding testing of old setting + test.

PR #1777, opened 2020-01-17
SDBELGA-262 Add fs.files info to events files.

ride90:
response example:
Screenshot 2020-01-17 at 11 40 10


petrjasek:
not sure this would work for amazon s3 users, if we just need filename would be better to store it somewhere on the item imo

ride90:
solved in planning https://github.com/superdesk/superdesk-planning/pull/1412

PR #1781, opened 2020-01-22
SDESK-4947 [PUBLISH] It is possible to publish related items locked by a user

ride90:


apps/publish/content/common.py
```diff
@@ -522,8 +534,19 @@ def _validate_associated_items(self, original_item, validation_errors=None):
                     validation_errors.extend(pre_errors)
 
             # check the locks on the items
-            if doc.get('lock_session', None) and original_item['lock_session'] != doc['lock_session']:
-                validation_errors.extend(['{}: packaged item cannot be locked'.format(doc['headline'])])
+            if doc.get('lock_session'):
+                if original_item['lock_session'] != doc['lock_session']:
+                    validation_errors.extend([
+                        '{}: packaged item is locked by another user'.format(
```
pablopunk: Actually this could mean it's locked by this user but on another session. Should we check `lock_user` instead? If not, maybe we can change the message to clarify that

ride90: right

PR #1788, opened 2020-02-03
Move Warning message while publishing to the back-end.

GyanP:


apps/publish/content/common.py
```diff
@@ -266,6 +268,25 @@ def _validate(self, original, updates):
...
+                and not publishing_warnings_confirmed
+                and original.get(ASSOCIATIONS)
+                and self.publish_type in [ITEM_PUBLISH, ITEM_CORRECT]):
+            for key, associated_item in original.get(ASSOCIATIONS).items():
+                if associated_item and archive_service._is_related_content(key):
+                    item = archive_service.find_one(req=None, _id=associated_item.get('_id'))
+                    if item.get('state') not in PUBLISH_STATES:
+                        error_msg = json.dumps({
+                            'id': 'publishing_warnings_confirmed',
+                            'message': _('There are unpublished related \
```
petrjasek: would rather make it a list of warnings so it's more flexible, eg:
```py
json.dumps({
    'warnings': [_('...')],
})
```
not sure we need that id for anything, but if so it would be good to have a constant for it and use it in both places

apps/publish/content/common.py
```diff
@@ -268,6 +271,26 @@ def _validate(self, original, updates):
         if len(validation_errors) > 0:
             raise ValidationError(validation_errors)
 
+    def _raise_if_unpublished_related_items(self, original):
+        archive_service = get_resource_service('archive')
+        publishing_warnings_confirmed = strtobool(request.args.get('publishing_warnings_confirmed') or 'False')
```
petrjasek: I think you should also check if request exists, it might not be set when auto publishing something via ingest routing etc:
```python
confirmed = request and strtobool(...)
```

petrjasek: also not sure what it should do in this case, if it should publish or not, @karelpetrak ?

petrjasek: my bad, this is not about publishing but showing a warning, so in case of auto publishing there is no need to warn

GyanP: How can I know if it's a auto publishing? Do we have different service for this?

petrjasek: request will be empty, so you can do
```python
if not request:
    return
```

karelpetrak: auto publishing via ingest routing should publish it

PR #1791, opened 2020-02-04
[SDESK-4984] Empty error message on insufficient privileges to create a content template

nrvikas:


apps/auth/auth.py
```diff
@@ -114,13 +114,14 @@ def check_permissions(self, resource, method, user):
             return True
 
         # Step 3: Intrinsic Privileges
+        message = 'Insufficient privileges for the requested operation.'
```
MarkLark86: Translation required for this error message. i.e.:
```
from flask_babel import _
....
    message = _('Insufficient privileges for the requested operation.')
```

PR #1792, opened 2020-02-04
fix related items refs

petrjasek:
make sure we only keep id and type for related item associations

SDESK-4954


GyanP:
Required for https://dev.sourcefabric.org/browse/SDESK-4960

PR #1793, opened 2020-02-04
SDESK-4947 Do not publish locked associated items during correction.

ride90:
SDESK-4947
- don't validate _if an associated item is locked_ during publishing if PUBLISH_ASSOCIATED_ITEMS is false
- use associations from `original_item` merged with`updates` during `_validate_associated_items`

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -544,14 +544,16 @@ class BasePublishService(BaseService):
                 if doc.get('lock_user'):
                     if original_item['lock_user'] != doc['lock_user']:
                         validation_errors.extend([
-                            '{}: packaged item is locked by another user'.format(
-                                doc.get('headline', doc['_id'])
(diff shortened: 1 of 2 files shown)


apps/publish/content/common.py
```diff
@@ -533,20 +539,21 @@ def _validate_associated_items(self, original_item, validation_errors=None):
...
-                        '{}: packaged item is locked by you. Unlock it and try again'.format(
-                            doc.get('headline', doc['_id'])
-                        )
-                    ])
+            if config.PUBLISH_ASSOCIATED_ITEMS:
+                # check the locks on the items
+                if doc.get('lock_user'):
+                    if original_item['lock_user'] != doc['lock_user']:
+                        validation_errors.extend([
+                            '{}: packaged item is locked by another user'.format(
```
petrjasek: you can use `_(` for the message so we could translate it later

PR #1796, opened 2020-02-06
Add mark_for_user privilege. [SDBELGA-300]

GyanP:
Required for https://github.com/superdesk/superdesk-client-core/pull/3341


petrjasek:
can you make it translatable?

PR #1803, opened 2020-02-13
Add method to retrieve the whole items chain. Move tests into tests dir

ride90:


apps/archive/archive.py
```diff
@@ -1026,6 +1026,73 @@ def _send_mark_user_notifications(self, activity_name, msg, resource=None, item=
...
+                except Exception:
+                    break
+            return item
+
+        item = get_item_translated_from(item)
+        # add item + translations
+        items_chain = [item]
+        items_chain += self.get_item_translations(item)
+
+        while True:
```
petrjasek: as a safety measure I would prefer something like `for i in range(0, 50):` to avoid infinite loops

PR #1807, opened 2020-02-13
[EDITOR UTILS] embed pre-processing + fixed HTML parsing:

jerome-poisson:
pre-processing of embed for customisation is now possible using
callables set in the new `EMBED_PRE_PROCESS` setting.

This patch also fix parsing of HTML, as part of the HTML could be lost
with DOM.parse_html + selecting only children (seen with QUMU embed).
As we use `lxml` as draftjs_exporter backend, we can use
`superdesk.etree.parse_html`.

SDESK-4939

### DIFF
--- a/docs/editor.rst
+++ b/docs/editor.rst
@@ -48,3 +48,15 @@ show how to add an embedded code at the top of the body::
        body_editor.prepend('embed', '<p class="some_class">some embedded HTML to
        prepend</p>')
        body_editor.update_item()
+
+Rendering customisation
+=======================
+
+Sometimes, you may want to modify the behavior of the renderer.
+
+embeds
+------
+
+Embeds can be modified before being rendered. For that, the ``EMBED_PRE_PROCESS`` setting
+can be used. It is an iterable linking to callable. Each callable will get the data dict
+where the ``html`` key contains the raw HTML, so it can modify it.
--- a/tests/editor_utils_test.py
+++ b/tests/editor_utils_test.py
@@ -11,13 +11,18 @@
 
 import json
 import uuid
-import unittest
+from superdesk.tests import TestCase
 import superdesk.editor_utils as editor_utils
 
 from superdesk.editor_utils import Editor3Content
 
 
-class Editor3TestCase(unittest.TestCase):
+class Editor3TestCase(TestCase):
+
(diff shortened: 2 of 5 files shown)


jerome-poisson:
@petrjasek FYI I've added a test + modified the parsing of HTML in `superdesk.etree` to fix handling of some corner cases (notably the use of `<script>` as it is the case with QUMU).

docs/editor.rst
```diff
@@ -48,3 +48,15 @@ show how to add an embedded code at the top of the body::
...
+Rendering customisation
+=======================
+
+Sometimes, you may want to modify the behavior of the renderer.
+
+embeds
+------
+
+Embeds can be modified before being rendered. For that, the ``EMBED_PRE_PROCESS`` setting
+can be used. It it an iterable linking to callable. Each callable will get the data dict
```
ride90: Typo "It it" -> "It is"

tests/editor_utils_test.py
```diff
@@ -11,13 +11,18 @@
...
 from superdesk.editor_utils import Editor3Content
 
 
-class Editor3TestCase(unittest.TestCase):
+class Editor3TestCase(TestCase):
+
+    def setUp(self):
+        super().setUp()
+        if "EMBED_PRE_PROCESS" in self.app.config:
+            del self.app.config["EMBED_PRE_PROCESS"]
```
ride90: why do we need this?

jerome-poisson: Because the config is global to the whole test, and when I modify `EMBED_PRE_PROCESS` at https://github.com/superdesk/superdesk-core/pull/1807/files#diff-80f1d1a029ab2f40ac5996f23c54282aR1614 it will impact all following test, which may fail if they use embed. I'm not sure it there is a cleaner way.

ride90: Got it thanks.
I think `setUp` is called before every test anyway and by default `EMBED_PRE_PROCESS` is not set anyway.

ride90: ahh, ok got it now. You can ignore my previous comment :)

docs/editor.rst
jerome-poisson: fixed, thx

PR #1809, opened 2020-02-18
[SDESK-5021] Extend FTP transmit to push associations

marwoodandrew:


superdesk/publish/formatters/ninjs_ftp_formatter.py
```diff
@@ -0,0 +1,60 @@
...
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+
+from .ninjs_formatter import NINJS2Formatter
+from superdesk.storage.amazon_media_storage import _guess_extension
```
petrjasek: imo would be good to move this into some `media_utils` module

superdesk/publish/transmitters/ftp.py
```diff
@@ -51,13 +54,67 @@ def _transmit(self, queue_item, subscriber):
 
         try:
             with ftp_connect(config) as ftp:
+
+                if config.get('push_associated', False):
```
petrjasek: is this config only available with ninjs formatter?

marwoodandrew: This is a config setting for the FTP transmitter, set in the config for the transmitter in client-core.
There is an assumption that the formatted item will be ninjs, but if there are no associations in whatever format it should not be a problem!

superdesk/publish/formatters/ninjs_ftp_formatter.py
marwoodandrew: I've moved the functions off into the media module

superdesk/publish/transmitters/ftp.py
petrjasek: I'm not sure about this `item = json.loads(queue_item['formatted_item'])` in case of eg newsml, maybe it should have some `try/catch` on it

marwoodandrew: Good point! I've put a try catch on it.

superdesk/publish/formatters/ninjs_ftp_formatter.py
```diff
@@ -0,0 +1,48 @@
...
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+
+from .ninjs_formatter import NINJS2Formatter
+from superdesk.media.renditions import get_rendition_file_name
+
+
+class FTPNinjsFormatter(NINJS2Formatter):
```
MarkLark86: We will need this in Superdesk v1.31, so it might be better to inherit from the original NINJSFormatter

PR #1812, opened 2020-02-20
Fix missing body_html field for updates of scheduled items. [SDBELGA-307]

GyanP:


apps/archive/archive_rewrite.py
```diff
@@ -227,8 +227,8 @@ def _create_rewrite_article(self, original, existing_item=None, desk_id=None):
             # send the document to the desk only if a new rewrite is created
             send_to(doc=rewrite, desk_id=(desk_id or original['task']['desk']), default_stage='working_stage')
 
-            # if we are rewriting a published item then copy the body_html
-            if original.get('state', '') in (CONTENT_STATE.PUBLISHED, CONTENT_STATE.CORRECTED):
+            # if we are rewriting a published item then copy the
```
petrjasek: seems like comment is missing the last wod

PR #1813, opened 2020-02-20
push resource:updated notification when there is an update

petrjasek:
SDESK-5053


superdesk/eve_backend.py
```diff
@@ -22,6 +22,35 @@
 from eve.methods.common import resolve_document_etag
 from elasticsearch.exceptions import RequestError, NotFoundError
 from superdesk.errors import SuperdeskApiError
+from superdesk.notification import push_notification
+
+
+SYSTEM_KEYS = set([
```
tomaskikutis: you might want to also add `_links` and `_id`

petrjasek: these are not stored in db so it won't make it there (`_id` is stored but mongo won't accept it in update payload)

PR #1818, opened 2020-02-26
As a user I want to filter content by Planning Agenda. [SDESK-5055]

GyanP:


apps/content_filters/filter_condition/filter_condition_parameters.py
```diff
@@ -205,3 +212,11 @@ def _get_stage_field_values(self, desks):
                 continue
             stages[i]['name'] = '{}: {}'.format(desk['name'], stage['name'])
         return tuple(i for i in stages if i)
+
+    def _get_planning_agendas(self):
+        agendas = []
+        agenda_service = get_resource_service('agenda')
+        cursor = agenda_service.find({'is_enabled': True})
```
petrjasek: you can do this via `return list(agenda_service.find({'is_enabled': True}))`

apps/content_filters/filter_condition/filter_condition_parameters.py
```diff
@@ -137,6 +138,12 @@ def get(self, req, lookup):
                   {'field': 'anpa_take_key',
                    'operators': ['in', 'nin', 'eq', 'ne', 'like', 'notlike', 'startswith', 'endswith']
                    },
+                  {'field': 'agendas',
```
petrjasek: actually you could add this field only if agenda is enabled, or even better create some api which planning could use to register a field..

petrjasek: but given that it's the first external field, I would probably just make it conditional and comment on adding api if needed next time..

GyanP: Is there a way to check if agendas or enabled or not? If not I can simply check the length of `agenda_values` that is coming from here `agenda_values = self._get_planning_agendas()`. Would it be fine?

petrjasek: I think it's this `if 'planning' in app.config.get('INSTALLED_APPS', []):` you do elsewhere

PR #1824, opened 2020-03-05
FIX: Related items order not same in editor and ninjs output. [SDESK-5083]

GyanP:
Introduced order in related items.
requires : https://github.com/superdesk/superdesk-client-core/pull/3371

### DIFF
--- a/superdesk/data_updates/00022_20200309-121947_archive.py
+++ b/superdesk/data_updates/00022_20200309-121947_archive.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8; -*-
+# This file is part of Superdesk.
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+#
+# Author  : GyanP
+# Creation: 2020-03-09 12:19
+
+from bson import ObjectId
+from bson.errors import InvalidId
+from superdesk import get_resource_service
+from superdesk.vocabularies import is_related_content
+from superdesk.commands.data_updates import DataUpdate
+
+
+class DataUpdate(DataUpdate):
+
+    resource = 'archive'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        related = list(get_resource_service('vocabularies').get(req=None, lookup={'field_type': 'related_content'}))
+        archive_service = get_resource_service('archive')
+        for resource in ('archive', 'published'):
+            service = get_resource_service(resource)
+            for item in mongodb_database[resource].find({'associations': {'$gt': {}}}):
+                update = False
+                associations = {}
+                for key, val in item['associations'].items():
+                    associations[key] = val
+                    if val and is_related_content(key, related) and val.get('order', None) is None:
+                        update = True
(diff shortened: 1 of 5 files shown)


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -301,14 +300,13 @@ def _get_associations(self, article, subscriber):
...
-        associations = OrderedDict()
+        associations = {}
         extra_items = {}
         media = {}
         content_profile = None
         archive_service = superdesk.get_resource_service('archive')
-        sorted_associations = OrderedDict(sorted(article.get(ASSOCIATIONS).items() or {}))
 
-        for key, item in sorted_associations.items():
+        for key, item in article.get(ASSOCIATIONS).items() or {}:
```
ride90: you can type `article.get(ASSOCIATIONS, {}).items()`

superdesk/data_updates/00022_20200309-121947_archive.py
```diff
@@ -0,0 +1,43 @@
...
+from superdesk.commands.data_updates import DataUpdate
+
+
+class DataUpdate(DataUpdate):
+
+    resource = 'archive'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        related = list(get_resource_service('vocabularies').get(req=None, lookup={'field_type': 'related_content'}))
+        archive_service = get_resource_service('archive')
```
ride90: why do we need it if we use `mongodb_database[resource]` ?

superdesk/data_updates/00022_20200309-121947_archive.py
```diff
@@ -0,0 +1,43 @@
...
+        for resource in ('archive', 'published'):
+            service = get_resource_service(resource)
+            for item in mongodb_database[resource].find({'associations': {'$gt': {}}}):
```
ride90: IMO this filter condition is weird. 
Wouldn't it be more explicit to use something like that
`{ associations: { $exists: true, $ne: {} }`  (not tested, just guessed)

superdesk/data_updates/00022_20200309-121947_archive.py
```diff
@@ -0,0 +1,43 @@
...
+from superdesk.vocabularies import is_related_content
+from superdesk.commands.data_updates import DataUpdate
+
+
+class DataUpdate(DataUpdate):
+
+    resource = 'archive'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        related = list(get_resource_service('vocabularies').get(req=None, lookup={'field_type': 'related_content'}))
```
ride90: this is redundant, it's done in `is_related_content` if `related_content` arg is not provided.

superdesk/data_updates/00022_20200309-121947_archive.py
GyanP: i forgot to remove.

superdesk/data_updates/00022_20200309-121947_archive.py
GyanP: good catch.I didn't looked at `is_related_content` definition.

PR #1825, opened 2020-03-05
[MACROS] log warning when a macro loading fail

jerome-poisson:
Macro loading was failing silently, this patch fix this behaviour.


jerome-poisson:
note: there is not issue attached.

PR #1826, opened 2020-03-06
[SDESK-4907] Impliment mapping between Media Topic codes and IPTC sub…

marwoodandrew:
…ject codes


superdesk/io/mediatopics.py
```diff
@@ -0,0 +1,180 @@
...
+        # The media topics json file is available at http://cv.iptc.org/newscodes/mediatopic?lang=x-all
+        # there are also language specific files that could be used.
+        media_topic_codes = os.path.join(dirname, 'data', 'cptall-x-all.json')
+        mediatopics = self._load_codes(media_topic_codes)
+        self.generate_mediatopic_to_subject_map(mediatopics.get('conceptSet'))
+
+    def _load_codes(self, filename):
+        with open(filename, 'rb') as f:
+            codes_bytes = f.read()
+            codes = loads(codes_bytes.decode('UTF-8'))
```
petrjasek: could be a bit simpler via `with open(filename, 'rb') as f: codes = json.load(f)`

marwoodandrew: Done

PR #1828, opened 2020-03-09
Talpa: Prevent publishing videos without a thumbnail. [SDESK-5082]

GyanP:


petrjasek:
this should be rather done in superdesk-tlp repo using item_validate signal like [here](https://github.com/superdesk/superdesk/blob/ansa/server/ansa/validate.py)

PR #1833, opened 2020-03-16
[SDBELGA-324] Handle empty document on ingest

idsvn:
From Karel
> ...Document is empty... error
this happens from time to time when a file is not saved on the server while the ingest is trying to process it. When it fails, ingest will retry and the file gets ingested in the end.

It also raises error when ingest file is empty, so we think you could use this to reproduce the error. The error is on file service but we think we should handle ftp service as well.

SDBELGA-324

### DIFF
--- a/superdesk/io/feeding_services/ftp.py
+++ b/superdesk/io/feeding_services/ftp.py
@@ -197,6 +197,11 @@ class FTPFeedingService(FeedingService):
         _, ext = os.path.splitext(filename)
         return ext.lower() in allowed_ext
 
+    def _is_empty(self, file_path):
+        """Test if given file is empty, return True if file does not exist
+        """
+        return not (os.path.isfile(file_path) and os.path.getsize(file_path) > 0)
+
     def _list_files(self, ftp, provider):
         self._timer.start('ftp_list')
         try:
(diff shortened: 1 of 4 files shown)


superdesk/io/feeding_services/ftp.py
```diff
@@ -197,6 +197,11 @@ def _is_allowed(self, filename, allowed_ext):
         _, ext = os.path.splitext(filename)
         return ext.lower() in allowed_ext
 
+    def _is_empty(self, file_path):
```
ride90: Does it make sense to put it somewhere else to avoid repeating code? We can create a file mixin and use it for `FileFeedingService` and `FTPFeedingService` or even put it in utils... It's a tiny method, maybe it's not worth it. @petrjasek ?

ride90:
@idsvn please fix behave tests.

superdesk/io/feeding_services/ftp.py
idsvn: Should we put it to BaseClass `FeedingService`? We intent to do so but not sure it's suitable?!

ride90: IMO `is_empty` is file related, so the base class is not the right place.
Not a big deal, you can leave it as it is now.

PR #1837, opened 2020-03-18
Unpublished item should get published as association (depending on config). [SDESK-5075]

GyanP:


petrjasek:
is it develop branch issue only or also affects 1.33?

GyanP:
Not sure. In the ticket, fix version is `1.33.1`. But, can be reproduced in both 1.33 and develop.

PR #1838, opened 2020-03-19
FIX: Not all fields in the place property are sent using Ninjs formatter. [SDESK-5080]

GyanP:
I will set `NINJS_PLACE_EXTENDED` to False before merging.

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -836,4 +836,4 @@ TANSA_CLIENT_BASE_URL = env('TANSA_CLIENT_BASE_URL', 'https://d02.tansa.com/tans
 TANSA_PROFILES = {}
 
 # Enable ninjs to send all the fields for place in output.
-NINJS_PLACE_EXTENDED = True
+NINJS_PLACE_EXTENDED = False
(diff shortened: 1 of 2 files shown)


petrjasek:
I'm not sure we should just change the output format, that might be problematic for receiving systems. so we would either need a config for it, or even use a custom format for publisher as this is not respecting the specs

GyanP:
I think having a config like `SEND_ALL_FIELDS_FOR_PLACE_IN_OUTPUT` and based on this decide to what to send. Or if we plan to create a custom format for a publisher then how to decide whether this is for the publisher or other?

petrjasek:
what about `NINJS_PLACE_EXTENDED` setting? it's just ninjs not all outputs

superdesk/default_settings.py
```diff
@@ -834,3 +834,6 @@ def local_to_utc_hour(hour):
 TANSA_LICENSE_KEY = env('TANSA_LICENSE_KEY')
 TANSA_CLIENT_BASE_URL = env('TANSA_CLIENT_BASE_URL', 'https://d02.tansa.com/tansaclient/')
 TANSA_PROFILES = {}
+
+# Enable ninjs to send all the fields for place in output.
+NINJS_PLACE_EXTENDED = True
```
petrjasek: would make it `False` by default so we don't change behaviour

GyanP: I will set it to false before merging. I already mentioned in the first comment.

PR #1839, opened 2020-03-24
'TIPS ingest macro' should also change content profile. [SDBELGA-304]

GyanP:


GyanP:
The default CP overwrites the CP setup by TIPS ingest macro. So, not sure whether this is the correct approach to solve this.

apps/duplication/archive_fetch.py
```diff
@@ -100,7 +100,7 @@ def fetch(self, docs, id=None, **kwargs):
             self.__fetch_associated_items(dest_doc, desk_id, stage_id, doc.get(ITEM_STATE, CONTENT_STATE.FETCHED))
 
             desk = get_resource_service('desks').find_one(req=None, _id=desk_id)
-            if desk and desk.get('default_content_profile'):
+            if desk and desk.get('default_content_profile') and not doc.get('macro'):
                 dest_doc['profile'] = desk['default_content_profile']
```
petrjasek: actually I think we should keep this, but probably as:
```py
dest_doc.setdefault('profile', desk['default_content_profile'])
```
otherwise we're assuming that macro will set profile which might not be the case

PR #1841, opened 2020-03-26
'TIPS ingest macro' should also change content profile. [SDBELGA-304]

GyanP:


apps/duplication/archive_fetch.py
```diff
@@ -101,7 +101,7 @@ def fetch(self, docs, id=None, **kwargs):
 
             desk = get_resource_service('desks').find_one(req=None, _id=desk_id)
             if desk and desk.get('default_content_profile') and not doc.get('macro'):
```
petrjasek: I would remove this `and not doc.get('macro')` part so it gets set also if there is macro which doesn't set profile

PR #1842, opened 2020-03-26
update eve/flask

petrjasek:
eve requires now all fields that should be returned in response to be part of the resource schema, so had to add some fields we generate on server, but it's more explicit so I think it's a good thing :tm: 

another significant change was that `data.find` returns now [result and count](https://github.com/pyeve/eve/blob/master/eve/io/mongo/mongo.py#L292), not just result, I guess we should add some internal api for this and not use those from framework internally so we don't have to handle such changes next time.

last but not least, there is [new version of cerberus](https://docs.python-cerberus.org/en/stable/upgrading.html)


apps/archive/archive_rewrite.py
```diff
@@ -269,6 +280,7 @@ def _get_ordinal(self, n):
             return str(n) + {1: 'st', 2: 'nd', 3: 'rd'}.get(n % 10, "th")
 
     def delete(self, lookup):
+        print("IN")
```
MarkLark86, line 283: Remove `print` statement

MarkLark86:
My main concern with upgrading Eve is the new `MERGE_NESTED_DOCUMENTS` functionality. Especially when it comes to fields like `associations`, we have to make sure the client nulls the fields that are removed when sending `PATCH`, where as at the moment what we send in the `PATCH` is expected to be the new sub-document/fields.

petrjasek:
that was the main reason why we wanted to do the upgrade, but I want to do that as a second step, it was already pretty big PR

superdesk/validator.py
```diff
@@ -56,59 +98,61 @@ def _validate_type_email(self, field, value):
...
             emails = value.split(',')
-            for email in emails:
-                self._validate_type_email(field, email)
+            if not all([self._validate_type_email(email) for email in emails]):
+                self._error(field, ERROR_PATTERN)
 
     def _validate_unique(self, unique, field, value):
-        """Validate unique with custom error msg."""
-
+        """
```
ride90, line 124: I got a lot of warnings after the develop server was started.

Almost all these warnings are like:
```
UserWarning: No validation schema is defined for the arguments of rule 'empty'
  "'%s'" % method_name.split('_', 2)[-1]
```

I think it can be solved by adding exact this price of docstring:

```
The rule's arguments are validated against this schema:
{'type': 'boolean'}
```

PR #1844, opened 2020-03-26
[CLI] added a users:import command

jerome-poisson:
users:import will create many users from a JSON file.

SDBELGA-343

### DIFF
--- a/apps/auth/db/commands.py
+++ b/apps/auth/db/commands.py
@@ -9,6 +9,9 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 import logging
+import json
+import csv
+from pathlib import Path
 from base64 import b64encode
 from flask import current_app as app
 import superdesk
@@ -16,6 +19,7 @@ from superdesk.utils import get_hash, is_hashed
 
 
 logger = logging.getLogger(__name__)
+USER_FIELDS_NAMES = {'username', 'email', 'password', 'first_name', 'last_name', 'sign_off', 'role'}
 
 
 class CreateUserCommand(superdesk.Command):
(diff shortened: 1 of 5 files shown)


petrjasek:
can we have this in belga repo?

petrjasek:
also wondering - does it create active users or they must activate later?

petrjasek:
closed by mistake

jerome-poisson:
> can we have this in belga repo?

@petrjasek why? This is a generally useful command, I don't see any reason to do it only for Belga.

jerome-poisson:
> I've tested it locally and it works.
> The only issue I've experienced is a path to JSON file as a positional argument for a command.
> It's a rare case and I think nobody will face it, in short on my local computer I was able to load a wrong python pyenv when executing a command like this
> 
> ```
> python manage.py users:import "/Users/olegpshenichniy/PycharmProjects/superdesk_dev/superdesk-core/tests/auth/fixtures/import_users_test.json"
> ```
> 
> (with or without quotes),

I don't understand your problem to be honest. How can a positional argument to a CLI command change the venv of Python? `python` is specified in your command that's what determine the venv used.

 I would suggest to call it without positional argument:
> 
> ```
> python manage.py users:import --import_file="/Users/olegpshenichniy/PycharmProjects/superdesk_dev/superdesk-core/tests/auth/fixtures/import_users_test.json"
> ```

Well positional arguments are used for mandatory arguments (which is the case for the JSON file, the command can't be run without). Using an optional argument make the command more tedious to write and doesn't follow usual CLI behaviours. So I don't this is a good idea.


> @jerome-poisson I think we should clarify a moment with `password`. Does it make sense to send an email with a link to reset a password and not provide it in a file as a plain text?

I was wondering also when doing the command. I was thinking this could be used for a first-use password which would make on-boarding easier, but maybe it makes more sense to just send a link (or maybe we can offer both options and let the admin decide). Let's me know what you think @ride90 and @petrjasek .

> IMO this feature is generic and not really Belga specific at the end, so we might keep it in the core.
yes I think so too.

ride90:
> I don't understand your problem to be honest. How can a positional argument to a CLI command change the venv of Python? python is specified in your command that's what determine the venv used.

Weird indeed, but the issue is real :) I think it's a pyenv issue. Anyway, you can keep the positional argument.

petrjasek:
>> @jerome-poisson I think we should clarify a moment with password. Does it make sense to send >> an email with a link to reset a password and not provide it in a file as a plain text?
>
> I was wondering also when doing the command. I was thinking this could be used for a first-use 
> password which would make on-boarding easier, but maybe it makes more sense to just send a 
> link (or maybe we can offer both options and let the admin decide). Let's me know what you think 
> @ride90 and @petrjasek

I think it's ok to populate the password as long as we store it in secure way, they might want to migrate existing users with their existing password (I think ansa did it somehow).

petrjasek:
also wondering if they use excel why not support csv rather than json

jerome-poisson:
> also wondering if they use excel why not support csv rather than json

because the issue was mentioning JSON, and I think it is more readable, but I can add a `--format CSV` argument, it's not a big deal.

petrjasek:
yep json is more readable for sure, just wondering how would they convert that excel sheet into json. if we can support both that would be easier for users imo

jerome-poisson:
@petrjasek @ride90 I've added support for CSV files. For now, I'm letting the current password behaviour, as it seems fine (it has been confirmed by Belga).

PR #1845, opened 2020-03-27
FIX: FTP ingest doesn't parse PATH from ingest settings correctly. [SDESK-5060]

GyanP:


superdesk/ftp.py
```diff
@@ -25,7 +25,10 @@ def ftp_connect(config):
             ftp.login(config.get('username'), config.get('password'))
         except ftplib.error_perm as e:
             raise IngestFtpError.ftpAuthError(exception=e)
-    if config.get('path'):
+    path = config.get('path')
+    if path:
+        if not path.startswith('/') or not path.startswith('./'):
+            config['path'] = '/' + path
```
petrjasek, line 31: I'm a bit lost here, why add it if there is `.lstrip('/')` on it?

GyanP: The error was generating [here](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feeding_services/ftp.py#L172) and to overcome this problem I have added '/' in the path. Since, this path can be used at other places I modified the path.

petrjasek: shouldn't we handle it rather there then?

PR #1849, opened 2020-04-09
FIX: Fetched items not always appear in the according desk/stage. [SDESK-5084]

GyanP:


petrjasek:
I'm not sure we should avoid that filter, what about making sure the version is set to something bigger when fetching an item

petrjasek:
but might be actually issue with the ingest

petrjasek:
we can handle it in parser like we do with [versioncreated](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feed_parsers/ap_media.py#L158-L159):
```py
if in_item.get('version'):
    item['version'] = in_item['version']
```
that would filter out `None` and `0` as well

petrjasek:
you should probably also remove it from [direct_copy_properties](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feed_parsers/ap_media.py#L40)

PR #1851, opened 2020-04-21
Determine content type on the server side

ride90:
SDBELGA-385


superdesk/storage/mimetype_mixin.py
```diff
@@ -0,0 +1,46 @@
...
+import mimetypes
+
+import magic
+from werkzeug import datastructures
+
+
+logger = logging.getLogger(__name__)
+
+
+class MimetypeMixin:
```
petrjasek, line 12: just wondering why mixing and not just a function?

ride90: `get_mimetype` method is very specific to `SuperdeskGridFSMediaStorage` and `AmazonMediaStorage`. IMO the best place for it is `eve.io.media.MediaStorage`, but since it's not an option I've used mixin for it. Having a function might make somebody think that it's reusable, but it's not. It has very specific args. 
I can move it to `superdesk/media/media_operations.py` as a function if you think it's a better place.

.travis.yml
```diff
@@ -14,6 +14,10 @@ before_install:
       if [ "$SKIP_SERVER_INSTALL" != "true" ]; then
         docker-compose -f .travis-docker-compose.yml up -d
       fi
+    - wget http://launchpadlibrarian.net/433926958/libmagic-mgc_5.37-3_amd64.deb
+    - wget http://launchpadlibrarian.net/433926961/libmagic1_5.37-3_amd64.deb
+    - sudo dpkg -i libmagic-mgc_5.37-3_amd64.deb
+    - sudo dpkg -i libmagic1_5.37-3_amd64.deb
```
ride90, line 20: I found an issue with `libmagic<5.37` (both on OSX and bionic), it does not recognize `xlsx` and returns `application/octet-stream` which is critical for Belg CDN. libmagic 5.37/5.38 does the job.
Bionic has a 5.32 (or lower).
I will mention it in the release notes that libmagic>=5.37 is required.

superdesk/storage/desk_media_storage.py
```diff
@@ -75,6 +77,11 @@ def put(self, content, filename=None, content_type=None, metadata=None, resource
         :param str folder: Folder that the file will be stored in
         :return str: The ID that was generated for this object
         """
+        # try to determine mimetype on the server
+        determined_content_type = self._get_mimetype(content, filename)
+        if determined_content_type:
```
petrjasek, line 82: imo might be interesting to log when `content_type != determined_content_type`, I would also maybe change a signature a bit to avoid that duplication:
```py
content_type = self._get_mimetype(content, filename, content_type)
```
and might skip libmagic when eg `contet_type` param starts with `image`

ride90: I think we can also skip when content type starts with `video` and `audio`

PR #1858, opened 2020-04-29
FIX: Spiking of unpublished items. [SDESK-5104]

GyanP:


petrjasek:
so what's the idea here in general? I think we should avoid those exceptions in `if` statements, why do we need those?

GyanP:
> so what's the idea here in general? I think we should avoid those exceptions in `if` statements, why do we need those?

We need those if statements because now the workflow is all changed.
Firstly whenever we try to spike an unpublished item it gives us the invalid state error.
Secondly we can't send the unpublished item to another desk because of the workflow.
So for that cases we need this if statements.

PR #1861, opened 2020-05-04
allow non-text content types

petrjasek:
and use those for validation. only 1 is allowed for each type.

SDESK-5217 SDESK-5157

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -49,6 +49,11 @@ class ContentTypesResource(superdesk.Resource):
             'type': 'string',
             'iunique': True,
         },
+        'item_type': {
+            'type': 'string',
+            'nullable': True,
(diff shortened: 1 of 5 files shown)


apps/content_types/content_types.py
```diff
@@ -49,6 +49,11 @@ class ContentTypesResource(superdesk.Resource):
             'type': 'string',
             'iunique': True,
         },
+        'item_type': {
+            'type': 'string',
+            'nullable': True,
```
ride90, line 54: Why nullable is allowed? 
We can make migration and set `text` for already created entries + use `unique` instead of `content_type_single_item_type`

petrjasek: you can have multiple `text` types (or `null` in this case), so that unique check won't work. I try to avoid migrations in general, users might not run them and here supporting both is not complex

PR #1865, opened 2020-05-07
[ARCHIVE] generate fields with a draftJS content state on save

jerome-poisson:
SDESK-4911


jerome-poisson:
we are lacking a sequence diagram for the Superdesk operation. I believe generating during the `save` is enough, but is there any use case when saving is not used?

jerome-poisson:
@tomaskikutis suggested that we add a check for a little while to check that client and backend generation do match, cf. https://dev.sourcefabric.org/browse/SDESK-4911?focusedCommentId=119033&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-119033

petrjasek:
I think we might need to call this also when saving correction

tomaskikutis:
We have that websocket notification when any resource changes, how about using that or a similar generic method instead of coupling this to features?

jerome-poisson:
I've added the safety check, generation on correction, and the tests should be fixed.

superdesk/editor_utils.py
```diff
@@ -282,7 +293,18 @@ def render(self):
                 del blocks[-1]
             html = self.exporter.render(content_state)
         else:
-            html = self.exporter.render(self.content_state)
+            try:
+                html = self.exporter.render(self.content_state)
+            except KeyError as e:
+                if e.args == ('text',):
+                    # "text" may be missing in some case (e.g. comments), and the exporter
+                    # doesn't support it. To avoid a crash, be render again after
```
petrjasek, line 301: `s/be/we/`

features/archive.feature
```diff
@@ -1116,3 +1116,62 @@ Feature: News Items Archive
...
+        Given "archive"
+        """
+        [{"_id": "test_editor_gen_1", "guid": "test_editor_gen_1", "headline": "test"}]
+        """
+
+        When we patch given
+        """
+        {
+            "fields_meta": {
+                "body_html": {
```
petrjasek, line 1131: can you also add headline to tests? I wonder how it handles single line fields

superdesk/editor_utils.py
jerome-poisson: good catch, thanks

features/archive.feature
jerome-poisson: sure

jerome-poisson: @petrjasek done

jerome-poisson:
@ride90 @petrjasek after testing again, I see that I have many inconsistencies between client and backend, notably the headline is text only with client generated (even if I  add bold and underline), while it's HTML with the backend. I'm not sure how to check it, investigating.

tomaskikutis:
I think it's hardcoded here https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring/views/article-edit.html#L42

To fix it properly we should move that property to a content profile

jerome-poisson:
oh right, thanks @tomaskikutis , I'm putting `WIP` label on this PR until this is fixed.
@petrjasek @ride90 what do you think? Should I do a quick fix with hardcoded values in a first time, and create a ticket to do it properly?

For the record, here is what I get with a headline with underline and bold:
```
WARNING:superdesk.editor_utils:Generated HTML inconsistency between client and backend:
client field: Grève du 16 janvier
backend field: <p><b><u>Grève</u></b> du 16 janvier</p>
field state: {'blocks': [{'key': 'nmcg', 'text': 'Grève du 16 janvier', 'type': 'unstyled', 'depth': 0, 'inlineStyleRanges': [{'offset': 0, 'length': 5, 'style': 'BOLD'}, {'offset': 0, 'length': 5, 'style': 'UNDERLINE'}], 'entityRanges': [], 'data': {'MULTIPLE_HIGHLIGHTS': {}}}], 'entityMap': {}}

```

jerome-poisson:
I have also inconsistency with special char like  non breakable space which is put verbatim by the backend, when `&nbsp;` is used by the client. Should be easy to fix.

jerome-poisson:
I've updated the PR with:
- hardcoding of text only values (`headline` and `slugline`), I've created https://dev.sourcefabric.org/browse/SDESK-5316 to implement a cleaner solution later
- remove warning for COMMENTS, there is no rendering of comments, so we just ignore them
- strip text only fields, to be consistent with client
- use client version when the is inconsistency between client and backend.

For the latter, and as I've said this morning, the goal is to inspect the log during the transition period, without breaking existing generation, but I'm not sure if it's the right thing to do, maybe it's better to go directly with backend version to catch problems early? Also I've used a warning, but an error would catch more attention.

There is still the non-breakable space issue, client generate a `&nbsp;` while in backend it's a simple space. I've inspected and the `nbsp` is not present in draftJS state received in the backend. After checking with @tomaskikutis, it seems that client replaces `nbsp` by common space, but this need more investigation. I'll put this in a separate ticket for now.

 @ride90 and @petrjasek I know that you have already validated this PR, but could you double check it with the modification before I merge? Thanks!

jerome-poisson:
**note**: here is the ticket about the NBSP issue ==> https://dev.sourcefabric.org/browse/SDESK-5320

PR #1867, opened 2020-05-08
Updating readonly property in content profile doesn't work for custom fields. [SDESK-5241]

GyanP:


features/content_profile.feature
```diff
@@ -1265,7 +1265,7 @@ Feature: Content Profile
                     },
                     "type": "list",
                     "required": true,
-                    "mandatory_in_list": {"scheme": {"category": "category"}}
+                    "mandatory_in_list": {"scheme": {"category": {"required": true}}}
```
petrjasek, line 1268: I guess such change would require upgrade script, wouldn't it?

GyanP: Oops, surely there should be one. Will write one.

tests/validate/validate_mandatory_in_list_test.py
```diff
@@ -32,7 +32,7 @@ def test_fail_validate_mandatory_in_list_for_subject(self):
         self.assertIn('is a required field', validator.errors['subject'])
 
     def test_fail_validate_mandatory_in_list_for_subject_and_category(self):
-        mandatory = {'scheme': {'subject': 'custom_subject', 'category': 'category'}}
+        mandatory = {'scheme': {'subject': {"required": True}, 'category': {"required": True}}}
```
petrjasek, line 35: shouldn't it be rather `custom_subject: {"required": True}`?

PR #1872, opened 2020-05-13
implement media usage tracking for all associations

petrjasek:
not only featuremedia

SDANSA-395

### DIFF
--- a/features/links.feature
+++ b/features/links.feature
@@ -1,3 +1,4 @@
+@wip
 Feature: Track usage of an item
 
     @auth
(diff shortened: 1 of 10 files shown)


features/links.feature
```diff
@@ -0,0 +1,57 @@
...
+        """
+        {"associations": {"pic": {"_id": "two", "type": "picture", "renditions": {}}}}
+        """
+        Then we get OK response
+
+        When we get "links?guid=two"
+        Then we get list with 1 items
+        """
+        {"_items": [
+            {"_id": "one", "type": "text"}
```
ride90, line 27: The real response will contain a full document(s) not only `_id` and `type`, right?

superdesk/metadata/item.py
```diff
@@ -452,6 +452,31 @@
             }
         }
     },
+
+    # track references to other objects,
+    # based on associations but allows queries
+    'refs': {
```
ride90, line 458: I think we might need a migration to fill items `refs` based on their existing associations.

features/links.feature
petrjasek: yep right

superdesk/metadata/item.py
petrjasek: ansa won't need it as they only care about last 24h, but if anyone else would want that to work for all content ever then yes

PR #1873, opened 2020-05-14
Ability to change schema(s) from the instance settings

ride90:
~~⚠️ This is not a finished PR, more like a POC (it works).~~

 The use case for Belga is the ability to search/order using fields from `archive.extra` in production API using elastic search. By default mapping for `extra` is not enabled, so it would be handy to have an ability to update a schema from the instance settings.

I've tested this approach and it works just fine, the only action needed is to reindex es.

Do I need to add some restrictions? Like not allowing new fields for example. Personally, I wouldn't add it because this config is set up by devs.

Usage example:

`settings.py`
```
...
SCHEMA_UPDATE = {
    'archive': {
        'extra': {
            'type': 'dict',
            'schema': {},
            'mapping': {
                'type': 'object',
                'enabled': True
            },
            'allow_unknown': True,
        }
    }
}
...
```

### DIFF
--- a/superdesk/resource.py
+++ b/superdesk/resource.py
@@ -125,6 +125,12 @@ class Resource():
                 # used in app:initialize_data
                 endpoint_schema['mongo_indexes__init'] = self.mongo_indexes
 
+            if app.config.get('SCHEMA_UPDATE', {}).get(self.endpoint_name):
+                schema_updates = app.config['SCHEMA_UPDATE'][self.endpoint_name]
+                log.warning('Updating {} schema with custom data for fields: {}'.format(
+                    self.endpoint_name.upper(), ', '.join(schema_updates.keys())))
+                self.schema.update(schema_updates)
+
         self.endpoint_schema = endpoint_schema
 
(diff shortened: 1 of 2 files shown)


jerome-poisson:
@ride90 I don't think restrictions are necessary on `setup.py`, indeed this is for dev or admins.

MarkLark86:
@ride90 I would agree, the restrictions are not necessary. At the same time though it would be ideal to have restrictions so the dev/admin does not break their instance (we should not allow another dev/admin to break the system).

What if we were to provide a warning rather than an info message in the log that the schema was changed (that's middle ground I suppose)?

superdesk/resource.py
```diff
@@ -13,6 +13,8 @@
 import superdesk
 from eve.utils import config
 
+from flask import current_app as app
```
petrjasek, line 16: you have this passed to `__init__`  as param so it's not used imo

ride90:
At the end `warning` level was used, since we `configure_logging` after the resources are registered and the default level for `superdesk` logger is `info`, so the message simply will not be visible.

PR #1879, opened 2020-05-22
FIX: Users should not be able to edit dictionaries to an existing name or code and save. [SDESK-5278]

GyanP:


apps/dictionaries/service.py
```diff
@@ -156,11 +156,10 @@ def on_create(self, docs):
                 pass
 
         for doc in docs:
-            if self.find_one(req=None, name=doc['name'],
-                             language_id=doc['language_id'],
-                             type=doc.get('type', DictionaryType.DICTIONARY.value)):
+            if self.is_valid_dictonary(doc):
```
petrjasek, line 159: I think that method should be renamed, it's a bit confusing that when `is_valid_dictionary` returns True you raise exception. maybe `is_duplicate_dictionary` or something

PR #1895, opened 2020-06-04
avoid stage visibility update during testing

petrjasek:


ride90:
@petrjasek , can you post more description for this PR? It's not fully clear what's the issue.

petrjasek:
when running e2e tests we prepopulate data - adding desks/stages/users, and when any of these is created we update list of stages user can/can't see so we can do queries faster.
with this one it updates that list of invisible stages only once when prepopulate is done and all users/desks/stages are in place.

PR #1896, opened 2020-06-05
FIX: Wrong POI data for renditions in NINJS. [SDESK-5301]

GyanP:


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -322,6 +322,14 @@ def _format_related(self, article, subscriber):
                     item = orig_item.copy()
 
                 item = self._transform_to_ninjs(item, subscriber, recursive=False)
+
+                # Keep original POI and get rid of all other POI.
+                renditions = item.get('renditions')
+                if renditions:
+                    for rendition in list(renditions):
```
petrjasek, line 329: not sure if this is an idiom, but it would be a bit more readable for me with:
```py
for rendition in renditions.keys():
```

superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -322,6 +322,14 @@ def _format_related(self, article, subscriber):
...
+                        if rendition != 'original' and renditions[rendition].get('poi'):
```
petrjasek, line 330: I think there might be cases when `renditions[rendition]` is `None` if image is too small for some custom one so it would be good to check it

PR #1901, opened 2020-06-17
add migration to introduce default content profiles

petrjasek:
it populates text/audio/video/picture/composite content types via init_data
and returns those as a profile if profile is not set on an item (eg. profile=picture).

SDESK-5217 SDESK-5157


apps/archive/common.py
```diff
@@ -204,7 +213,11 @@ def on_create_item(docs, repo_type=ARCHIVE):
             # set the source for the article
             set_default_source(doc)
 
-        if 'profile' not in doc and app.config.get('DEFAULT_CONTENT_TYPE', None):
+        ignore_profiles = DEFAULT_PROFILES.copy()
+        ignore_profiles.add(None)
+
+        if doc.get('profile') in ignore_profiles and doc.get('type') == 'text' and \
+                app.config.get('DEFAULT_CONTENT_TYPE', None):
             doc['profile'] = app.config.get('DEFAULT_CONTENT_TYPE', None)
```
ride90, line 221: just to make it easier to understand: `app.config['DEFAULT_CONTENT_TYPE']`
Since at this point `DEFAULT_CONTENT_TYPE` is set anyway.

apps/archive/common.py
```diff
@@ -630,6 +643,9 @@ def handle_existing_data(doc, pub_status_value='usable', doc_type='archive'):
         if doc_type == 'archive' and not is_flag_in_item(doc, 'marked_for_not_publication'):
             set_flag(doc, 'marked_for_not_publication', False)
 
+        if doc.get('type'):
+            doc.setdefault('profile', doc['type'])
```
ride90, line 647: Method docstring is missed

apps/content_types/content_types.py
```diff
@@ -195,11 +196,14 @@ def _update_template_fields(self, updates, original):
...
 
     def find_one(self, req, **lookup):
+        is_edit = req and 'edit' in req.args
         doc = super().find_one(req, **lookup)
-        if doc and req and 'edit' in req.args:
+        if doc and is_edit:
             prepare_for_edit_content_type(doc)
         if doc:
             clean_doc(doc)
+        if not doc and not is_edit and lookup.get('_id') in ('text', 'picture', 'audio', 'video', 'composite'):
```
ride90, line 205: we can use `DEFAULT_PROFILES` here instead of `('text', 'picture', 'audio', 'video', 'composite')`

superdesk/data_updates/00024_20200615-153837_content_types.py
```diff
@@ -0,0 +1,49 @@
...
+    resource = 'content_types'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        # set type for existing profiles to text
+        print(
+            'updated existing profiles',
+            mongodb_collection.update_many({'item_type': None}, {'$set': {'item_type': 'text'}}).modified_count,
+        )
+        # generate new types based on core conf
+        for item_type in ('text', 'audio', 'video', 'picture', 'composite'):
```
ride90, line 29: also here `DEFAULT_PROFILES`

MarkLark86 (approved):

PR #1903, opened 2020-06-18
[DESKS] new endpoint to get overview

jerome-poisson:
by checking `desks/DESK_ID/stages_overview`, a count of items per stage can be
retrieved.
If `all` is used for `DESK_ID`, the a count of items for all desks is
retrieved.

SDESK-5208

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -11,13 +11,14 @@ import json
 import itertools
 
 import superdesk
-from flask import current_app as app
+from flask import current_app as app, request
 
 from superdesk import get_resource_service
 from superdesk.errors import SuperdeskApiError
 from superdesk.resource import Resource
 from superdesk import config
 from superdesk.utils import SuperdeskBaseEnum
+from superdesk.timer import timer
 from bson.objectid import ObjectId
 from superdesk.services import BaseService
 from superdesk.notification import push_notification
@@ -135,6 +136,9 @@ def init_app(app):
     endpoint_name = 'sluglines'
     service = SluglineDeskService(endpoint_name, backend=superdesk.get_backend())
     SluglineDesksResource(endpoint_name, app=app, service=service)
+    endpoint_name = 'stages_overview'
+    service = StagesOverviewService(endpoint_name, backend=superdesk.get_backend())
+    StagesOverviewResource(endpoint_name, app=app, service=service)
 
 
 superdesk.privilege(name='desks', label='Desk Management', description='User can manage desks.')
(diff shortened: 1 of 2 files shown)


apps/desks.py
```diff
@@ -537,6 +540,24 @@ def _find_other_sluglines(self, family_id, slugline, versioncreated, desk_id):
...
+    resource_methods = ['GET']
+
+
+class OverviewDesksService(BaseService):
+
+    def on_fetched(self, doc):
+        desk_id = request.view_args['desk_id']
+        desk_match = {"$ne": None} if desk_id == "all" else ObjectId(desk_id)
+        aggregation = app.data.pymongo().db.archive.aggregate([
+            {"$match": {"task.desk": desk_match}},
```
petrjasek, line 555: it might need some state filter here, otherwise it also counts published items (which might be fine, just not sure)

apps/desks.py
```diff
@@ -537,6 +540,24 @@ def _find_other_sluglines(self, family_id, slugline, versioncreated, desk_id):
...
+    resource_title = "desk_overview"
+    resource_methods = ['GET']
+
+
+class OverviewDesksService(BaseService):
+
+    def on_fetched(self, doc):
+        desk_id = request.view_args['desk_id']
+        desk_match = {"$ne": None} if desk_id == "all" else ObjectId(desk_id)
+        aggregation = app.data.pymongo().db.archive.aggregate([
```
petrjasek, line 554: can you use a [timer](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/timer.py#L11) here?
I wonder how much time such aggregation will take

MarkLark86: Why use Mongo instead of Elasticsearch for aggregations? Elasticsearch is much more equipped for this form of query

apps/desks.py
jerome-poisson: I was counting published filter on purpose, but you're right that I don't if it's need. @pavlovicnemanja should we count them in the summary?

pavlovicnemanja: image

We need total items per stage. Grand Total of all stages in desk can be done on frontend part.

apps/desks.py
jerome-poisson: @petrjasek you want that I log the query time for each query? Because on my local instance it won't tell much, I have just a few items it will be really quick. But logging the query time can be interesting on bigger instances.

@MarkLark86 right ES may be more appropriate, I'll update it.

petrjasek: yep I would like to check it later on some bigger instances

jerome-poisson: @MarkLark86 @petrjasek I've updated the PR with timer and it now uses ES

I've measured aggregation with Mongo and ES for comparison, and it seems that on my small local instance, Mongo is a lot faster:

```
overview aggregation mongo '5ef4b152049b9aa6829e8bab': 1.328ms
overview aggregation ES '5ef4b152049b9aa6829e8bab': 7.539ms
overview aggregation mongo 'all': 1.546ms
overview aggregation ES 'all': 6.509ms
```
So I'm not if the move to ES is a good one, we may need to check on bigger instances.

jerome-poisson:
@pavlovicnemanja you can work with this PR, I'm waiting for review to merge it, and I may switch back to mongo if performance prove to better there, but I should keep the API.

apps/desks.py
jerome-poisson: so to be clear, it's including published items right? Nothing to change then.

apps/desks.py
MarkLark86: Elasticsearch will start to show it's benefits with large data sets. I have witnessed aggregations with the analytics module on 1 years worth of data at AAP complete in milliseconds. Mongo is good at storage, and Elastic is good for searching and aggregations. If Elasticsearch is enabled for a resource, we should always use it

jerome-poisson: @MarkLark86 alright makes sense. Thanks for the feedback, I stay with ES then.

features/desks.feature
```diff
@@ -410,3 +410,66 @@ Feature: Desks
...
+            }
+        """
+        When we get "/desks/all/stages_overview"
+        Then we get existing resource
+        """
+            {
+                "_items": [
+                    {
+                        "stage": "#desks.incoming_stage#",
+                        "count": 3
```
petrjasek, line 467: just wondering why is the count here different compared to the previous query? isn't it the same stage id?

jerome-poisson: previous query is filtering on `Sports desk` while this one gather all desks at once, including the 2 items in `incoming` stage from `Politics desk` (note the querying URL).

petrjasek: I see, it's actually single stage used on 2 desks

PR #1906, opened 2020-06-22
[SDBELGA-308] Fix twitter rate limit error

idsvn:
- Use [Application only auth](https://developer.twitter.com/en/docs/basics/authentication/oauth-2-0/application-only) to have more req per 15 minutes (1500 vs 900 for get user timeline)
- Raise error when rate limited exceeded (code 88)

SDBELGA-308

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ install_requires = [
     'mongolock>=1.3.4,<1.4',
     'PyYAML>=4.2b1,<5.0',
     'lxml>=3.8,<4.2',
-    'python-twitter==3.3',
+    'python-twitter==3.5',
     'chardet<4.0',
     'pymongo>=3.8,<3.9',
     'croniter<0.4',
--- a/superdesk/io/feeding_services/twitter.py
+++ b/superdesk/io/feeding_services/twitter.py
@@ -12,7 +12,7 @@ import re
 import hashlib
 from dateutil import parser
 
-from superdesk.errors import IngestTwitterError, IngestApiError
+from superdesk.errors import IngestTwitterError
 from superdesk.io.registry import register_feeding_service
 from superdesk.io.feeding_services import FeedingService
 from superdesk.metadata.utils import generate_guid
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -44,7 +44,7 @@
     'mongolock>=1.3.4,<1.4',
     'PyYAML>=4.2b1,<5.0',
     'lxml>=3.8,<4.2',
-    'python-twitter==3.3',
+    'python-twitter==3.5',
```
idsvn, line 47: v3.3 has **App auth** but does not compatible with python 3, see [commit](https://github.com/bear/python-twitter/commit/3ee3c5a55b903e0bb62d3388c0ed594f45de5b35)
v3.5 has [fix typo](https://github.com/bear/python-twitter/commit/eed985f1cb5cd3dadfbd42732bb02c319a4f4653) that cause `'set' object does not support indexing`. This also because sometime twitter api return 0, not sure why and we could not reproduce it

superdesk/io/feeding_services/twitter.py
```diff
@@ -103,6 +100,11 @@ def _update(self, provider, update, test=False):
                 elif exc.message[0].get('code') == 32:
                     # invalid credentials
                     raise IngestTwitterError.TwitterLoginError(exc, provider)
+                elif exc.message[0].get('code') == 88:
+                    # rate limit exceeded
+                    raise IngestTwitterError.TwitterRateLimitError(exc, provider)
+                else:
+                    raise IngestApiError.apiGeneralError(exc, provider)
```
idsvn, line 107: Should we add TwitterAPIGeneralError instead of using apiGeneralError from IngestApiError?

superdesk/io/feeding_services/twitter.py
```diff
@@ -43,11 +43,11 @@ class TwitterFeedingService(FeedingService):
         },
         {
             'id': 'access_token_key', 'type': 'text', 'label': 'Twitter Access Token Key',
-            'placeholder': 'Twitter access_token_key', 'required': True
+            'placeholder': 'Twitter access_token_key', 'required': False
         },
         {
             'id': 'access_token_secret', 'type': 'password', 'label': 'Twitter Access Token Secret',
-            'placeholder': 'Twitter access_token_secret', 'required': True
+            'placeholder': 'Twitter access_token_secret', 'required': False
```
idsvn, line 50: No longer use, make it optional in configuration instead of removing it to avoid confusing. Should we remove it?

superdesk/io/feeding_services/twitter.py
ride90: Since this API error is twitter related I would add `TwitterAPIGeneralError `

superdesk/io/feeding_services/twitter.py
ride90: If it's not used anymore it's better rather remove it to avoid confusion.
I assume this update should not break already running twitter ingests, right?

idsvn: Yes, There is **NO** breaking changes, we have run it for a few days and the old twitter ingests are working fine

ride90:
Hi @idsvn , please add TwitterAPIGeneralError.

PR #1910, opened 2020-06-25
[DESKS] added assignments aggregation type for desk overview

jerome-poisson:
Assignments states can now be aggregated for desk overview in the same
way as for stages. Because the aggregation is really similar, the
endpoint has been refactorised to be unique, client must now use the
`desks/[DESK_ID]/overview/[aggregation_type]` endpoint where
`[aggregation_type]` can be either `stages` or `assignments`.
`[DESK_ID]` can have the special `all` value to get all desks at once.

SDESK-5355

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -136,9 +136,9 @@ def init_app(app):
     endpoint_name = 'sluglines'
     service = SluglineDeskService(endpoint_name, backend=superdesk.get_backend())
     SluglineDesksResource(endpoint_name, app=app, service=service)
-    endpoint_name = 'stages_overview'
-    service = StagesOverviewService(endpoint_name, backend=superdesk.get_backend())
-    StagesOverviewResource(endpoint_name, app=app, service=service)
+    endpoint_name = 'desk_overview'
+    service = OverviewService(endpoint_name, backend=superdesk.get_backend())
+    OverviewResource(endpoint_name, app=app, service=service)
 
 
 superdesk.privilege(name='desks', label='Desk Management', description='User can manage desks.')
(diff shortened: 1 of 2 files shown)


petrjasek:
seems like it's pretty same like the other one, what about making the aggregation a param on a single endpoint?

jerome-poisson:
@petrjasek yes I've been hesitating to do that. OK so I'll factorize the two once the other one is merged.

ride90:
@jerome-poisson , I agree with @petrjasek , it looks exactly the same as the one for Stages, so merging them into one is the way to go.

jerome-poisson:
yes that's the plan

jerome-poisson:
@petrjasek @ride90 factorisation is done.
@pavlovicnemanja the API is the same except that the URI is slightly modified, see the commit message for details.

petrjasek:
just wondering about the statuses - isn't it a planning thing?

jerome-poisson:
> just wondering about the assignment statuses - isn't it a planning thing?

@pavlovicnemanja we've had a little chat about this with @petrjasek , and it's not perfectly clear.

Do you need planning data or superdesk items like this is currently done?

pavlovicnemanja:
@jerome-poisson hey yes, only assignments part is from planning. Others are from superdesk.

jerome-poisson:
@petrjasek @pavlovicnemanja I've updated to use assignments for planning. I've been hesitating to put this in planning instead of core, but at the end I've chosen to keep the unique endpoint with desk (as it is still a desk overview even if it concern a planning data).

PR #1911, opened 2020-06-26
send list of registered transmission services to client

petrjasek:
SDCP-247

### DIFF
--- a/superdesk/publish/__init__.py
+++ b/superdesk/publish/__init__.py
@@ -25,6 +25,7 @@ logger = logging.getLogger(__name__)
 
 registered_transmitters = {}
 transmitter_errors = {}
+registered_transmitters_list = []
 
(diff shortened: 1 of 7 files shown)


superdesk/publish/__init__.py
```diff
@@ -35,6 +36,11 @@
 def register_transmitter(transmitter_type, transmitter, errors):
     registered_transmitters[transmitter_type] = transmitter
     transmitter_errors[transmitter_type] = dict(errors)
+    registered_transmitters_list.append({
+        'type': transmitter_type,
+        'name': transmitter.NAME or transmitter_type,
```
ride90, line 41: I would use `getattr(transmitter, 'NAME', transmitter_type)` just in case `NAME` does not exist as attr

PR #1914, opened 2020-06-29
Create a macro to be used in internal destinations for translations. [SDCP-231]

GyanP:


apps/duplication/archive_translate.py
```diff
@@ -92,7 +92,11 @@ def _translate_item(self, guid, language, task=None, service=None, **kwargs):
             item['task'] = task
 
         extra_fields = ['translation_id', 'translated_from']
-        _id = archive_service.duplicate_item(item, extra_fields=extra_fields, operation='translate')
+
+        if state:
+            _id = archive_service.duplicate_item(item, extra_fields=extra_fields, state=state, operation='translate')
+        else:
+            _id = archive_service.duplicate_item(item, extra_fields=extra_fields, operation='translate')
```
petrjasek, line 99: imo just:
```py
id = archive_service.duplicate_item(item, extra_fields=extra_fields, state=state, operation='translate')
```
should be enough, that `duplicate_item` method handles state being `None`

PR #1917, opened 2020-06-30
FIX: Issues with editing images. [SDESK-5369]

GyanP:


superdesk/media/media_editor.py
```diff
@@ -39,6 +39,10 @@ class MediaEditorResource(Resource):
             'type': 'dict',
             'required': True,
             'mapping': not_analyzed
+        },
+        'renditions': {
+            'type': 'dict',
+            'mapping': not_analyzed
```
petrjasek, line 45: I'm not sure about mapping here, it's not elastic resource anyhow so it shouldn't be used

GyanP: I'm also not sure, why it was done? Should I remove it for all?

petrjasek: I think so

GyanP: Ok.

PR #1921, opened 2020-07-02
FIX: Unsaved data is lost when users navigate away or minimize the article. [SDESK-5356]

GyanP:
There is an unknown fields errors `{"target_types": "unknown field", "target_subscribers": "unknown field", "target_regions": "unknown field"}`

### DIFF
--- a/apps/archive/common.py
+++ b/apps/archive/common.py
@@ -90,34 +90,6 @@ ARCHIVE_SCHEMA_FIELDS = {
         'allowed': item_operations,
         'index': 'not_analyzed'
     },
-    'target_regions': {
-        'type': 'list',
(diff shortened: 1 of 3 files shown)


GyanP:
> I think you can remove it from https://github.com/superdesk/superdesk-core/blob/develop/apps/archive/common.py#L93-L120

I can't remove it because this has been used in `ContentTemplatesApplyResource`, https://github.com/superdesk/superdesk-core/blob/develop/apps/templates/content_templates.py#L392

petrjasek:
it's still using that basic schema there https://github.com/superdesk/superdesk-core/blob/develop/apps/archive/common.py#L708

PR #1922, opened 2020-07-06
As a CP editor I want my 'Dateline' to be populated based on the Geonames service. [SDCP-254]

GyanP:
requires https://github.com/superdesk/superdesk-cp/pull/51


superdesk/places/places_autocomplete.py
```diff
@@ -37,7 +37,7 @@ def get(self, req, lookup):
         assert req.args.get('name'), {'name': 1}
         params = [
             ('name_startsWith', req.args.get('name')),
-            ('lang', req.args.get('lang')),
+            ('lang', 'fr' if req.args.get('lang') == 'fr-CA' else req.args.get('lang')),
```
petrjasek, line 40: imo would be good to use something generic like `lang.split('-')[0]` so it's not just french

PR #1929, opened 2020-07-14
FIX: language is reset in metadata section of the template when making changes in parent content profile. [SDCP-265]

GyanP:
We keep updating the template with the new schema without checking if there is any metadata value present before.

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -180,6 +180,8 @@ class ContentTypesService(superdesk.Service):
         Finds the templates that are referencing the given
         content profile an clears the disabled fields
         """
+
+        # these are the only fields of templates that don't depend on the schema.
         template_metadata_fields = ['language', 'usageterms']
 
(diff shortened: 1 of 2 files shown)


GyanP:
> can you write a test for this pls?

Ya sure.

apps/content_types/content_types.py
```diff
@@ -180,6 +180,8 @@ def _update_template_fields(self, updates, original):
         Finds the templates that are referencing the given
         content profile an clears the disabled fields
         """
+        template_metadata_fields = ['language', 'usageterms']
```
petrjasek, line 183: why is it just those 2 fields which are affected?

GyanP: Because these are the only fields of templates that don't depend on the schema.

petrjasek: can you add some comment there regarding it? just to make it clear in the code.

GyanP: Sure, I'll.

PR #1930, opened 2020-07-14
make user.username pattern configurable

petrjasek:


petrjasek:
it's when we use azure and we get external username

PR #1936, opened 2020-07-17
re-implement word counting to work with unicode letters

tomaskikutis:
SDESK-5352

### DIFF
--- a/tests/io/feed_parsers/nitf_tests.py
+++ b/tests/io/feed_parsers/nitf_tests.py
@@ -88,7 +88,7 @@ class AAPTestCase(NITFTestCase):
         self.assertEqual(self.item.get('anpa_category')[0]['qcode'], 'a')
 
     def test_word_count(self):
-        self.assertEqual(349, self.item.get('word_count'))
+        self.assertIsInstance(self.item.get('word_count'), int)
 
 
 class APExampleTestCase(NITFTestCase):
@@ -105,7 +105,7 @@ class APExampleTestCase(NITFTestCase):
         self.assertEqual(self.item.get('ednote'), 'For global distribution')
 
     def test_word_count(self):
-        self.assertEqual(1063, self.item.get('word_count'))
+        self.assertIsInstance(self.item.get('word_count'), int)
 
 
 class IPTCExampleTestCase(NITFTestCase):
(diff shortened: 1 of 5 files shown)


ride90:
@tomaskikutis check failing tests and linter

tests/io/feed_parsers/pa_nitf_test.py
```diff
@@ -54,7 +54,7 @@ def test_slugline(self):
         self.assertEqual(self.item.get('format'), 'HTML')
         self.assertEqual(4, len(self.item.get('subject')))
         self.assertIn('Trivia (Oct 14)', self.item.get('keywords'))
-        self.assertEqual(637, self.item.get('word_count'))
+        self.assertTrue(type(self.item.get('word_count')) is int)
```
petrjasek, line 57: you can do `self.assertIsInstance(self.item.get('word_count'), int)`

tests/io/feed_parsers/nitf_tests.py
```diff
@@ -88,7 +88,7 @@ def test_anpa_category(self):
         self.assertEqual(self.item.get('anpa_category')[0]['qcode'], 'a')
 
     def test_word_count(self):
-        self.assertEqual(349, self.item.get('word_count'))
+        self.assertIsInstance(self.item.get('word_count'), int)
```
ride90, line 91: Checking the type is ok, but we also must ensure the count equality.

tomaskikutis: The question is what is the correct number? It's not practical to go over hundreds of words to count it manually in order to be able to compare the results. There are also various ways to count formatting. Even if you counted words manually it's likely that you would get a different result than it was specified in the assert. That's why I prefer to stick to specific scenarios in `text_utils.py`.

ride90: In this case what all these test_word_count tests do?

ride90: ok, let's keep it as it is

tomaskikutis: > In this case what all these test_word_count tests do?

I would remove them, but kept just to check that there's some number populated. Not sure if it's very useful.

PR #1939, opened 2020-07-21
Users that are disabled or deactivated should not appear in 'mark for user' list. [SDBELGA-404]

GyanP:


superdesk/users/services.py
```diff
@@ -363,6 +363,22 @@ def get(self, req, lookup):
...
+            '$and': [
+                {'is_active': True},
+                {'is_enabled': True},
+            ]
+        }
+
+        if req and not params:
+            req.where = json.dumps(query)
+        elif req and params and params.get('all'):
+            req.where = None
```
petrjasek, line 380: not sure you should reset this, that might not allow you to search within all users, but just omitting that query should be enough

ride90: I would not touch a user service at all. We can specify it on the client-side when we query a list of users for marking.

petrjasek: the idea was to return only active users by default and those not activated/disabled only when explicitly requested so we don't have to filter the users all over the client

superdesk/users/services.py
```diff
@@ -363,6 +363,19 @@ def get(self, req, lookup):
                 lookup['is_author'] = bool(int(is_author))
             else:
                 logger.warn('bad value of is_author argument ({value})'.format(value=is_author))
+
+        """filtering out inactive users and disabled users"""
+
+        params = json.loads(request.args.get('where')) if request and request.args.get('where') else {}
+
+        # Filtering inactive users
+        if not req.args.get('show_inactive') and params.get('is_active', True):
```
petrjasek, line 372: not sure why you need the `params` here, wouldn't just `if not reg.args.get('show_inactive'):` do?

GyanP: > wouldn't just if not reg.args.get('show_inactive'): do?

No, the reason for this is if we want only the inactive_user then we will hit the URL as `users?where=is_active: false`, so now if we check only for the `if not reg.args.get('show_inactive'):` and not for params then this condition will get true and it will assign lookup['is_active'] as True and we will not get the result which we want.

petrjasek: I think if we want inactive it should do `/users?where=is_active:false&show_inactive=true`, this `show_inactive` should be always used on pages like user management just `where` would be changing

superdesk/users/services.py
```diff
@@ -363,6 +363,19 @@ def get(self, req, lookup):
                 lookup['is_author'] = bool(int(is_author))
             else:
                 logger.warn('bad value of is_author argument ({value})'.format(value=is_author))
+
+        """filtering out inactive users and disabled users"""
+
+        params = json.loads(request.args.get('where')) if request and request.args.get('where') else {}
```
petrjasek, line 369: I think now you don't need this

PR #1940, opened 2020-07-22
Sdesk 5362 endpoint sam storage destination sets

dev-tsoni:
Add sams proxy endpoints in superdesk-core for storage destinations and sets, using flask blueprints.


apps/auth/auth.py
```diff
@@ -108,6 +108,9 @@ def check_permissions(self, resource, method, user):
         if not user:
             return True
 
+        if resource == '_blueprint':
```
MarkLark86, line 111: Can you add some comments into `Step 1` in this function's docstring explaining this auth condition

superdesk/sams/decorator.py
```diff
@@ -0,0 +1,14 @@
+from functools import wraps
+from flask import request
+
+
+def blueprint_auth(app):
```
MarkLark86, line 5: Can you add some comments here as to what this decorator is used for.
Also, this file should probably reside under `superdesk/auth/decorator.py` so all blueprints have access to this generic auth functionality

superdesk/sams/decorator.py
```diff
@@ -0,0 +1,14 @@
+from functools import wraps
+from flask import request
```
MarkLark86, line 2: You can also use add `current_app as app` here so the decorator doesn't require the app to be passed in

superdesk/sams/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+        'HOST': app.config.get('SAMS_HOST'),
+        'PORT': app.config.get('SAMS_PORT')
+    }
+    client = SamsClient(configs)
+
+    @destinations_bp.before_request
+    @sets_bp.before_request
+    @blueprint_auth(app)
+    def before_request():
+        """ Add authentication before request to all blueprint """
```
MarkLark86, line 21: Lint error here: `./superdesk/sams/__init__.py:21:1: D210 No whitespaces allowed surrounding docstring text`

apps/auth/auth.py
dev-tsoni: Added

superdesk/sams/decorator.py
dev-tsoni: Comments added and file moved to superdesk/auth/decorator.py

superdesk/sams/decorator.py
dev-tsoni: Added

superdesk/sams/__init__.py
dev-tsoni: Resolved

superdesk/default_settings.py
```diff
@@ -364,7 +367,7 @@ def local_to_utc_hour(hour):
 ]
 
 #: Specify what modules should be enabled
-INSTALLED_APPS = []
+INSTALLED_APPS = ['superdesk.sams']
```
ride90, line 370: Are we going to have SAMS enabled by default?

dev-tsoni: Missed it, have removed 'superdesk.sams' from INSTALLED_APPS. If sams is enabled this will be added in settings.py at superdesk/server in customer repository.

PR #1941, opened 2020-07-23
[AUTH] [FEEDING_SERVICE]: fixed Google Auth login + GMail feeding ser…

jerome-poisson:
…vice

Google Auth login was using a deprecated library (`flask-oauthlib`)
which was not compatible anymore with recent versions of Flask. This
patch fixes it by using authlib.

Furthermore, Google is deactivating traditional login/pass for IMAP with
GMail, thus this patch implement a new "gmail" feeding service which
uses OAuth for log-in.

SDESK-5200


superdesk/io/registry.py
```diff
@@ -161,10 +162,21 @@ def service(service_id):
             if restricted_parsers is not None:
                 restricted_parsers = list(restricted_parsers.keys())
 
+            fields = getattr(feeding_service_class, 'fields', [])
+            # we need to transtype LazyString to str before serialization
+            for idx, field in enumerate(fields):
+                copied = False
+                for key, value in field.items():
+                    if isinstance(value, LazyString):
```
jerome-poisson, line 170: this can probably be avoided by using a `default=str` in the JSON encoder.

superdesk/auth/oauth.py
```diff
@@ -24,53 +24,197 @@
...
+class OAuth2TokenService(BaseService):
+    """Store OAuth tokens"""
+
+
+def token2dict(
+    identifier: str,
+    email: str,
+    token: OAuth2Token,
+    name: str = "google",
+) -> dict:
```
ride90, line 110: I think it's a time to add a type checker, afaik we don't use any right now.

superdesk/io/feeding_services/gmail.py
```diff
@@ -0,0 +1,93 @@
...
+        self._update(provider, update=None, test=True)
+
+    def authenticate(self, provider: dict, config: dict) -> imaplib.IMAP4_SSL:
+        oauth2_token_service = superdesk.get_resource_service('oauth2_token')
+        token = oauth2_token_service.find_one(req=None, name='google', identifier=provider['name'])
+        if token is None:
+            raise IngestEmailError.notConfiguredError(ValueError(l_("You need to log in first")), provider=provider)
+        imap = imaplib.IMAP4_SSL('imap.gmail.com')
+
+        if token['expires_at'].timestamp() < time.time() - 600:
```
ride90, line 76: shouldn't it be: token['expires_at'].timestamp() < time.time() **+(plus)** 600
?

superdesk/auth/oauth.py
jerome-poisson: indeed, we should evaluate those available, but `mypy` is a pretty popular choice

superdesk/io/feeding_services/gmail.py
jerome-poisson: oh right, good catch, thanks!

jerome-poisson: fixed

superdesk/auth/oauth.py
```diff
@@ -24,53 +24,197 @@
...
 
+logger = logging.getLogger(__name__)
 
 bp = superdesk.Blueprint('oauth', __name__)
+oauth = None
+REFRESH_TOKEN_URL = "https://oauth2.googleapis.com/token"
 
 
 def init_app(app):
+    global oauth
```
petrjasek, line 50: just wondering why having it global?

jerome-poisson: it's used below in `configure_google` and `refresh_google_token` (like at https://github.com/superdesk/superdesk-core/pull/1941/files/5b3ea4f737544c2dc2ca9b60b0adef8bc0f95375#diff-2498d9983039e50bffec686ea374aca9R216).

I've been thinking about putting everything in a class which would be cleaner, but the flask decorators were not too happy about that, and I've not spent too much time checking what was going wrong, as it's was not bad to stay close to the existing code, to limit the diff.

superdesk/auth/oauth.py
jerome-poisson: resolving it as you're integrating mypy in a separate ticket

jerome-poisson:
@ride90 @petrjasek it seems that this PR won't have review before my holidays, so feel free to merge it if it's fine, thanks!

superdesk/io/registry.py
petrjasek: it should be handled now by [SuperdeskJSONEncoder](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/json_utils.py#L15)

superdesk/auth/oauth.py
```diff
@@ -24,53 +24,197 @@
...
+        else:
+            extra_scopes = []
+            refresh = False
+
+        configure_google(app, extra_scopes=extra_scopes, refresh=refresh)
+
+
+class OAuth2TokenResource(Resource):
+    schema = {
+        'identifier': {
```
petrjasek, line 72: wouldn't `_id` work for that? it has a unique index on it so that's a plus

jerome-poisson: yes it would be better indeed, I need to modify client for that, and I have realised that I don't have ingest provider ID when the `LOG-IN WITH GMAIL` button is pressed, and also that name can be missing at this point (it's mandatory, but when the button is clicked, the ingest is not yet saved).
I'm trying to find a workaround.

superdesk/auth/oauth.py
```diff
@@ -24,53 +24,199 @@
...
```
petrjasek, line 72: seems like this is not used anymore

superdesk/auth/oauth.py
```diff
@@ -24,53 +24,199 @@
...
+
+def token2dict(
+    token_id: str,
+    email: str,
+    token: OAuth2Token,
+    name: str = "google",
+) -> dict:
+    """Convert authlib's OAuth2Token to a dict usable with datalayer
+
+    :param identifier: name used to associate the token with a provider
```
petrjasek, line 113: this param is not matching

superdesk/io/feeding_services/gmail.py
```diff
@@ -0,0 +1,93 @@
...
+            'id': 'email',
+            'type': 'string',
+            'label': l_('email'),
+            'readonly': True,
+            'placeholder': l_("This field will be automatically filled once you've logged using log-in button above"),
+        },
+        {
+            'id': 'mailbox', 'type': 'text', 'label': l_('Mailbox'),
+            'default_value': 'INBOX',
+            'placeholder': 'Mailbox', 'required': True,
```
petrjasek, line 51: we should probably translate placeholders as well

petrjasek (approved):

ride90 (approved):

PR #1946, opened 2020-07-29
Global search shows 10000 in item count even if there is more items

ride90:
SDESK-5430


petrjasek:
I think we might also need this on ingest/archive/published when using just single source, maybe we should rather have a config in eve-elastic and set this globally? also planning is affected I suppose

ride90:
@petrjasek I think eve-elastic config would be the cleanest. I will PR it.

PR #1947, opened 2020-07-30
handle publishing of binaries via ftp

petrjasek:
for other format types than ninjs

### DIFF
--- a/superdesk/media/renditions.py
+++ b/superdesk/media/renditions.py
@@ -392,7 +392,7 @@ def get_rendition_file_name(rendition):
     :param rendition:
     :return:
     """
-    media_id = str(rendition.get('media'))
+    media_id = str(rendition['media'])
     ext = os.path.splitext(media_id)[-1]
     return media_id.replace('/', '-') + (
         guess_media_extension(rendition.get('mimetype', '')) if not ext else '')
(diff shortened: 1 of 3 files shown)


superdesk/media/renditions.py
```diff
@@ -392,6 +392,7 @@ def get_rendition_file_name(rendition):
     :param rendition:
     :return:
     """
-    ext = os.path.splitext(rendition.get('media'))[-1]
-    return rendition.get('media').replace('/', '-') + (
+    media_id = str(rendition.get('media'))
```
ride90, line 395: I believe that rendition always has a `media`, but in theory, `media_id` might be `"None"`(string) here.

tests/publish/transmitters/ftp_transmitter_tests.py
```diff
@@ -69,68 +132,65 @@ def test_it_can_connect(self):
...
+                'item_id': 'someid',
+                '_current_version': 3,
+                'associations': ASSOCIATIONS,
+            }
+        ])
+
+        subscriber = {}
+        service._transmit(queue_item, subscriber)
+
+        print('mock', ftp_mock)
```
ride90, line 194: print

superdesk/media/renditions.py
petrjasek: yep I think so to, so will make it explicit not using `.get()`

PR #1952, opened 2020-08-03
counting words - count whitespace as zero words

tomaskikutis:
SDESK-5352


tests/text_utils_test.py
```diff
@@ -5,6 +5,9 @@
 
 class WordCountTestCase(unittest.TestCase):
 
+    def test_word_count_whitespace_string():
```
ride90, line 8: should be `def test_word_count_whitespace_string(self)`

PR #1955, opened 2020-08-06
[TEXT CHECKERS] new `ai` category with iMatrics autotagger

jerome-poisson:
`ai` category in text checkers is used to groups services doing article
analysis using machine learning or other AI fields.

iMatrics is an autotagging service using machine learning.

SDNTB-643

### DIFF
--- a/superdesk/text_checkers/ai/__init__.py
+++ b/superdesk/text_checkers/ai/__init__.py
@@ -1,0 +1,83 @@
+# Copyright 2020 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+from superdesk.resource import Resource
+from superdesk.services import BaseService
+from superdesk.errors import SuperdeskApiError
+from .. import tools
+from .base import registered_ai_services, AIServiceBase
+import superdesk
+
+AUTO_IMPORT = True
+
+
+class AIResource(Resource):
+    schema = {
+        'service': {
+            'type': 'string',
+            'required': True,
+        },
+        'item_id': {
+            'type': 'string',
+            'required': True,
+        },
+    }
+    datasource = {
+        'projection': {'analysis': 1}
+    }
+    internal_resource = False
+    resource_methods = ['POST']
+    item_methods = ['GET']
+
+
+class AIService(BaseService):
+    r"""Service managing article analysis with machine learning/AI related services
+
+    When doing a POST request on this service, the following keys can be used (keys with a \* are required):
+
+
+    ==========  ===========
+    key         explanation
+    ==========  ===========
+    service \*  name of the service to use
+    item_id \*  _id of the item in archive collection
+    ==========  ===========
+
+    e.g. to get autotagging with iMatrics service:
+
+    .. sourcecode:: json
+
+        {
+            "service": "imatrics",
+            "item_id": "some_id"
+        }
+
+    """
+
+    def create(self, docs, **kwargs):
+        # we override create because we don't want anything stored in database
+        doc = docs[0]
+        service = doc["service"]
+        item_id = doc.get('item_id')
+        try:
+            service = registered_ai_services[service]
+        except KeyError:
+            raise SuperdeskApiError.notFoundError("{service} service can't be found".format(service=service))
+
+        analyzed_data = service.analyze(item_id)
+        docs[0].update({"analysis": analyzed_data})
+        return [0]
+
+
+def init_app(app):
+    if AUTO_IMPORT:
+        tools.import_services(app, __name__, AIServiceBase)
+
+    endpoint_name = 'ai'
+    service = AIService(endpoint_name, backend=superdesk.get_backend())
+    AIResource.schema["service"]["allowed"] = list(registered_ai_services)
+    AIResource(endpoint_name, app=app, service=service)
+    superdesk.intrinsic_privilege(endpoint_name, method=['POST'])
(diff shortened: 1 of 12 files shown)


tests/text_checkers/ai/imatrics_test.py
```diff
@@ -0,0 +1,136 @@
...
+                    "weight": 1
+                },
+                {
+                    "media_topics": [],
+                    "title": "Service",
+                    "uuid": "44f52663-52f9-3836-ac45-ae862fe945a3",
+                    "weight": 0.8387794287831216
+                }
+            ]
+        }
```
petrjasek, line 134: I'm not sure if we should expose all that api data to clients, imo would be good to convert it to some internal format so we can have generic ui widget, eg I wouldn't send weight but we could just use it to sort on backend, also I would rather stick with ninjs specs so using `name` and `code` and `subject/object/event/person/place` on top level, not sure if this api can return anything but subjects though

superdesk/text_checkers/ai/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+            'type': 'string',
+            'required': True,
+        },
+        'item_id': {
+            'type': 'string',
+            'required': True,
+        },
+    }
+    datasource = {
+        'projection': {'analysis': 1}
```
ride90, line 29: I think we don't have `analysis` defined in the schema

superdesk/text_checkers/tools.py
```diff
@@ -0,0 +1,46 @@
...
+
+def import_services(
+    app,
+    pkg_name: str,
+    base_cls: type,
+) -> None:
+    """Import all services based on base_cls in given package
+
+    This method will import python modules and look for a base_cls subclass there
+    If found, the subclass will be instanciated
```
ride90, line 25: typo in `instanciated`

tests/text_checkers/ai/imatrics_test.py
jerome-poisson: yeah I was not sure about this one, it's already sorted like with it in the backend, so I guess I can just get rid of it.
Yes this API can return locations, persons, etc.

superdesk/text_checkers/ai/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+import superdesk
+
+AUTO_IMPORT = True
+
+
+class AIResource(Resource):
+    schema = {
+        'service': {
+            'type': 'string',
+            'required': True,
```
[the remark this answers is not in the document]
jerome-poisson: yes it's only iMatrics, I've put it in a more genertic endpoint as I suspect we'll have more machine learning service in the future (it's trendy nowadays).
I'll add the allowed as you suggest.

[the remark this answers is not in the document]
jerome-poisson: @ride90 thinking twice about it, if I limit names with `allowed` we're loosing new services detection (we just have to create a new service and it will be automatically added to the endpoint), so I think using `allowed` is not a good idea at the end.

superdesk/text_checkers/ai/__init__.py
jerome-poisson: `analysis` is used for returned value. If I don't put the project here, it will be removed by Eve.

superdesk/text_checkers/tools.py
jerome-poisson: good catch, thx

superdesk/text_checkers/ai/__init__.py
ride90: you can also switch off projection per domain see https://github.com/superdesk/superdesk-core/pull/1961/files
but having `'projection': {'analysis': 1}` is also 👍

tests/text_checkers/ai/imatrics_test.py
jerome-poisson: @petrjasek I've put a mapping, but the screenshots at https://wiki.sourcefabric.org/display/NR/Auto-tagging+tool shows iMatrics terms, so I'm not sure if this what the customer expects.
Please check the mapping, how would you map "category"?

petrjasek: good question, I suppose it's some kind of subject but not sure

superdesk/text_checkers/ai/__init__.py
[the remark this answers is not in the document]
petrjasek: you can do something like in [ingest](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/ingest_provider_model.py#L71)

tests/text_checkers/ai/imatrics_test.py
petrjasek: otherwise the mapping looks good, I just think we have to also update item schema so it can hold references to places/persons etc otherwise the widget won't be able to make use of it

superdesk/text_checkers/ai/__init__.py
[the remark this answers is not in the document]
jerome-poisson: oh indeed, haven't thought of doing it dynamically, good point I'll update.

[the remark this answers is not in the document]
jerome-poisson: done in latest update

tests/text_checkers/ai/imatrics_test.py
jerome-poisson: `place` was already there, I've added `person` in latest update.

jerome-poisson: also I'm trying to separate first-name from last-name, but it's a best guess has there is no information in the documentation I have on how the person in formatted beside the `Name LastName` name of the type.

ride90 (approved):

PR #1956, opened 2020-08-06
allow coping of items from desk to personal space

petrjasek:
SDANSA-417

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -828,6 +828,12 @@ WORKFLOW_ALLOW_MULTIPLE_UPDATES = False
 #:
 WORKFLOW_ALLOW_DUPLICATE_TO_NON_MEMBERS = False
 
+#: Allow users to copy content from desk to personal
+#:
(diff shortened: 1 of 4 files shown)


superdesk/default_settings.py
```diff
@@ -828,6 +828,12 @@ def local_to_utc_hour(hour):
 #:
 WORKFLOW_ALLOW_DUPLICATE_TO_NON_MEMBERS = False
 
+#: Allow users to copy content from desk to personal
+#:
+#: .. versionadded:: 1.34
+#:
+WORKFLOW_ALLOW_COPY_TO_PERSONAL = True
```
ride90, line 835: Does it make sense to keep it disabled by default and enable only for ansa?

petrjasek: @karelpetrak any feedback on this?

karelpetrak: It can be enabled by default with the option to disable it via the settings.

PR #1964, opened 2020-08-17
SDESK-5482 Fix regex.sub flags

ride90:
⚠️ I've marked it as **2.0** because it's required by Belga and Belga is running on **release/2.0**
I will cherry-pick it after it's merged

### DIFF
--- a/tests/text_utils_test.py
+++ b/tests/text_utils_test.py
@@ -7,6 +7,7 @@ class WordCountTestCase(unittest.TestCase):
 
     def test_word_count_p_tags(self):
         self.assertEqual(2, text_utils.get_word_count('<p>foo<strong>s</strong></p><p>bar</p>'))
+        self.assertEqual(500, text_utils.get_word_count('<p>word</p>' * 500))
 
     def test_word_count_brs(self):
         self.assertEqual(2, text_utils.get_word_count('<p>foo<br><br>bar</p>'))
--- a/superdesk/text_utils.py
+++ b/superdesk/text_utils.py
@@ -31,19 +31,20 @@ def get_text_word_count(text):
     :return int: word count
     """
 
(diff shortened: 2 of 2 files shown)


tests/text_utils_test.py
```diff
@@ -7,6 +7,7 @@ class WordCountTestCase(unittest.TestCase):
 
     def test_word_count_p_tags(self):
         self.assertEqual(2, text_utils.get_word_count('<p>foo<strong>s</strong></p><p>bar</p>'))
+        self.assertEqual(500, text_utils.get_word_count('<p>word</p>' * 500))
```
tomaskikutis, line 10: Why not inline `<p>word</p><p>word</p><p>word</p>` ? Does it only break after word count reaches 500?

ride90: It used to break after 40

ride90:
> What was the issue? Why did you drop unicode flag?

The issue is that [regex.sub](https://docs.python.org/3/library/re.html#re.sub) has only 3 positional arguments, `count` and `flags` are keyword arguments, so it's better to use them explicitly.
In our case, since `count` was used instead of `flag` for `regex.MULTILINE | regex.UNICODE` (which is 40). 

`regex.UNICODE` is redundant because in python 3 every string by default is unicode.

tests/text_utils_test.py
tomaskikutis: Can you add a comment explaining the "magic" breaking point of 40 in the code?

tomaskikutis:
Could you restore the Unicode flag? The algorithm is meant to be duplicated in the client as closely as possible and that flag helps not to forget to add it in the client since it's not a default there.

ride90:
@tomaskikutis done

PR #1965, opened 2020-08-17
mypy integration

ride90:


apps/desks.py
```diff
@@ -607,8 +608,12 @@ def on_fetched(self, doc):
             for b in response.hits['aggregations']['overview']['buckets']
         ]
 
-    def _users_aggregation(self, desk_id: str) -> None:
+    def _users_aggregation(self, desk_id: str) -> List:
```
jerome-poisson, line 611: would be nice to specify the type of items here: `-> List[dict]`

apps/desks.py
```diff
@@ -675,7 +680,7 @@ def _users_aggregation(self, desk_id: str) -> None:
         overview = []
         for a in users_aggregation:
             role = a['_id']
-            authors_dict = {}
+            authors_dict: Dict[str, Any] = {}
```
jerome-poisson, line 683: I think this syntax is not available in Python 3.5, that's one of the reason I want to get rid of it. If this PR is for Superdesk 2.x, it may not be a pb though. I won't repeat this comment, but it also applies in other locations in this patch.

apps/validate/validate.py
```diff
@@ -435,8 +435,8 @@ def _validate(self, doc, fields=False, **kwargs):
             v.allow_unknown = True
             try:
                 v.validate(doc['validate'], validation_schema)
-            except TypeError as e:
-                logger.exception('Invalid validator schema value "%s" for ' % str(e))
+            except TypeError as ex:
```
jerome-poisson, line 438: just curious, why changing it? I'm always using `e` and I see no reason to write one more char here.

setup.cfg
```diff
@@ -11,4 +11,11 @@ output_file = server.pot
 mapping_file = babel.cfg
 
 [nosetests]
-exclude-dir = prod_api/tests/
\ No newline at end of file
+exclude-dir = prod_api/tests/
+
+[mypy]
+python_version = 3.6
```
jerome-poisson, line 17: ok so we ditch 3.5 for good then?

superdesk/celery_app.py
```diff
@@ -77,7 +77,7 @@ def handle_exception(exc):
     logger.exception(exc)
 
 
-class AppContextTask(TaskBase):
+class AppContextTask(TaskBase): # type: ignore
```
jerome-poisson, line 80: why those `type: ignore` are needed everywhere? Is it something for mypy of something else? It's making the source harder to read. Typing in python being optional and because it can be done progressively, I wonder why this is needed at all.

apps/desks.py
ride90: @jerome-poisson what syntax? Type definition and variable initialization in one row?

apps/validate/validate.py
ride90: It's a mypy bug I think. It was complaining about var `e` used outside of except block :)
https://github.com/superdesk/superdesk-core/pull/1965/files#diff-4d8b5e2477ab2c27d07c028d86b5cf71R442

setup.cfg
ride90: I've put 3.6 because Travis does the check and it uses 3.6

superdesk/celery_app.py
ride90: In this case, mypy was complaining that `TaskBase` can not be a base class. 
Mypy does not only type checking. I did not find a configuration to ignore this type of check.

apps/desks.py
ride90: @jerome-poisson just marked this pr to 2.1

jerome-poisson: type hint in random assignment, IIRC Python 3.5 only supports type hints in functions/methods definitions, class attributes and other variable can only be typed from Python 3.6+

apps/validate/validate.py
jerome-poisson: Would it be possible to fix Mypy (update?) or ignore this error in its settings? That could save my wrist/fingers for future damages.

ride90: @jerome-poisson , I agree that this part is dangerous for wrists, will check how to fix/ignore it.

setup.cfg
MarkLark86: It was discussed in the core group, Superdesk v2.0 will drop support for Python <= 3.5

superdesk/celery_app.py
MarkLark86: That's because `Celery.Task` is actually a function not a Class.

apps/legal_archive/resource.py
```diff
@@ -66,4 +66,4 @@ class LegalPublishQueueResource(LegalResource, PublishQueueResource):
     item_schema.update(PublishQueueResource.schema)
     schema = item_schema
 
-    datasource = {'source': LEGAL_PUBLISH_QUEUE_NAME}
+    datasource = {'source': LEGAL_PUBLISH_QUEUE_NAME} # type: ignore
```
MarkLark86, line 69: Why is `datasource` requiring `type: ignore` here but none of the other classes require it?

superdesk/default_settings.py
```diff
@@ -26,7 +26,7 @@
 try:
     from urllib.parse import urlparse
 except ImportError:
-    from urlparse import urlparse
+    from urlparse import urlparse # type: ignore
```
MarkLark86, line 29: As we have moved to Python 3.6, can we just use `urllib.parse`?

superdesk/publish/publish_queue.py
```diff
@@ -9,6 +9,7 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 import logging
+from typing import Dict, Any
```
MarkLark86, line 12: Is this used in this file?

apps/desks.py
ride90: @jerome-poisson indeed https://www.python.org/dev/peps/pep-0526/
but this feature (mypy) is for SD 2.1, so we are fine ;)

apps/validate/validate.py
ride90: @jerome-poisson the issue is that mypy complains about reassigning another type to variable `e` right after try/except block (https://github.com/superdesk/superdesk-core/pull/1965/files#diff-4d8b5e2477ab2c27d07c028d86b5cf71R442). It's a generic hint and I think it's useful to have it switched on. Python does not really care about it, but from code style point of view it's better to have this hint on.
What I don't understand in this particular case is why mypy complains about it "here", because variable `e` is only defined in `except` context.
I will explore it and possibly switch it off if possible, the current behavior is weird IMO.

ride90: Unfortunately, there is no way to switch it off.
Mypy has only this:

> allow_redefinition (bool, default False)
> Allows variables to be redefined with an arbitrary type, as long as the redefinition is in **the same block and nesting level as the original definition**.

superdesk/default_settings.py
ride90: makes sense

superdesk/celery_app.py
ride90: There are some unclear moments for me with dynamic types in mypy. 
Basically `Celery.Task` is a property which returns a base class and mypy complains about it (_error: Variable "superdesk.celery_app.TaskBase" is not valid as a type_).
I've tried to define a type for `TaskBase`:

```
TaskBase: Type[celery.Task] = celery.Task
```

but mypy has another opinion:

```
Name 'celery.Task' is not defined
```

Anyway, I've decided to ignore this one for now :)

apps/legal_archive/resource.py
ride90: `PublishQueueResource` redefines type defined in original `Resource.datasource`.

apps/validate/validate.py
petrjasek: isn't the error rather this then? `for e in error_list:`

petrjasek (approved):

apps/validate/validate.py
ride90: It is, variable e in for e in error_list context is used in more places that e from except TypeError as e: context, so I've changed the one which requires fewer changes :)

petrjasek: I see

PR #1966, opened 2020-08-18
Improve display of validation error messages. [SDESK-5444]

GyanP:


apps/validate/validate.py
```diff
@@ -21,14 +21,63 @@
...
-REQUIRED_FIELD = 'is a required field'
-MAX_LENGTH = 'max length is {length}'
-STRING_FIELD = 'require a string value'
-DATE_FIELD = 'require a date value'
-REQUIRED_ERROR = '{} is a required field'
-INVALID_CHAR = 'contains invalid characters'
+from flask_babel import lazy_gettext, _
+
+
+REQUIRED_FIELD = _('is a required field')
```
petrjasek, line 27: I think you have to use lazy for all those top level variables

apps/validate/validate.py
```diff
@@ -79,6 +79,17 @@
     'usageterms': lazy_gettext('Usage Terms'),
 }
 
+ERROR_MESSAGES = {
+    'REQUIRED_FIELD_MESSAGE': lazy_gettext('is a required field'),
```
petrjasek, line 83: I would use the constants so just `REQUIRED_FIELD` as key and also later when using it `ERROR_MESSAGES[REQUIRED_FIELD]`

petrjasek (approved):

PR #1967, opened 2020-08-19
FIX: Post error 400 is thrown on saving newly added datetime field from settings. [SDESK-5441]

GyanP:


superdesk/vocabularies/vocabularies.py
```diff
@@ -168,7 +168,14 @@ class VocabulariesResource(Resource):
         'custom_field_config': {
             'type': 'dict',
             'nullable': True,
-            'schema': {},
+            'schema': {
+                'increment_steps': {
```
ride90, line 172: what this `increment_steps` does?

GyanP: Inside `Settings->metadata->other fields`, add new field and select the Field type = datetime then there you will get the increment steps which actually gives us the flexibility to extend the time of given interval.

ride90 (approved):

PR #1968, opened 2020-08-21
Translations on the backend. [SDCP-300]

GyanP:


apps/preferences.py
```diff
@@ -61,6 +61,11 @@ def sync_field(field, dest, default):
             sync_field('label', v, default)
             sync_field('category', v, default)
 
+        if v.get('label'):
+            v['label'] = _(v['label'])
```
MarkLark86, line 65: Gettext will only work with static strings. These translations will be provided when `v['label']` and `v['category']` is assigned a value

MarkLark86 (approved):

PR #1977, opened 2020-09-04
Convert error object to json string.

GyanP:


petrjasek (approved):

petrjasek:
just wondering what is actually broken?

GyanP:
> just wondering what is actually broken?

The Validation error object is not in the JSON format. Earlier we are handling and converting the error object into the JSON format from the front end but doing JSON parse in the front end throws an error, so now we will be handling it from the backend itself.

petrjasek:
so we're passing there a dict right? I would rather handle it in the output and not there, like here:
https://github.com/superdesk/superdesk-core/blob/develop/superdesk/factory/app.py#L49-L54

GyanP:
> so we're passing there a dict right?

Actually, this is the format we are getting: `"['CATEGORIES is a required field', 'BODY HTML is a required field', 'DATELINE is a required field', 'HEADLINE is a required field', 'SLUGLINE is too long', 'SUBJECT is a required field']"`
So that's why I used **json.dumps**

petrjasek:
ok just not sure why we would send this as a string in a json which client has to parse, we could send it as:
```js
{"validation_errors": [...]}
```

petrjasek (approved):
my bad, I forgot it's eve doing the string conversion on patch

PR #1979, opened 2020-09-07
fix vine dependency version

petrjasek:


petrjasek:
after releasing 5.0 (6.9.) celery or rather kombu is broken, I suppose they didn't set the version properly

ride90 (approved):

petrjasek:
actually porting this to 1.32 as well, but there won't be any release atm

PR #1981, opened 2020-09-09
migration: Update usageterms schema_field on db

pablopunk:
SDESK-5423

Since [this PR](https://github.com/superdesk/superdesk-core/pull/1729) the vocabulary `usageterms` has a `schema_field`, but some instances (like sd-master) don't have that in the database, so there are some errors related to that like SDESK-5423, that won't reproduce on newer instances but it does on sd-master.

### DIFF
--- a/superdesk/data_updates/00024_20200909-142600_vocabularies.py
+++ b/superdesk/data_updates/00024_20200909-142600_vocabularies.py
@@ -23,4 +23,3 @@ class DataUpdate(DataUpdate):
 
     def backwards(self, mongodb_collection, mongodb_database):
         pass
-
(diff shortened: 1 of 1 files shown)


petrjasek (approved):

superdesk/data_updates/00024_20200909-142600_vocabularies.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class DataUpdate(DataUpdate):
+
+    resource = 'vocabularies'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+      for vocabulary in mongodb_collection.find({'_id': 'usageterms'}):
+        if 'schema_field' not in vocabulary:
+          mongodb_collection.update({'_id': vocabulary.get(config.ID_FIELD)},
```
petrjasek, line 21: I think the indentation is wrong, we use 4 spaces

PR #1984, opened 2020-09-14
avoid default desk overriding when using saml auth

petrjasek:
SDCP-363

### DIFF
--- a/superdesk/users/services.py
+++ b/superdesk/users/services.py
@@ -519,15 +519,16 @@ class DBUsersService(UsersService):
             role = get_resource_service('roles').find_one(req=None, name=ignorecase_query(role_name))
             if role:
                 data['role'] = role['_id']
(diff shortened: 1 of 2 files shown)


superdesk/users/services.py
```diff
@@ -519,15 +519,16 @@ def _process_external_data(self, _data, update=False):
             role = get_resource_service('roles').find_one(req=None, name=ignorecase_query(role_name))
             if role:
                 data['role'] = role['_id']
-        if data.get('desk') or app.config.get('USER_EXTERNAL_DESK'):
+        if not update and data.get('desk') or app.config.get('USER_EXTERNAL_DESK'):
```
MarkLark86, line 522: It would make it easier to read the order of operations if you place the first two operations inside a bracket, i.e.
`(not update and data.get('desk')) or app.config.get('USER_EXTERNAL_DESK')`

MarkLark86 (approved):

PR #1987, opened 2020-09-15
[SDESK-5418] SAMS Eve MediaStorage provider

MarkLark86:


MarkLark86:
@petrjasek @GyanP I have converted this to `draft` in case I need to make any changes when working on the UI side

petrjasek (commented):
looks good, I would just think about this resource to sams mapping, imo it would make sense to be defined on the resource somehow, maybe using schema and like we have elastic mapping there we could have sams mapping

MarkLark86:
> looks good, I would just think about this resource to sams mapping, imo it would make sense to be defined on the resource somehow, maybe using schema and like we have elastic mapping there we could have sams mapping

I agree, I don't like this implementation either. Being as though SAMS is an opt-in plugin though, I don't think it would be best to have SAMS specifics added to the source schema directly. Unless we dynamically add this mapping to the schema from within SAMS on `init_app` bootstrap? Any other ideas?

petrjasek (approved):

PR #1988, opened 2020-09-15
Send to from desk to personal space. [SDANSA-424]

GyanP:


apps/duplication/__init__.py
```diff
@@ -50,6 +50,8 @@ def init_app(app):
                         description='Duplicate Content within a Desk')
     superdesk.privilege(name='translate', label='Translate Content within a Desk',
                         description='Translate Content within a Desk')
+    superdesk.privilege(name='send_to_personal', label='Send Content to Personal desk',
```
petrjasek, line 53: would be good to translate the label and description

petrjasek (commented):
pls add a test

petrjasek (approved):

PR #1989, opened 2020-09-15
fix place parser for NITF

petrjasek:
now it would generate `[{"name": None, "code": None}]` for files
with `evloc` element present without any matching attributes.

SDNTB-653


ride90:
Petr, you wanted to say it would not generate `[{"name": None, "code": None}]`, right?

petrjasek:
yep right, I mean now as before this fix

MarkLark86 (approved):

jerome-poisson (approved):

PR #1999, opened 2020-09-28
wrong search filter label for Wire (category) CV. [SDCP-368]

GyanP:
Added the translations object at the root of the vocabulary.

### DIFF
--- a/apps/prepopulate/data_init/vocabularies.json
+++ b/apps/prepopulate/data_init/vocabularies.json
@@ -13,6 +13,20 @@
             "name": {"translate": true},
             "qcode": {},
             "subject": {}
+        },
+        "translations": {
+            "display_name": {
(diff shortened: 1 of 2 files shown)


apps/prepopulate/data_init/vocabularies.json
```diff
@@ -6,13 +6,28 @@
...
         "schema": {
             "name": {"translate": true},
             "qcode": {},
             "subject": {}
+        },
+        "translations": {
+            "display_name": {
+                "ar": "فئات",
+                "da": "kategorier",
+                "de_DE": "Kategorien",
```
petrjasek, line 22: would be good to be consistent with those language keys, I think the idea is to stick with `de-DE` like format

petrjasek (approved):

PR #2001, opened 2020-10-01
Restrict content_templates API endpoint according to user privileges. [SDANSA-441]

GyanP:


tomaskikutis:
Creating personal templates [doesn't work](https://user-images.githubusercontent.com/2076953/95072025-dd5f2680-070a-11eb-86a1-cb9dd806bf00.png) for users without template-related privileges.

apps/templates/content_templates.py
```diff
@@ -194,7 +192,6 @@ class ContentTemplatesService(BaseService):
 
     def on_create(self, docs):
         for doc in docs:
-            self._validate_privileges(doc)
```
petrjasek: why not validate here?

apps/templates/content_templates.py
```diff
@@ -386,7 +383,7 @@ def _validate_privileges(self, doc):
         privileges = user.get('active_privileges', {}) if user else {}
 
         if (user and not doc.get('is_public')
-                and user != doc.get('user') and not privileges.get('personal_template')):
+                and user.get(config.ID_FIELD) != doc.get('user') and not privileges.get('personal_template')):
             raise SuperdeskApiError.badRequestError('You dont have the privilege to modify another user template')
```
petrjasek, line 387: would be good to add gettext here for those error messages

apps/templates/__init__.py
```diff
@@ -23,6 +23,9 @@ def init_app(app):
     service = ContentTemplatesService(endpoint_name, backend=superdesk.get_backend())
     ContentTemplatesResource(endpoint_name, app=app, service=service)
     superdesk.privilege(name=CONTENT_TEMPLATE_PRIVILEGE, label='Templates', description='Create templates')
+    superdesk.privilege(name='personal_template',
+                        label='Manage other user personal template',
+                        description='Manage other user personal template')
```
petrjasek, line 28: gettext missing for label/description

GyanP:
This small change is because at the creation time we have only the login user there will be no user in the document.

tomaskikutis:
Here are more issues I've found:

* Non-administrator users, without `personal_template` privilege can see(in a list) private templates of others. If a user has `content_templates` he can also MODIFY and DELETE private templates without `personal_template` privilege.
* Non-administrator users, without `content_templates` privilege can see in a list public templates(editing and deleting isn't allowed as expected).
* Incorrect error message is returned from the server when trying to delete a **private** template of another user, an error is thrown saying that I don't have a privilege to delete a **public** template.
```{"_status": "ERR", "_message": "You dont have the privilege to manage the public template", "internal_error": 400}```

File: apps/templates/content_templates.py [hunk 4]
```diff
@@ -192,8 +191,27 @@ class ContentTemplatesResource(Resource):
 
 class ContentTemplatesService(BaseService):
 
+    def get(self, req, lookup):
+        active_user = g.get('user', {})
+        privileges = active_user.get('active_privileges', {})
+        results = super().get(req, lookup)
```
petrjasek, line 197: imo it would be better to modify the lookup if needed, not sure if pagination would work otherwise.
shoudl be fine to add if needed:
```python
lookup.update({'$or': [{'is_public': True}, {'user': active_user['_id']}})
if privileges.get('personal_template'):
    lookup['$or'].append({'is_public': False})
```
not sure if it's all covered with that, just for illustration

petrjasek (approved):

PR #2016, opened 2020-10-16
fix(templates): Template fields changed before post are reverted

MarkLark86:
Field(s) are reverted if they're changed after creating metadata from template and before archive.post


marwoodandrew (approved):

MarkLark86:
Will implement the fix in Planning instead, as the current solution does not work and anything else would be a hack

PR #2017, opened 2020-10-16
Create a macro to set metadata to routed stories from desk's default template. [SDCP-375]

GyanP:


superdesk/macros/set_default_template_metadata.py
```diff
@@ -0,0 +1,83 @@
+import logging
+import superdesk
+from flask_babel import lazy_gettext
+from superdesk import get_resource_service
+
+fields_to_exclude = superdesk.app.config['FIELDS_TO_EXCLUDE']
+fields_to_override = superdesk.app.config['FIELDS_TO_OVERRIDE']
```
petrjasek, line 7: I would name it more specifically, something like:
`DEFAULT_TEMPLATE_METADATA_MACRO_EXCLUDE`

superdesk/macros/set_default_template_metadata.py
```diff
@@ -0,0 +1,83 @@
+import logging
+import superdesk
+from flask_babel import lazy_gettext
+from superdesk import get_resource_service
+
+fields_to_exclude = superdesk.app.config.get('DEFAULT_TEMPLATE_METADATA_MACRO_EXCLUDE')
+fields_to_override = superdesk.app.config.get('DEFAULT_TEMPLATE_METADATA_MACRO_OVERRIDE')
```
petrjasek, line 7: you should only resolve this when macro is called, otherwise the `app.config` might not be loaded yet

superdesk/macros/set_default_template_metadata.py
```diff
@@ -72,7 +72,7 @@ def set_default_template_metadata(item, **kwargs):
             to_add.append(sub)
 
     # and now we add the new one
-    item.get('subject').extend([i for i in to_add if i.get('scheme')])
+    item.get('subject', []).extend([i for i in to_add if i.get('scheme') and to_add])
```
petrjasek, line 75: this wouldn't set anything on item if subject wasn't there before, you can use rather setdefault:
```py
item.setdefault('subject', []).extend(...)
```

petrjasek (approved):

PR #2025, opened 2020-10-22
Update the state when publish from personal is set. [SDANSA-434]

GyanP:


apps/archive/archive.py
```diff
@@ -320,7 +320,8 @@ def on_update(self, updates, original):
         self._validate_updates(original, updates, user)
 
         if self.__is_req_for_save(updates):
-            update_state(original, updates)
+            publish_from_personal = flask.request.args.get('publish_from_personal')
```
petrjasek, line 323: I think there might be cases when this is used and there is no request (when using celery), would be good to check if there is a request

petrjasek (approved):

PR #2026, opened 2020-10-23
User without template related privilege will not be able to see all the templates. [SDANSA-441]

GyanP:


apps/templates/content_templates.py
```diff
@@ -197,9 +197,13 @@ def get(self, req, lookup):
...
             lookup = {}
-        lookup.update({'$or': [{'is_public': True}, {'user': active_user.get('_id')}]})
-        if privileges.get('personal_template'):
-            lookup['$or'].append({'is_public': False})
+        if not privileges.get('content_templates') and not privileges.get('personal_template'):
+            lookup.update({'$and': [{'is_public': False}, {'user': active_user.get('_id')}]})
+        elif not privileges.get('content_templates') and privileges.get('personal_template'):
+            lookup.update({'$or': [{'is_public': False}, {'user': active_user.get('_id')}]})
+        elif privileges.get('content_templates') and not privileges.get('personal_template'):
+            lookup.update({'$or': [{'is_public': True}, {'user': active_user.get('_id')}]})
```
petrjasek, line 205: I think there are only 2 cases - everyone can see public right? so then I can either see also other user templates or not.

GyanP: There is a total of 4 cases:
1) If the user doesn't have the `content_template` or `personal_template` privilege, then the user should only see their personal template.
2) If the user has `content_template` privilege then the user should be able to see all the public template.
3) If the user has `personal_template` privilege enabled then the user should be able to see other user's personal templates as well.
4) If the user has both the privileges then the user should be able to see all the templates.

tomaskikutis: That's correct. 4th point is more of a conclusion than the rule.

* no privileges - can work with personal templates.
* content_template - can work with all public templates.
* personal_template - can work with private templates of other users.

petrjasek: so you need a privilege to be able to use a public template? I thought it's just for management

GyanP: > so you need a privilege to be able to use a public template?

We are just managing it using `content_templates` privilege.

tomaskikutis: > so you need a privilege to be able to use a public template? I thought it's just for management

My bad. I meant to say to "create/view/edit" templates when I wrote "can work with". Anyone should be able to **read** public templates. Not sure what should be done with private ones. If I create an article with a private template and send it to some desk - people wouldn't be able to view/edit articles without being able to access template.

petrjasek (approved):

PR #2027, opened 2020-10-23
add mapping for dateline.located.place

petrjasek:


superdesk/metadata/item.py
```diff
@@ -110,6 +110,16 @@ class Formats(NamedTuple):
 
 ITEM_EVENT_ID = 'event_id'
 
+geopoint = {
+    'type': 'dict',
+    'mapping': {'type': 'geo_point'},
+    'nullable': True,
+    'schema': {
+        'lat': {'type': 'integer'},
```
ride90, line 118: I think it's float

petrjasek: of course :facepalm:

ride90 (approved):

PR #2030, opened 2020-10-28
FIX: Placeline not prefilled from user preferences in user-created/edited templates. [SDCP-381]

GyanP:


apps/templates/content_templates.py
```diff
@@ -506,7 +506,12 @@ def render_content_template(item, template, update=False):
     :param dict template: template
     :return dict: updates to the item
     """
+
     kwargs = dict(item=item, user=get_user())
+    dateline_present_in_user_preferences = (kwargs['user'].get('user_preferences', {})
+                                            .get('dateline:located', {}).get('located'))
+    if dateline_present_in_user_preferences:
+        TEMPLATE_DATA_IGNORE_FIELDS.add('dateline')
```
petrjasek, line 514: once you add it there it will use that for all following requests no matter if dateline will be in user preferences or not. imo it should make a copy of the fields and then add it just for this call

apps/templates/content_templates.py
```diff
@@ -506,18 +506,19 @@ def render_content_template(item, template, update=False):
     :param dict template: template
     :return dict: updates to the item
     """
-
+    NEW_TEMPLATE_DATA_IGNORE_FIELDS = TEMPLATE_DATA_IGNORE_FIELDS.copy()
```
petrjasek, line 509: it shouldn't be capitalized imo, that implies it's a constant, but you modify it later

petrjasek (approved):

PR #2031, opened 2020-10-29
Templates privilege cannot use desk templates for creating new article [SDANSA-441]

GyanP:
when we hit `/content_templates` API and If a user doesn't have any privileges then all templates in the database collection should be returned except private templates of others when that privilege is not present.

### DIFF
--- a/apps/templates/content_templates.py
+++ b/apps/templates/content_templates.py
@@ -197,16 +197,16 @@ class ContentTemplatesService(BaseService):
         privileges = active_user.get('active_privileges', {})
         if not lookup:
             lookup = {}
-        manage = req.args.get('manage')
-
-        if manage and not privileges.get('content_templates') and not privileges.get('personal_template'):
-            lookup.update({'$and': [{'is_public': False}, {'user': active_user.get('_id')}]})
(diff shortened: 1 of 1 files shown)


apps/templates/content_templates.py
```diff
@@ -197,13 +197,16 @@ def get(self, req, lookup):
...
+
+        if manage and not privileges.get('content_templates') and not privileges.get('personal_template'):
             lookup.update({'$and': [{'is_public': False}, {'user': active_user.get('_id')}]})
-        elif not privileges.get('content_templates') and privileges.get('personal_template'):
+        elif manage and not privileges.get('content_templates') and privileges.get('personal_template'):
             lookup.update({'$or': [{'is_public': False}, {'user': active_user.get('_id')}]})
-        elif privileges.get('content_templates') and not privileges.get('personal_template'):
+        elif (manage and privileges.get('content_templates')
+                and not privileges.get('personal_template')
+                or not manage and not privileges.get('personal_template')):
```
petrjasek, line 208: this condition seems kind of complicated, can you pls split the manage and not manage part

petrjasek (approved):

PR #2035, opened 2020-11-03
[DESK][OVERVIEW] add filtering by slugline, headline, and byline

jerome-poisson:
A post query can be used to specify filters to the query. When used,
filters key(s) must be one of `slugline`, `headline` or `byline`, and the
value if a list of string to match.

SDESK-5642

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -561,14 +561,43 @@ class SluglineDeskService(BaseService):
 
 class OverviewResource(Resource):
     url = r'desks/<regex("([a-f0-9]{24})|all"):desk_id>/overview/<regex("stages|assignments|users"):agg_type>'
+    privileges = {'POST': 'desks'}
     resource_title = "desk_overview"
-    resource_methods = ['GET']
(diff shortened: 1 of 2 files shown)


apps/desks.py
```diff
@@ -561,20 +561,49 @@ def _find_other_sluglines(self, family_id, slugline, versioncreated, desk_id):
 
 class OverviewResource(Resource):
     url = r'desks/<regex("([a-f0-9]{24})|all"):desk_id>/overview/<regex("stages|assignments|users"):agg_type>'
+    privileges = {'POST': 'desks'}
     resource_title = "desk_overview"
-    resource_methods = ['GET']
+    resource_methods = ['GET', 'POST']
+    schema = {
```
ride90, line 567: I think that it's a good time to introduce `flake8-quotes` plugin into the core.
The current schema has a single quote style, but the code below is replacing all single quotes to double quotes 😄

ride90 (approved):

PR #2039, opened 2020-11-06
FIX: Create a macro to set metadata to routed stories from desk's default template. [SDCP-375]

GyanP:


superdesk/macros/set_default_template_metadata.py
```diff
@@ -68,7 +68,7 @@ def set_default_template_metadata(item, **kwargs):
     to_add = []
     for sub in subject:
         if (sub.get('scheme')
-                and not any(subjects.get('scheme') == sub.get('scheme') for subjects in item.get('subject'))):
+                and not any(subjects.get('scheme') == sub.get('scheme') for subjects in item.get('subject', {}))):
```
petrjasek, line 71: it should be rather `[]`

petrjasek (approved):

PR #2040, opened 2020-11-09
Product test page shows incorrect results and has a broken UI. [SDESK-5630]

GyanP:


petrjasek (approved):

petrjasek:
I wonder do we have a behave test for this?

GyanP:
> I wonder do we have a behave test for this?

I don't think we have any behave tests related to this.

PR #2041, opened 2020-11-10
[SDESK-5583] Add url_for_external to MediaStorage Providers

MarkLark86:
Update NinJS formatter to use this new method for attachment url


superdesk/attachments.py
```diff
@@ -73,6 +74,23 @@ def is_attachment_public(attachment):
...
+    :return: None if the attachment is not public, otherwise the public URL to the file
+    """
+
+    if attachment.get('attachment'):  # retrieve object reference
+        attachment = superdesk.get_resource_service('attachments').find_one(req=None, _id=attachment['attachment'])
+
+    if attachment.get('internal'):
+        return None
+
+    return current_app.media.url_for_external(attachment['media'], RESOURCE)
```
ride90, line 91: Is there any chance that `media` not in `attachment`?

superdesk/storage/amazon_media_storage.py
```diff
@@ -73,6 +73,19 @@ def url_for_media(self, media_id, content_type=None):
...
+
+        Returns a URL for use with the Content/Production API
+
+        :param str media_id: The ID of the asset
+        :param str resource: The name of the resource type this Asset is attached to
+        :rtype: str
+        :return: The URL for external use
+        """
+
+        return f'/assets/{str(media_id)}'
```
ride90, line 87: I think `str` is redundant here.
Kudos for using f-strings 👍

superdesk/storage/desk_media_storage.py
```diff
@@ -63,6 +63,19 @@ def url_for_download(self, media_id, content_type=None):
...
+
+        Returns a URL for use with the Content/Production API
+
+        :param str media_id: The ID of the asset
+        :param str resource: The name of the resource type this Asset is attached to
+        :rtype: str
+        :return: The URL for external use
+        """
+
+        return f'/assets/{str(media_id)}'
```
ride90, line 77: same here

superdesk/attachments.py
MarkLark86: Attachments **should** have a `media` field.
I have placed a check and logged a warning if this is not the case

ride90 (approved):

PR #2043, opened 2020-11-11
As a user I want to run a macro when I create an article based on a template schedule. [SDNTB-660]

GyanP:


petrjasek:
pls add a test

petrjasek (approved):

PR #2045, opened 2020-11-11
[DESK][OVERVIEW] add items when a filter is used

jerome-poisson:
When a filter is used, the whole items are returned (up to 100).
The items can be found in a list at `docs` key.

SDESK-5642

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -660,6 +660,9 @@ class OverviewService(BaseService):
                 for text in f_data:
                     should.append({"match": {f_name: text}})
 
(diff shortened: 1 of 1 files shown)


jerome-poisson:
I haven't included `docs` in tests because the json check that we currently use in behave will not skip the unspecified keys so deep in the hierarchy (so `_current_version` and all won't match). Anyway I've tested it and it's working.

petrjasek (approved):

ride90 (approved):

PR #2049, opened 2020-11-18
fix build on readthedocs

petrjasek:
it's timing out on install, due to many dev dependencies


petrjasek:
there is also a new readthedocs check now activated

ride90 (approved):

PR #2053, opened 2020-11-24
Correction workflow: As a SD admin I want to have a config to introduce a corrections workflow. [SDCP-390]

GyanP:


petrjasek:
I wonder why we need those `correction_of` and `corrected_by` fields?

GyanP:
> I wonder why we need those `correction_of` and `corrected_by` fields?

We need both of these for making a link between the **correction** and **being corrected** articles because we have an action button **cancel correction** which is used to reset the article to its previous state.

petrjasek:
yep but the guid should be the same for both, you can only correct single item, not any other version of it right?

apps/publish/content/correct.py
```diff
@@ -78,6 +80,22 @@ def set_state(self, original, updates):
         else:
             super().set_state(original, updates)
 
+    def change_being_corrected_to_published(self, updates, original):
+        if app.config.get('CORRECTIONS_WORKFLOW') and original.get('state') == 'correction':
+            publish_service = get_resource_service('published')
+            being_corrected_articles = publish_service.find({
```
petrjasek, line 86: is it actually possible to have more than 1? if not using `find_one` would be better to make it explicit

petrjasek: I might have been wrong here, if it's published collection there might be more items with same guid, that would be the case in archive

petrjasek (approved):

PR #2054, opened 2020-11-24
implement media storage proxy

petrjasek:
it can handle both mongo and amazon together.
plus add command to move media from mongo to amazon.

SDESK-5653

### DIFF
--- a/superdesk/storage/migrate.py
+++ b/superdesk/storage/migrate.py
@@ -1,4 +1,3 @@
-
 import codecs
 import superdesk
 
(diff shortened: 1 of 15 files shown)


superdesk/storage/migrate.py
```diff
@@ -0,0 +1,40 @@
...
+                file.read(),
+                filename=file.filename,
+                content_type=file.content_type,
+                metadata=file.metadata,
+                _id=str(file._id),
+                ContentMD5=codecs.encode(codecs.decode(file.md5, 'hex'), 'base64').decode().strip(),
+            )
+            mongo.delete(file._id)
+            counter += 1
+            if counter == limit:
```
MarkLark86, line 34: Rather than manually using a counter, would it be better to use the `limit` function on the cursor itself (or as a parameter to the `find` method)?

superdesk/storage/migrate.py
```diff
@@ -0,0 +1,40 @@
...
+    def run(self, limit):
+        mongo = app.media._storage[0]
+        amazon = app.media._storage[1]
+
+        files = mongo.fs().find()
+        print("starting to migrate {}/{} files".format(limit, files.count()))
+
+        counter = 0
+        for file in files:
+            amazon.put(
```
MarkLark86, line 24: Encapsulating the upload in a `try...except` would be ideal in case of an exception being raised. No need to abort the entire process, and further iterations wouldn't abort on the same resource

MarkLark86 (approved):

PR #2074, opened 2020-12-09
SDESK-5671 ADD: Endpoint to lock and unlock asset.

dev-tsoni:


superdesk/sams/assets.py
```diff
@@ -148,3 +148,25 @@ def get_assets_compressed_binary(asset_ids):
         item_ids=asset_ids
     )
     return zip_binary.content
+
+
+@assets_bp.route("/sams/assets/lock/<asset_id>", methods=["PATCH"])
+def lock_asset(asset_id):
+    docs = {"lock_action": "edit"}
```
MarkLark86, line 155: `lock_action` should be provided by the front-end (later we may add further actions on Assets that will require a different lock action)

dev-tsoni: Okay.

MarkLark86 (approved):

PR #2079, opened 2020-12-14
FIX: cancel_correction for an corrected item should reset the state to corrected.

GyanP:


apps/archive/archive_correction.py
```diff
@@ -48,34 +48,54 @@ def on_update(self, updates, original):
...
-            published_article = published_service.find_one(req=None,
-                                                           guid=original.get('guid'),
-                                                           correction_sequence=original.get('correction_sequence'))
-        else:
-            published_article = published_service.find_one(req=None,
-                                                           guid=original.get('guid'))
-
         published_service.patch(id=published_article.get(config.ID_FIELD), updates=published_item_updates)
 
+        #  if remove_correction, set the item values back to the old one.
```
petrjasek, line 89: not sure what should be in the history, but might be better to just revert to previous version than doing this copy from published which has other metadata

petrjasek (approved):

PR #2087, opened 2020-12-18
FIX: 'Subject is a required field', although there is only an initialized Index field in the authoring. [SDESK-5702]

GyanP:


petrjasek (commented):
test pls

petrjasek (approved):

PR #2091, opened 2020-12-22
FIX: Associated media that was sent to another desk is not moved to desk output stage upon publishing the containing article. [SDESK-5709]

GyanP:
Reason:
The content update event refreshes the stages, but the push notification for the content update **(push_content_notification)** just send the stage for the article and not for the associated item because we just pop it out, and because of that, the content update event gets only the article stage and it refreshes only the article stage and not the associated item stage.

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -719,7 +719,6 @@ class BasePublishService(BaseService):
 
                 if associated_item.get('state') not in PUBLISH_STATES:
                     # This associated item has not been published before
-                    associated_item.get('task', {}).pop('stage', None)
                     remove_unwanted(associated_item)
 
                     # get the original associated item from archive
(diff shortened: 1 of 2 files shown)


apps/publish/content/common.py
```diff
@@ -719,7 +719,6 @@ def _publish_associated_items(self, original, updates=None):
 
                 if associated_item.get('state') not in PUBLISH_STATES:
                     # This associated item has not been published before
-                    associated_item.get('task', {}).pop('stage', None)
```
ride90, line 722: I wonder if we can break some logic by removing this line.

ride90: I wonder if the logic from this PR does not conflict with current changes.

GyanP: > I wonder if the logic from this PR does not conflict with current changes.

I have tested [this](https://github.com/superdesk/superdesk-core/commit/493d549b5647359a221863d2b4fbfb99ccf41c9f) PR, and it's not in conflict with the current changes.

ride90 (approved):

PR #2094, opened 2020-12-29
Improve expiry options for the authentication token of the Content API in the subscriber config. [SDESK-5707]

GyanP:


superdesk/publish/subscriber_token.py
```diff
@@ -31,5 +32,6 @@ class SubscriberTokenService(superdesk.Service):
     def create(self, docs, **kwargs):
         for doc in docs:
             doc['_id'] = get_random_token()
-            doc.setdefault('expiry', utcnow() + timedelta(days=7))
+            if not doc.get('never_expire'):
+                doc.setdefault('expiry', utcnow() + timedelta(days=7))
```
ride90, line 36: can we set `never_expire` and `expiry` together?

GyanP: Nope, If we have `never_expire` as null, then only we should set the expiry otherwise there is no need to set it.

petrjasek: it might be easier for client if we set expiry readonly and rather send some `expires_after` number which would be the number of days, if not set it would never expire.

petrjasek: maybe rather `expiry_days` to make the name explicit

petrjasek (changes requested):
you should also change the auth for content api to handle empty expiry

petrjasek (approved):

ride90 (approved):

PR #2095, opened 2020-12-30
FIX: Associated media that was sent to another desk is not moved to desk output stage upon publishing the containing article. [SDESK-5709]

GyanP:
If the associated item is sent to another stage, the stage should update everywhere.

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -737,7 +737,7 @@ class BasePublishService(BaseService):
                         })
                         continue
 
-                    if orig_associated_item.get('task', {}).get('stage'):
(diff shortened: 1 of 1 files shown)


petrjasek (commented):
would be good to add some comment why setting the stage, shouldn't it also check if the stage is not set on the `associated_item` ?

GyanP:
> shouldn't it also check if the stage is not set on the `associated_item` ?

we just need to check if the `association_item` has the `task` property if it does we need to update the old stage with the new one, (if exists) otherwise just add the stage.

petrjasek (approved):

PR #2098, opened 2021-01-04
fix(initialize_data): Dont terminate command on duplicate key error

MarkLark86:


petrjasek (approved):

MarkLark86:
@petrjasek This should be backported to 2.0, as I believe this may be the cause for CP UAT not getting updated planning_types

PR #2099, opened 2021-01-05
As a translator I´d like to have the update/writethru and correction count set to '0' and some metadata changed when performing the translation action on an updated story. [SDCP-392]

GyanP:


petrjasek:
we should handle such things via macro and not in core

GyanP:
> we should handle such things via macro and not in core

but we are handling it with a config `UPDATE_TRANSLATION_METADATA`. and one of the reasons I am handling inside core is we don't have to set it up in the UI, this needs to run behind the scenes without setting it up anywhere.

apps/duplication/archive_translate.py
```diff
@@ -29,6 +28,8 @@
 
 package_service = PackageService()
 
+UPDATE_TRANSLATION_METADATA_MACRO = superdesk.app.config.get('UPDATE_TRANSLATION_METADATA_MACRO')
```
petrjasek: I think it's better to use the `app.config` only in the service, otherwise this might be resolved before the settings are loaded properly

petrjasek (approved):

PR #2100, opened 2021-01-05
reformat code via black

petrjasek:
and use it on github actions


.github/workflows/nose-tests.yml
```diff
@@ -39,9 +39,6 @@ jobs:
         python -m pip install --upgrade pip wheel setuptools
         pip install -r dev-requirements.txt
 
-    - name: Flake8
```
ride90, line 42: I think we should add flake8 check into lint.yml

.github/workflows/nose-tests.yml
```diff
@@ -39,9 +39,6 @@ jobs:
...
-      run: flake8
-
     - name: Mypy
```
ride90, line 42: Does it make sense to move mypy check from nose-tests.yml to lint.yml ?

ride90 (approved):

PR #2105, opened 2021-01-07
SDESK-5713 ADD: Added unlock_asset_by_user on session end.

dev-tsoni:


superdesk/sams/__init__.py
```diff
@@ -53,3 +56,7 @@ def after_request(response):
         label=_('SAMS Manage Assets'),
         description=_('Allows management of SAMS Assets')
     )
+
+
+def unlock_assets_on_logout(user_id, session_id):
+    unlock_asset_by_user(user_id)
```
MarkLark86, line 62: It is possible a user could have multiple sessions open, so we only want to unlock Assets tied to a users `user_id` and `session_id`

MarkLark86 (approved):

PR #2106, opened 2021-01-08
SDESK-5714 ADD: websocket notifications for sams resources.

dev-tsoni:


MarkLark86:
Looks good. Don't forget to add notifications for `lock` and `unlock` actions as well.

When unlocking all Assets for a User/Session combination, I think we can just send those two IDs (`user_id` & `session_id`) in the websocket message with a name such as `sams:asset:session_unlock`.

Also it would be better to prefix all these notifications with `sams:` so that we can easily filter messages by for SAMS

MarkLark86:
After further looking into the requirements of websockets in client-core extensions, we also need to add the following to ALL websocket notifications for SAMS:
`extension='sams'`

So based on the above comments, a notification for Set created would be:
```python
push_notification(
    'sams:set:created',
    item_id=post_response.json()['_id'],
    user_id=get_user_id(True),
    session_id=get_auth()['_id'],
    _etag=post_response.json()['_etag'],
    extension='sams'
)
```

MarkLark86 (approved):

PR #2107, opened 2021-01-11
FIX: A story cannot be saved with Vatican in the placeline. [SDCP-422]

GyanP:


petrjasek (approved):

features/places.feature
```diff
@@ -1,6 +1,7 @@
 Feature: Places
 
     @auth
+    @wip
```
petrjasek, line 4: I guess this shouldn't be there

GyanP: oppss.

PR #2116, opened 2021-01-18
[SEARCH] fix `html_field_analyzer` case handling

jerome-poisson:
Custom `html_field_analyzer` was not using `lowercase` filter, so case
was kept. As a result, words with uppercase in them could not be found
with the search API.

SDCP-423

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -185,6 +185,7 @@ ELASTICSEARCH_SETTINGS = {
                 },
                 "html_field_analyzer": {
                     "type": "custom",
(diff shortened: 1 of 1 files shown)


jerome-poisson:
note the ES indexes must be flushed after applying this patch (`python manage.py app:flush_elastic_index --sd`).

petrjasek (approved):

petrjasek (changes requested):
actually we should do it also for content api index settings, just a few lines below

petrjasek (approved):

PR #2117, opened 2021-01-19
FIX: Resend action does not work for items in state Being corrected. [SDESK-5756]

GyanP:
[SDESK-5756](https://dev.sourcefabric.org/browse/SDESK-5756) [SDESK-5757](https://dev.sourcefabric.org/browse/SDESK-5757)

### DIFF
--- a/apps/publish/content/resend.py
+++ b/apps/publish/content/resend.py
@@ -72,7 +72,7 @@ class ResendService(Service):
     def _validate_article(self, article_id, article_version):
         archive_article = get_resource_service(ARCHIVE).find_one(req=None, _id=article_id)
 
-        if (app.config.get('CORRECTIONS_WORKFLOW') and archive_article.get(ITEM_STATE) == 'correction'):
+        if app.config.get("CORRECTIONS_WORKFLOW") and archive_article.get(ITEM_STATE) == "correction":
             publish_service = get_resource_service("published")
             archive_article = publish_service.find_one(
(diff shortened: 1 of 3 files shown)


apps/publish/content/resend.py
```diff
@@ -72,6 +72,12 @@ def _validate_subscribers(self, subscriber_ids, article):
     def _validate_article(self, article_id, article_version):
         archive_article = get_resource_service(ARCHIVE).find_one(req=None, _id=article_id)
 
+        if app.config.get("CORRECTIONS_WORKFLOW") and archive_article.get(ITEM_STATE) == "correction":
+            publish_service = get_resource_service("published")
+            archive_article = publish_service.find_one(
```
petrjasek, line 77: this is not `archive_article` really, not sure why it wouldn't find the article by `article_id` ?

GyanP: When the correction workflow config is set and we try to correct the item, it changes the state of the item present inside archive collection as `correction` and the item present inside the published collection as `being_corrected` whereas if we publish an item, it will change the state to `published` in both the collections (archive collection and published collection)

if I use this: `archive_article = get_resource_service(ARCHIVE).find_one(req=None, _id=article_id)`, it will return the item which is present inside the archive collection and the item present there has a state of `correction` but we need the item which has a state of `being_corrected` which is present inside the published collection.

petrjasek (approved):

PR #2121, opened 2021-01-21
As a system admin I want to display a system-wide message to all logged in users. [SDESK-5659]

GyanP:


apps/system_message/service.py
```diff
@@ -0,0 +1,65 @@
...
+        Send a notification
+        :param doc:
+        :return:
+        """
+        push_notification("system_message:deleted", _id=[doc.get(config.ID_FIELD)])
+
+    def _validate_message(self, doc):
+        max_characters = app.config.get("MAX_CHARACTERS_FOR_SYSTEM_MESSAGE", 200)
+        if doc.get("message") and len(doc["message"]) > max_characters:
+            raise SuperdeskApiError.badRequestError(_(f"Max {max_characters} characters are allowed"))
```
petrjasek, line 61: not sure this would work for gettext

apps/system_message/resource.py
```diff
@@ -0,0 +1,29 @@
...
+            "required": True,
+        },
+        "message_title": {"type": "string", "required": True},
+        "message": {"type": "string", "required": True},
+        "user": {"type": "string", "required": True},
+    }
+
+    resource_methods = ["GET", "POST"]
+    item_methods = ["GET", "PATCH", "DELETE"]
+    no_privileges = True
```
petrjasek, line 29: wouldn't it be easier to just define a new privilege?

apps/system_message/resource.py
```diff
@@ -0,0 +1,29 @@
...
+    schema = {
+        "is_active": {"type": "boolean", "default": False},
+        "type": {
+            "type": "string",
+            "allowed": ["warning", "alert", "primary/info", "success"],
+            "required": True,
+        },
+        "message_title": {"type": "string", "required": True},
+        "message": {"type": "string", "required": True},
+        "user": {"type": "string", "required": True},
```
petrjasek, line 24: what is this used for?

apps/system_message/service.py
```diff
@@ -0,0 +1,65 @@
...
+        self._validate_administrator("update")
+
+    def on_updated(self, updates, original):
+        """
+        Send notifification
+        :param updates:
+        :param original:
+        :return:
+        """
+        push_notification("system_message:udpated", _id=[original.get(config.ID_FIELD)])
```
petrjasek, line 45: typo in notification name

apps/system_message/resource.py
GyanP: only the "administrator" users should be able to perform create, update and delete operations, you think we should define a new privilege?

apps/system_message/resource.py
GyanP: Just in case, if we need it in the future to know who created the message.

apps/system_message/resource.py
petrjasek: yep I just wonder why we handle this differently than other authorizations

apps/system_message/resource.py
petrjasek: and is that a name or user id?

GyanP: it's the `user_id`. I think I got your point here, I'll change it to `user_id`

apps/system_message/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def init_app(app):
+    endpoint_name = "system_message"
+    service = SystemMessagesService(endpoint_name, backend=superdesk.get_backend())
+    SystemMessagesResource(endpoint_name, app=app, service=service)
+
+    superdesk.privilege(name="system_messages", label="System Message", description="User can manage system messages.")
```
petrjasek, line 24: this can use some `lazy_gettext`

apps/system_message/resource.py
```diff
@@ -0,0 +1,29 @@
...
+    schema = {
+        "is_active": {"type": "boolean", "default": False},
+        "type": {
+            "type": "string",
+            "allowed": ["warning", "alert", "primary/info", "success"],
+            "required": True,
+        },
+        "message_title": {"type": "string", "required": True},
+        "message": {"type": "string", "required": True},
+        "user_id": {"type": "string", "required": True},
```
petrjasek, line 24: I think it should be rather `Resource.rel("user", required=True)` so it will validate the value

petrjasek (approved):

PR #2126, opened 2021-01-26
replace `_` with `lazy_gettext`

petrjasek:
redoing https://github.com/superdesk/superdesk-core/pull/2122 which broke client e2e tests

### DIFF
--- a/superdesk/vocabularies/__init__.py
+++ b/superdesk/vocabularies/__init__.py
@@ -14,7 +14,7 @@ from superdesk.signals import item_published
 from .vocabularies import VocabulariesResource, VocabulariesService, is_related_content
 from .commands import UpdateVocabulariesInItemsCommand  # noqa
 from .keywords import add_missing_keywords
-from flask_babel import _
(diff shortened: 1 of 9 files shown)


superdesk/vocabularies/__init__.py
```diff
@@ -14,7 +14,7 @@
 from .vocabularies import VocabulariesResource, VocabulariesService, is_related_content
 from .commands import UpdateVocabulariesInItemsCommand  # noqa
 from .keywords import add_missing_keywords
-from flask_babel import _
+from flask_babel import _, lazy_gettext
```
dev-tsoni, line 17: do we need `_` here, if we are replacing it? I am just asking.

petrjasek: yep it's used in the file for some error messages, we don't need to replace it everywhere

dev-tsoni: Okay.

dev-tsoni (approved):

PR #2129, opened 2021-01-27
check privilege translations via mypy

petrjasek:
it needs stubs for flask_babel and actually validating imports


apps/publish/__init__.py
```diff
@@ -68,15 +71,33 @@ def init_app(app):
     service = UnpublishService(endpoint_name, backend=get_backend())
```
GyanP, line 71: this should be Any as well

apps/saved_searches/__init__.py
```diff
@@ -47,24 +50,30 @@ def init_app(app):
     service = SavedSearchItemsService(endpoint_name, backend=superdesk.get_backend())
```
GyanP, line 50: same here

GyanP (approved):

apps/publish/__init__.py
petrjasek: what do you mean?

petrjasek: it must be set only when service is first declared in the function

PR #2130, opened 2021-01-29
Introduce the preview for the publish queue also for non-text items. [SDESK-5491]

GyanP:
Inserts the latest versions of associated items into archive_versions.

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -861,6 +861,7 @@ class BasePublishService(BaseService):
                         remove_unwanted(association_updates)
                         publish_service.patch(id=association_updates.pop(config.ID_FIELD), updates=association_updates)
 
+            self._update_archive(associated_item, updates.get(ASSOCIATIONS, {}).get(associations_key))
         self._refresh_associated_items(original)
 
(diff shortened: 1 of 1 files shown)


apps/publish/content/common.py
```diff
@@ -861,6 +860,7 @@ def _publish_associated_items(self, original, updates=None):
                         remove_unwanted(association_updates)
                         publish_service.patch(id=association_updates.pop(config.ID_FIELD), updates=association_updates)
 
+            insert_into_versions(doc=associated_item)
```
petrjasek, line 863: can you add  a comment when this should be actually called? there are too many ifs..

GyanP: sure.

petrjasek: actually not sure why it's needed, if the item is being published now it would get the version saved automatically, wouldn't it? and if it was published already it's not modified so there should be no need to put it there

GyanP: if the item is being published it would get the version saved automatically, but it will not save/update the version of the associated item in the archive_version, that's why we need to add it here.

petrjasek (approved):

PR #2132, opened 2021-02-02
get_items vocabularies service helper.

ride90:


superdesk/vocabularies/vocabularies.py
```diff
@@ -444,6 +445,64 @@ def get_article_cv_item(self, item, scheme):
         article_item.update({"scheme": scheme})
         return article_item
 
+    def get_items(self, _id: str, qcode: str = None, is_active: bool = None,
```
petrjasek, line 448: imo `is_active` should be rather `True` by default

superdesk/vocabularies/vocabularies.py
```diff
@@ -444,6 +445,64 @@ def get_article_cv_item(self, item, scheme):
...
+                  name: str = None, lang: str = None) -> Union[List, None]:
```
petrjasek, line 449: might be easier for usage to return always a `List`, or `None` maybe if the cv doesn't exist, but not sure if that's needed

tests/vocabularies/vocabularies_service_get_items_tests.py
```diff
@@ -0,0 +1,269 @@
...
+                "name": "PREVIEW",
+                "qcode": "PREVIEW",
+                "is_active": True,
+                "translations": {
+                    "name": {
+                        "nl": "Voorbericht",
+                        "fr": "Avant-Papier"
+                    }
+                }
+            }]
```
petrjasek, line 144: imo would be good if the items are ready for use, so it would be without `is_active` and with `scheme` already set so you could then just do `item['subject'].extend(items)`

petrjasek (approved):

PR #2136, opened 2021-02-04
analyze_wildcard config for es query_string.

ride90:
The same as https://github.com/superdesk/superdesk-core/pull/2134 , but for **2.*** cores

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -190,7 +190,11 @@ ELASTICSEARCH_SETTINGS = {
                     "char_filter": ["html_strip_filter"],
                 },
             },
-        }
+        },
+        "query_string": {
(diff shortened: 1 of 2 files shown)


superdesk/default_settings.py
```diff
@@ -190,7 +190,11 @@ def local_to_utc_hour(hour):
                     "char_filter": ["html_strip_filter"],
                 },
             },
-        }
+        },
+        "query_string": {
```
petrjasek, line 194: seems like this config is not available in 7, and given that we set it per query, I suppose it's not needed anyhow right?

ride90: I did not test it yet (PR is not finished). The option in the es7 docs. 
I will test it and let you know.

PR #2139, opened 2021-02-09
FIX: Media contacts cannot be saved when Country is added to the address. [SDESK-5786]

GyanP:


apps/contacts/resource.py
```diff
@@ -81,7 +81,17 @@ class ContactsResource(Resource):
...
         "contact_state": {"type": "string", "required": False},
         "postcode": {"type": "string", "required": False},
-        "country": {"type": "string", "required": False},
+        "country": {
+            "type": "dict",
+            "schema": {
+                "name": {"type": "string", "required": False},
+                "qcode": {"type": "string", "required": False},
+                "translations": {
+                    "type": "dict",
```
petrjasek, line 90: would it work without schema? maybe you could add/update a test for this

GyanP: yes, it would work, but I thought it would be good if we define the schema for making it specific. what's your suggestion here, should I remove it and make it just as `"country": {"type": "dict", "required": False}`

petrjasek: I mean the schema for `translations`

GyanP: Ah ok, yes it's working properly without schema.

petrjasek (approved):

petrjasek (approved):

PR #2147, opened 2021-02-17
[DESKS][OVERVIEW] fixed items aggregation for big collections

jerome-poisson:
aggregation of items on big collections was not accurate due to default
size that ElasticSearch is using of terms aggregation. This is a quick
fix to work around that by using maximum size, but documentation
recommends to use pagination with composite aggregation (see
https://www.elastic.co/guide/en/elasticsearch/reference/7.11/search-aggregations-bucket-composite-aggregation.html),
so an other patch will be done later to use proper pagination.

SDESK-5822

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -32,6 +32,7 @@ from flask_babel import _
 
 
 logger = logging.getLogger(__name__)
+SIZE_MAX = 2147483647
 
 
 class DeskTypes(SuperdeskBaseEnum):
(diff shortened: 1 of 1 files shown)


apps/desks.py
```diff
@@ -32,6 +32,7 @@
 
 
 logger = logging.getLogger(__name__)
+SIZE_MAX = 2147483647
```
ride90, line 35: you can write 2_147_483_647 to make it more readable

ride90 (approved):

apps/desks.py
jerome-poisson: :+1:

PR #2150, opened 2021-02-22
translation macro: switch placeline to french value. [SDCP-451]

GyanP:


superdesk/places/places_autocomplete.py
```diff
@@ -50,3 +50,21 @@ def get(self, req, lookup):
...
+            ("geonameId", req.args.get("geonameId")),
+            ("lang", req.args.get("lang", "en").split("-")[0]),
+            ("style", req.args.get("style", app.config["GEONAMES_SEARCH_STYLE"])),
+        ]
+
+        if req.args.get("featureClass"):
+            params.append(("featureClass", req.args.get("featureClass")))
+        else:
+            for feature_class in app.config["GEONAMES_FEATURE_CLASSES"]:
+                params.append(("featureClass", feature_class.upper()))
```
petrjasek, line 66: do you need this? imo that's only for search

GyanP: Good Point.

petrjasek (approved):

PR #2154, opened 2021-02-23
Change the signature to get the geoname. [SDCP-451]

GyanP:


superdesk/places/places_autocomplete.py
```diff
@@ -51,14 +51,14 @@ def get(self, req, lookup):
...
 
-    def get_detail(self, req, lookup):
-        assert req.args.get("geonameId"), {"geonameId": 1}
-        params = [
-            ("geonameId", req.args.get("geonameId")),
-            ("lang", req.args.get("lang", "en").split("-")[0]),
-            ("style", req.args.get("style", app.config["GEONAMES_SEARCH_STYLE"])),
-        ]
+    def get_place(self, geoname_id, language):
+        if geoname_id and language:
```
petrjasek, line 55: why this if here? would rather do assert, but language could be optional. if you call this you must have `geoname_id` imo

GyanP: >  why this if here?
This if was just to check if `geoname` id is passed or not, if it's passed then ok if not then we need not have to perform any operation.

> would rather do assert
So you mean instead of if statement I should rather use something like:
```
assert geoname_id
        params = [
            ("geonameId", geoname_id),
            ("lang", language),
            ("style", app.config.get("GEONAMES_SEARCH_STYLE", "full")),
        ]

```
> language could be optional

Right.

petrjasek: yep exactly

petrjasek (approved):

PR #2162, opened 2021-03-05
FIX: Media contacts cannot be saved when 'State/province or region' is added to the address. [SDESK-5831]

GyanP:


superdesk/data_updates/00029_20210305-132352_contacts.py
```diff
@@ -0,0 +1,44 @@
...
+                        "contact_state": {
+                            "name": region[0].get("name"),
+                            "qcode": region[0].get("qcode"),
+                            "translations": region[0].get("translations"),
+                        }
+                    }
+                else:
+                    contact_state = {
+                        "contact_state": {
+                            "name": document["contact_state"],
```
MarkLark86, line 35: It might be a good idea to always have a `qcode`, even if it has the same value as `name`

MarkLark86 (approved):

PR #2164, opened 2021-03-10
SDESK-5581 ADD: endpoint to get tags using search-query

dev-tsoni:
Endpoint to get existing tags.

### DIFF
--- a/superdesk/sams/assets.py
+++ b/superdesk/sams/assets.py
@@ -206,7 +206,7 @@ def unlock_asset_by_user(user_id, session_id):
     return unlock_asset_response.status_code
 
 
-@assets_bp.route("/sams/assets/tags", methods=["GET"])
+@assets_bp.route("/sams/assets/tags", methods=["POST"])
 def get_assets_tags():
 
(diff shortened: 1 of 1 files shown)


superdesk/sams/assets.py
```diff
@@ -204,3 +204,16 @@ def unlock_asset_by_user(user_id, session_id):
             "sams:asset:session_unlock", user_id=get_user_id(True), session_id=get_auth()["_id"], extension="sams"
         )
     return unlock_asset_response.status_code
+
+
+@assets_bp.route("/sams/assets/tags/<search_query>", methods=["GET"])
```
MarkLark86, line 209: `search_query` should be a parameter not a suffix to the resource location.
In this case:

**resource**: /sams/assets/tags
**params**: query

This is because the resource here is `tags`, and `query` controls the response from that resource.

Example:
`GET /sams/assets/tags?query=Sports`

The `query` param should also be optional, so it should be possible to retrieve all `tags` from all Assets

MarkLark86 (approved):

PR #2165, opened 2021-03-10
[OAUTH] send error data to frontend using javascript Message

jerome-poisson:
SDESK-5800


jerome-poisson:
This is an initial PR which will be updated when https://dev.sourcefabric.org/browse/SDESK-5794 will be taken, according to discussion with frontend dev.

petrjasek (approved):

PR #2167, opened 2021-03-17
FIX: Desk Default content template is not available in planning scheduled export. [SDCP-495]

GyanP:


petrjasek (approved):
test would be great

GyanP:
> test would be great

sure

features/desks.feature
```diff
@@ -754,3 +754,35 @@ Feature: Desks
                 ]
             }
         """
+    @wip
```
petrjasek, line 757: :eyes:

petrjasek (approved):

features/desks.feature
GyanP: just removed.

PR #2169, opened 2021-03-18
[SDESK-5890] Support timeouts for HTTP Push transmitter

MarkLark86:
@petrjasek This is to be back-ported to v1.32+

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -661,7 +661,7 @@ ENABLE_PROFILING = False
 FTP_TIMEOUT = 300
 
 #: default timeout when publishing using the `http_push` transmitter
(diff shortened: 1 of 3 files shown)


petrjasek:
I would use a lower value, or specify it as a tuple so when the host is not reachable it would fail faster,
something like `(5, 30)` or so. we could have config for the second number only, or maybe the whole tuple?

MarkLark86:
> I would use a lower value, or specify it as a tuple so when the host is not reachable it would fail faster,
> something like `(5, 30)` or so. we could have config for the second number only, or maybe the whole tuple?

I think a tuple works here. If the customer wants a different value, we can place it in their `settings.py`

petrjasek (approved):

PR #2173, opened 2021-03-22
As a SD editor I like to resend media files with new stories [SDESK-5781]

GyanP:


superdesk/publish/publish_queue.py
```diff
@@ -116,6 +137,43 @@ def _set_queue_state(self, updates, original):
...
+            return
+        article = get_resource_service("published").find_one(req=None, guid=doc["item_id"])
+        if not article:
+            return
+        return doc.get("publishing_action") == "corrected" or article.get("operation") == "being_corrected"
+
+    def resend_association_items(self, doc):
+        try:
+            associated_items = (
+                json.loads(doc.get("formatted_item", {})).get("associations", {}) if doc.get("formatted_item") else {}
```
petrjasek, line 159: would this work when using other formats like jimi in case of cp?

superdesk/publish/publish_queue.py
```diff
@@ -116,6 +137,43 @@ def _set_queue_state(self, updates, original):
...
+            )
+        except Exception:
+            return
+
+        if associated_items:
+            for key, associated_value in associated_items.items():
+                associated_article = get_resource_service("published").find_one(
+                    req=None, guid=associated_value.get("guid")
+                )
+                subscriber = get_resource_service("subscribers").find_one(req=None, _id=doc["subscriber_id"])
```
petrjasek, line 169: the subscriber where the doc is sent might not actually be the same where the media goes, in case of cp they have one subscriber for text items and another for media, the idea is to republish it when the text item is published and send it whereever it should go based on products/subscribers

superdesk/publish/publish_queue.py
```diff
@@ -97,6 +99,25 @@ def on_create(self, docs):
...
+
+        if not is_publish and not app.config.get("SENT_ONCE"):
+            for doc in docs:
+                if app.config.get("RESEND_ASSOCIATION_ITEMS_WITH_NEW_STORIES_AND_UPDATES_AND_CORRECTION"):
+                    self.resend_association_items(doc)
+                elif app.config.get(
+                    "RESEND_ASSOCIATION_ITEMS_WITH_NEW_STORIES_AND_UPDATES"
+                ) and not self.is_corrected_document(doc):
+                    self.resend_association_items(doc)
+                elif app.config.get("RESEND_ASSOCIATION_ITEMS_WITH_NEW_STORIES", True) and not (
```
petrjasek, line 114: I suppose there should be just 1 config with different values, eg:
```py
PUBLISH_RESEND_MEDIA = "never" | "updates" | "corrections"
```

superdesk/publish/publish_queue.py
```diff
@@ -116,6 +132,45 @@ def _set_queue_state(self, updates, original):
             else updates.get("state") or QueueState.PENDING.value
         )
 
+    def is_updated_document(self, doc):
+        if not doc.get("item_id"):
+            return
+        article = get_resource_service("published").find_one(req=None, guid=doc["item_id"])
```
petrjasek, line 138: not sure this will give you latest item in general, it can be any version of that published item

petrjasek (commented):
wouldn't this resend the media for each publish queue item? like if you send the same story to 3 subscribers, how many times it would republish those associated items? imo this should be handled before going to publish queue, in the enqueue probably..

GyanP:
yes, suppose we have 3 subscribers and all have products inside.
So if we send the same story to 3 subscribers then their associates will resend and a total of 9 items will be resend. you want to resend the associate items 3 times right if I click on resend then only 3 stories should be resend and those 3 items will be associated right?

petrjasek:
yep that's right. it should send the story 3x and each associated item 3x (if all products match)

apps/publish/enqueue/enqueue_service.py
```diff
@@ -254,11 +254,17 @@ def publish(self, doc, target_media_type=None, content_type=None):
         :raises PublishQueueError.item_not_queued_error:
                 If the nothing is queued.
         """
+        is_publish = True
```
petrjasek, line 257: imo the name is a bit misleading, it's always publishing something, should be probably `first_version` or something

superdesk/publish/publish_queue.py
```diff
@@ -116,6 +130,32 @@ def _set_queue_state(self, updates, original):
...
+        if not (doc.get("item_id") and doc.get("item_version", "version")):
+            return
+        article = get_resource_service("published").find_one(
+            req=None, guid=doc["item_id"], _current_version=doc.get("item_version", "version")
+        )
+        if not article:
+            return
+        return doc.get("publishing_action") == "corrected" or article.get("operation") == "being_corrected"
+
+    def resend_association_items(self, doc):
```
petrjasek, line 143: I'm wondering why it's actually done in publish queue, seems like we're splitting the logic in 2 places, but why?
seems like all the info is already there, but it goes in publish queue to call the enqueue service eventually

GyanP:
We have resend option in publish_queue also and publish_queue do not use the enqueue service, in all other scenario enqueue service is called except resend option in publish_queue.

superdesk/publish/publish_queue.py
GyanP: We have resend option in publish_queue also and publish_queue do not use the enqueue service, in all other scenario enqueue service is called except resend option in publish_queue.

petrjasek: resend action we have now in the ui has nothing to do with this really, doesn't it?

GyanP: when we do resend item(3 dot menu) or publish then first enqueue service is called after that publish queue but when we resend (from publish queue list) then directly publish queue called.
that is the reason we implemented that in publish queue instead of enqueue service as in all the cases whether it is resend items (3 dot menu) or publish or send correction or resend (publish queue list) publish queue will always be called.
after that, if we have config and associated items then using enqueue service publish method we will resend association as well.

petrjasek: I think this shouldn't change the behaviour of resend called from publish queue, that one should always just resend that very item you click on, and not its associations

GyanP: So that means resend (from publish queue list) will resend only that main item but not the associated items?

petrjasek: right

apps/publish/enqueue/enqueue_service.py
```diff
@@ -466,9 +473,19 @@ def queue_transmission(self, doc, subscribers, subscriber_codes=None, associatio
...
 
         try:
+            if config.PUBLISH_ASSOCIATIONS_RESEND and not publish_first_version:
+                if config.PUBLISH_ASSOCIATIONS_RESEND == "corrections":
+                    self.resend_association_items(doc)
+                elif config.PUBLISH_ASSOCIATIONS_RESEND == "updates" and doc.get("state") not in [
+                    "corrected",
+                    "being_corrected",
+                ]:
+                    self.resend_association_items(doc)
```
petrjasek, line 483: would it actually resend the associations if it's completely new item with already published associations?

petrjasek, line 483: would it actually resend the associations if it's completely new item with already published associations?

GyanP: yes, if we have a new story with already published associations then it will resend associations.

petrjasek (approved):

apps/publish/enqueue/enqueue_service.py
```diff
@@ -555,6 +572,37 @@ def queue_transmission(self, doc, subscribers, subscriber_codes=None, associatio
...
+        for item in associated_items.values():
+            item_id = item.get("_id")
+            if item_id and item_id not in associations.keys():
+                associations[item_id] = item
+        return associations
+
+    def resend_association_items(self, doc):
+        associated_items = doc.get(ASSOCIATIONS)
+        if associated_items:
+            associated_items = self.remove_duplicate_association(associated_items)
```
petrjasek, line 586: you're changing the structure here from `assoc-name: item` to `item-id: item` which is a bit error prone imo.
maybe something like `unique_associations = self.get_unique_associations(associated_items)` and return `associations.values()` from that method, you don't need the keys later

GyanP: yes, we don't need keys later so I will rename the `remove_duplicate_association` to `get_unique_associations` and return the `associations.values()`.

apps/publish/enqueue/enqueue_service.py
```diff
@@ -555,6 +572,36 @@ def queue_transmission(self, doc, subscribers, subscriber_codes=None, associatio
...
+    def resend_association_items(self, doc):
+        associated_items = doc.get(ASSOCIATIONS)
+        if associated_items:
+            for association in self.get_unique_associations(associated_items):
+                # resend only media association
+                if association.get("type") not in MEDIA_TYPES:
+                    continue
+
+                archive_article = get_resource_service("archive").find_one(
+                    req=None, _id=association.get("guid", association.get("_id"))
```
petrjasek, line 592: it's better to use `_id` for the `_id` lookup, I think with some media we had it different so guid won't match _id

petrjasek (approved):

PR #2184, opened 2021-03-31
Update NTB NITF publishing formats with new metadata from iMatrics. [SDNTB-647]

GyanP:
Add the place field which is present in the Imatrics widgets.

### DIFF
--- a/apps/content_types/content_types.py
+++ b/apps/content_types/content_types.py
@@ -562,7 +562,7 @@ def apply_schema(item):
     :param item: item to apply schema to
     """
     # fields that can be added to article without being added to CP eg: using widgets
-    allowed_keys = ["attachments", "refs"]
+    allowed_keys = ["attachments", "refs", "place"]
 
     if item.get("type") == "event":
         return item.copy()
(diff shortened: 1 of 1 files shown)


apps/content_types/content_types.py
```diff
@@ -562,7 +562,7 @@ def apply_schema(item):
     :param item: item to apply schema to
     """
     # fields that can be added to article without being added to CP eg: using widgets
-    allowed_keys = ["attachments", "refs"]
+    allowed_keys = ["attachments", "refs", "place"]
```
petrjasek, line 565: what about `organisation` ?

GyanP: we don't need the organisation or person because if it's not in the content profile then we do not need them anywhere, However that's not the case for place, if place is not present in the content profile then also we need it.

petrjasek: imatrics return organisation don't they? and we need those in the output

petrjasek: I would rather add all those imatrics fields now so we don't have to create core release later for that

petrjasek (approved):

PR #2190, opened 2021-04-12
Default 'coming up' time should be empty in update [SDBELGA-493]

GyanP:


apps/archive/archive_rewrite.py
```diff
@@ -279,6 +279,10 @@ def _create_rewrite_article(self, original, existing_item=None, desk_id=None):
             # if True then reset to the default priority value.
             rewrite["priority"] = int(config.DEFAULT_PRIORITY_VALUE_FOR_MANUAL_ARTICLES)
 
+        # Disable and empty date time for coming_up field
+        if original.get("extra", {}).get("DueBy"):
+            rewrite["extra"]["DueBy"] = None
```
petrjasek, line 284: we shouldn't hardcode that belga field name in the core, we need to find another way via config/hook

PR #2197, opened 2021-04-21
change the corrections workflow to show items with corrected state in the Desk output where they were published [SDCP-516]

GyanP:


apps/publish/content/correct.py
```diff
@@ -114,7 +114,7 @@ def on_update(self, updates, original):
         update_word_count(updates, original)
         flush_renditions(updates, original)
         self.change_being_corrected_to_published(updates, original)
-        self.send_to_original_desk(updates, original)
+        self.send_to_current_desk(updates, original)
```
petrjasek, line 117: do we actually have to send it anywhere? it should be already there

GyanP: Good catch.

petrjasek (approved):

PR #2203, opened 2021-04-29
As a user I want to have the Coming Up field active when I create a new story. [SDBELGA-466]

GyanP:
use jinja2 templates for the current date-time.

### DIFF
--- a/features/templates.feature
+++ b/features/templates.feature
@@ -720,7 +720,7 @@ Feature: Templates
           }]
         }
         """
-
+    @wip
     @auth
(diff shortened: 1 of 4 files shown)


features/templates.feature
```diff
@@ -720,7 +720,7 @@ Feature: Templates
           }]
         }
         """
-
+    @wip
```
petrjasek, line 723: :eyes:

GyanP: My mistake.

petrjasek (approved):

PR #2207, opened 2021-05-06
[MACROS] expose `replace_type` (#2159)

petrjasek:
`replace_type` was not exposed by the API, as a result the client could
not apply the item update correctly.

fix SDBELGA-460


petrjasek:
@jerome-poisson I saw we have this branch which was not merged yet, do we need it still?

jerome-poisson:
yes I think we still do, this field is checked in frontend, and without it some macro may not be executed.

jerome-poisson (approved):

PR #2208, opened 2021-05-11
Check AP media API requests format [SDESK-5966]

GyanP:


marwoodandrew (commented):
I'm don't believe this needs to be done, I think the requests library will take care of it!

petrjasek:
yep that's right, it does encode automatically:
```py
>>> import requests
>>> requests.Request('GET', 'http://example.com', params={"test": "1<2"}).prepare().url
'http://example.com/?test=1%3C2'
```

PR #2212, opened 2021-05-20
FIX: Changes made to a related item are not saved after publish [SDESK-5972]

GyanP:


petrjasek (commented):
test pls

features/content_publish.feature
```diff
@@ -4599,3 +4599,114 @@ Feature: Content Publishing
...
+        """
+      When we publish "123" with "publish" type and "published" state
+      Then we get OK response
+      When we get "/archive/234"
+      Then we get existing resource
+        """
+        {
+            "_id": "234",
+            "guid": "234",
+            "slugline": "related item test",
```
petrjasek, line 4704: so without the fix here you would get `related item` ?

GyanP: Yes, we will get it.

petrjasek (approved):

PR #2213, opened 2021-05-28
add additional metadata to imatrics analyzer

petrjasek:
SDNTB-692

### DIFF
--- a/superdesk/text_checkers/ai/imatrics.py
+++ b/superdesk/text_checkers/ai/imatrics.py
@@ -75,12 +75,14 @@ class IMatrics(AIServiceBase):
             "name": concept["title"],
             "qcode": concept["uuid"],
             "source": "imatrics",
+            "aliases": concept["aliases"],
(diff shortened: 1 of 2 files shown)


superdesk/text_checkers/ai/imatrics.py
```diff
@@ -149,13 +152,18 @@ def analyze(self, item: dict) -> dict:
             "pubStatus": False,
             "headline": headline,
             "body": body,
+            "language": item["language"],
```
jerome-poisson, line 155: are we sure that `language` is always set here?

jerome-poisson (approved):

superdesk/text_checkers/ai/imatrics.py
petrjasek: it is https://github.com/superdesk/superdesk-core/blob/develop/apps/archive/common.py#L190-L191

PR #2217, opened 2021-06-01
[FEEDING SERVICE][GMAIL]Set email on log-in + log-out button

jerome-poisson:
This patch needs a modification in client to save the provider before the
`Log-in with GMail` button can be used.

When user is successfully logged-in, the email address is set to the
corresponding feeding service.

The `Log-in with GMail` is now only shown if user is not already
logged-in (i.e. if we don't have an e-mail address set).

When user is logged-in, a `Log-out` button is shown instead, using it
will revoke the OAuth token, delete it from mongo and remove the email
from ingest provider configuration.

SDESK-5800

### DIFF
--- a/superdesk/auth/oauth.py
+++ b/superdesk/auth/oauth.py
@@ -182,19 +182,14 @@ def configure_google(app, extra_scopes: Optional[List[str]] = None, refresh: boo
                         '"security/Third-party apps with account access") then try to log-in again'
                     )
 
-
             # token_id is actually the provider id
             ingest_providers_service = superdesk.get_resource_service("ingest_providers")
             provider = ingest_providers_service.find_one(req=None, _id=token_id)
             if provider is None:
-                logger.warning(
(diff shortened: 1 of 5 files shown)


jerome-poisson:
**NOTE:** to be coordinated with @tomaskikutis before merging.

petrjasek (approved):

jerome-poisson:
@tomaskikutis I've fixed the token retrieval issue when logging out but there are other issue notably etag mismatch when we log-out. I see that we're going back from holidays approximately at the same time, so maybe we can fix that then.

@petrjasek I've let the `WIP` field because of the issue mentioned with Tomas, but otherwise this PR can be merge if necessary. As the frontend part has been merged, it may be needed to have the changes from Tomas.

jerome-poisson:
After second thought, I'm removing the `WIP` flag, I'd rather merge this PR to move on and fix issues in a new one.

PR #2218, opened 2021-06-01
add last id param to app:index_from_mongo task [SDESK-5992]

GyanP:


GyanP:
means we have to convert `last_id` to `ObjectId(last_id)` when we use in a filter, but I think archive collection takes directly` last_id`?

superdesk/commands/index_from_mongo.py
```diff
@@ -67,6 +68,9 @@ def copy_resource(cls, resource, page_size):
                 else:
                     break
 
+            if last_id:
+                print(last_id)
```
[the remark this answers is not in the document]
GyanP: yep, I will print last_id after the filter we are fetching from items like `last_id = items[-1][config.ID_FIELD]` so that we will get the last_id of the items processed every time.

petrjasek:
> means we have to convert `last_id` to `ObjectId(last_id)` for the filter args but I think archive collection takes directly` last_id`?

yep it's only needed for certain collections so you can do:
```py
try:
    last_id = ObjectId(last_id)
except ValueError:
    pass
```

superdesk/commands/index_from_mongo.py
```diff
@@ -68,10 +69,11 @@ def copy_resource(cls, resource, page_size, last_id):
                 else:
                     break
 
-            if last_id:
-                print(last_id)
-
             print("{} Inserted {} items in {:.3f} seconds".format(time.strftime("%X %x %Z"), success, time.time() - s))
+
+            # print last ID for items
+            print(items[-1][config.ID_FIELD])
```
petrjasek, line 75: I think this actually prints after each bucket but you only want to do it at the very end when this `for items ...` cycle is done

superdesk/commands/index_from_mongo.py
```diff
@@ -92,6 +94,11 @@ def get_mongo_items(cls, mongo_collection_name, page_size, last_id):
         db = app.data.get_mongo_collection(mongo_collection_name)
         args = {"limit": bucket_size, "sort": [(config.ID_FIELD, pymongo.ASCENDING)]}
 
+        try:
+            last_id = ObjectId(last_id)
```
petrjasek, line 98: you should do it only if last_id is set because `ObjectId(None)` generates a valid ObjectId

superdesk/commands/index_from_mongo.py
```diff
@@ -36,23 +37,24 @@ class IndexFromMongo(superdesk.Command):
...
-    def run(self, collection_name, all_collections, page_size):
+    def run(self, collection_name, all_collections, page_size, last_id):
         if not collection_name and not all_collections:
             raise SystemExit("Specify --all to index from all collections")
         elif all_collections:
             app.data.init_elastic(app)
             resources = app.data.get_elastic_resources()
             for resource in resources:
-                self.copy_resource(resource, page_size)
+                self.copy_resource(resource, page_size, last_id)
```
petrjasek, line 51: imo it only makes sense to use the last id with single resource, so if it's set with all collections it might stop

superdesk/commands/index_from_mongo.py
```diff
@@ -87,10 +90,17 @@ def get_mongo_items(cls, mongo_collection_name, page_size):
...
+
         while True:
             if last_id:
+                try:
+                    last_id = ObjectId(last_id)
+                except Exception:
+                    pass
                 args.update({"filter": {config.ID_FIELD: {"$gt": last_id}}})
+                # print last ID for items
+                print(last_id)
```
petrjasek, line 102: is this temporary?

GyanP: Actually, as per the ticket we have to print` last_id` and I added it here, so we will get the last_id of the items processed every time.

petrjasek: not sure we need it every time, would be enough at the end. and pls add an info regarding what resource is it

superdesk/commands/index_from_mongo.py
```diff
@@ -87,10 +90,17 @@ def get_mongo_items(cls, mongo_collection_name, page_size):
...
+
             cursor = db.find(**args)
             if not cursor.count():
                 break
```
petrjasek, line 106: imo that print should go here:
```py
if not cursor.count():
    print("last id", mongo_collection_name, last_id)
    break
```

petrjasek (approved):

PR #2224, opened 2021-06-16
FIX: New corrections workflow: items get stuck during corrections workflow [SDCP-551]

GyanP:


petrjasek (approved):
would be good to handle those error cases we have so those items are not stuck there

apps/archive/archive_correction.py
```diff
@@ -95,17 +95,23 @@ def on_update(self, updates, original):
...
             # modify item in published.
-            published_service.system_update(
+            _published_item = published_service.system_update(
                 published_article.get(config.ID_FIELD), published_item_updates, published_article
             )
+            if _published_item is None or (
+                published_item_updates.get("operation") == CONTENT_STATE.BEING_CORRECTED
+                and _published_item.get("state") != CONTENT_STATE.BEING_CORRECTED
+            ):
+                raise SuperdeskApiError.badRequestError(message=_("Being corrected item is not generated"))
```
petrjasek, line 105: not sure this is bad request error, should be more like assert error

petrjasek: eg:
```py
assert _published_item is not None and (...)
```

GyanP: I think we can use `preconditionFailedError` instead of `badRequestError`  or assert is more convenient?

petrjasek: precondition is used for etag validation, do you think it could be an issue here?

GyanP: No, I think we will not get etag issue. I will add changes with assert.

petrjasek (approved):

PR #2231, opened 2021-06-29
FIX: Resend media association items [SDESK-5781]

GyanP:


superdesk/default_settings.py
```diff
@@ -968,3 +968,4 @@ def local_to_utc_hour(hour):
 #: .. versionadded:: 2.1
 #:
 APPS_DATA_UPDATES_PATHS = []
+PUBLISH_ASSOCIATIONS_RESEND = "default"
```
petrjasek, line 971: can you call the default value `new` as `default` doesn't say much about the behaviour and document in the comment above the other options?

apps/publish/enqueue/enqueue_service.py
```diff
@@ -254,9 +254,11 @@ def publish(self, doc, target_media_type=None, content_type=None):
         :raises PublishQueueError.item_not_queued_error:
                 If the nothing is queued.
         """
-        publish_first_version = True
-        if doc.get("rewrite_of") or doc.get("state") == "corrected" or doc.get("translated_from"):
-            publish_first_version = False
+        publish_first_version = False
+        if target_media_type and config.PUBLISH_ASSOCIATIONS_RESEND:
```
petrjasek, line 258: what is this case? could you add a comment?

apps/publish/enqueue/enqueue_service.py
```diff
@@ -254,9 +254,11 @@ def publish(self, doc, target_media_type=None, content_type=None):
         :raises PublishQueueError.item_not_queued_error:
                 If the nothing is queued.
         """
-        publish_first_version = True
-        if doc.get("rewrite_of") or doc.get("state") == "corrected" or doc.get("translated_from"):
-            publish_first_version = False
+        publish_first_version = False
```
petrjasek, line 257: what does this mean? new item being published? maybe `new_item` would be better name

apps/publish/enqueue/enqueue_service.py
```diff
@@ -474,7 +476,13 @@ def queue_transmission(
 
         try:
             if config.PUBLISH_ASSOCIATIONS_RESEND and not publish_first_version:
-                if config.PUBLISH_ASSOCIATIONS_RESEND == "corrections":
+                if (
+                    config.PUBLISH_ASSOCIATIONS_RESEND == "default"
+                    and not doc.get("rewrite_of")
+                    and doc.get("state") not in ["corrected", "being_corrected"]
```
petrjasek, line 482: would be more readable if you store that before those conditions:
```py
is_correction = doc["state"] in ["...]
is_update = doc.get("rewrite_of")
is_new = not is_correction and not is_update
if config.PUBLISH_ASSOCIATIONS_RESEND == "new" and is_new:
   ...
```

apps/publish/enqueue/enqueue_service.py
petrjasek: so if `target_media_type` is set it shouldn't even transmit the item?

apps/publish/enqueue/enqueue_service.py
```diff
@@ -475,19 +475,16 @@ def queue_transmission(
             subscriber_codes = {}
 
         try:
-            if config.PUBLISH_ASSOCIATIONS_RESEND and not publish_first_version:
-                if (
-                    config.PUBLISH_ASSOCIATIONS_RESEND == "default"
-                    and not doc.get("rewrite_of")
-                    and doc.get("state") not in ["corrected", "being_corrected"]
-                ):
+            if config.PUBLISH_ASSOCIATIONS_RESEND and not new_item:
```
petrjasek, line 478: with that I'm not sure how does it resend for new items?

GyanP: Yes, it will resend a new item also but I think we should keep its name `publish_first_version` as we discussed earlier https://github.com/superdesk/superdesk-core/pull/2173#discussion_r613023888 so that it will not make any confusion.

petrjasek: where does it resend new items?

GyanP: `new_item` is just a bool variable to prevent from resending the same association multiple times and as per the current behavior `media items should just be sent once (current behavior)` which means we are already resending media association for new items.

petrjasek: so it's like if we have same file attached multiple times?

GyanP: No, Suppose we have a **new story with only one association** and as per the `current behavior` it is already resending the media association means we have 2 entries in the publishing queue. after that, if config PUBLISH_ASSOCIATIONS_RESEND is set then from the main story we are fetching association and resending again then we will have one more entry in the publishing queue. so total we have 3 entries(main story, association, again association) instead of 2(main story, association).so prevent extra entries I have added few checks.

petrjasek: I see but does it resend for new items if the associations was already published before?

GyanP: yes, it will resend. I have cross-checked it.

petrjasek: ok, so maybe calling that param `sent` would be better than, I thought it's about new item being published

apps/duplication/archive_translate.py
```diff
@@ -23,6 +23,7 @@
 from superdesk.utc import utcnow
 from apps.packages import PackageService
 from flask_babel import _
+from eve.utils import config
```
MarkLark86, line 26: I can't see where this is used. Is this import needed?

apps/publish/content/resend.py
```diff
@@ -48,6 +48,14 @@ def create(self, docs, **kwargs):
         article_version = doc.get("version")
         article = self._validate_article(article_id, article_version)
         subscribers = self._validate_subscribers(doc.get("subscribers"), article)
+        if config.PUBLISH_ASSOCIATED_ITEMS:
```
MarkLark86, line 51: This can be replaced with the `remove_is_queued` function that was added

MarkLark86 (approved):

PR #2236, opened 2021-07-13
Languages content filter does not work [SDESK-6049]

GyanP:


petrjasek:
does it get values from languages cv?

petrjasek:
add test pls

GyanP:
> does it get values from languages cv?

Yes, I think because during setup of content filter condition we use the languages cv and it will work like we will get the "language" value from the article and check with content filter condition langauge value.

petrjasek (approved):

PR #2237, opened 2021-07-13
FIX: resend media associations on send correction when config value is 'corrections'[SDESK-5781]

GyanP:


petrjasek:
what is this fixing actually?

GyanP:
Resend media association file into publish queue, when we correct the main text story and do `send correction`.

petrjasek:
I still don't get it, you call now `remove_is_queued`, so what does it do differently?

petrjasek (approved):

PR #2241, opened 2021-07-20
Provide the highlighting of search results also in the Monitoring view [SDBELGA-522]

GyanP:


petrjasek:
why it's not working on archive resource? I guess we have the same issue on ingest/published resources. would be good to fix the cause of it

GyanP:
> why it's not working on archive resource? I guess we have the same issue on ingest/published resources. would be good to fix the cause of it

The archive resource directly `get` request is calling and it is not adding the `highlight class` but on search resource we already implemented that functionality so to reuse the code I used `search` request.

GyanP:
> why it's not working on archive resource? I guess we have the same issue on ingest/published resources. would be good to fix the cause of it

while search(in monitoring view) directly `get` request is calling for the archive resource and we don't have highlight class that time but in search, we have already implemented functionality for that one so to reuse the code I directly called the search resource.

petrjasek:
what functionality you mean?

GyanP:
> what functionality you mean?

On search two APIs are calling **search** and **archive** so when `search service` is working and we are getting highlighted text but when the `archive service` is working we are not getting highlighted text because on `search service `they are modifying query_string  and adding the highligh_query for some fields so to reuse all the code and its functionality I directly called the `search service`.

petrjasek:
I see, I would rather make that part reusable, because using the search service you can get content from all collections while it should only return archive content

GyanP:
As per my observation, I think items( like in progress, ingest, etc), archive service is responsible and for the item(published, corrected), search service is responsible for highlighting text.

petrjasek:
archive service only returns content from archive collection, search can handle that plus ingest/published/archived

petrjasek (commented):
pls add a test and would it be possible to avoid that code duplication between search service and archive ?

petrjasek (approved):

PR #2243, opened 2021-07-22
Search on sign-off in monitoring view (marked for user) [SDBELGA-417]

GyanP:


petrjasek (commented):
when do you actually set it?

GyanP:
> when do you actually set it?

on client-side, when we do mark for user we send marked_for_sign_off along with marked_for_user.

petrjasek (approved):

PR #2244, opened 2021-07-28
SDNTB-700 ADD: Implement FTP ingest service in planning.

dev-tsoni:


petrjasek (approved):

petrjasek:
btw would the standard file feed ingest work now with events json?

MarkLark86:
> btw would the standard file feed ingest work now with events json?

It will work with Events JSON, but It won't for the ICS parser. Will need some changes to File, HTTP & Email Feeding Services in core, then we should be able to remove those Event specific services from Planning

dev-tsoni:
> > btw would the standard file feed ingest work now with events json?
> 
> It should work with Events JSON, but It won't for the ICS parser. Will need some changes to File, HTTP & Email Feeding Services in core, then we should be able to remove those Event specific services from Planning

We can add these changes altogether in single ticket.

petrjasek:
lets do the ftp first, that's what we need now for qa

superdesk/io/commands/update_ingest.py
```diff
@@ -697,7 +697,7 @@ def get_ingest_collection(feeding_service, item):
         ingest_collection = feeding_service.service
 
     # If the type of item is event, set the collection to events
-    elif item.get("type") and item.get("type") == "event":
+    elif item.get(ITEM_TYPE) and item.get(ITEM_TYPE) == "event":
```
petrjasek, line 700: imo `elif item.get(ITEM_TYPE) == "event"` would be enough

petrjasek (approved):

PR #2249, opened 2021-08-06
Add read_only field used in original_metadata cv for vocabularies [SDBELGA-555]

GyanP:


GyanP:
As an editor, I like to see and search based on all keywords coming with the ingest sources AFP, ANP, and DPA [SDBELGA-555], added new cv  original_metadata, and to make it read-only updated vocabularies resource.

petrjasek:
we can already set readonly on content profile for a field, wouldn't it work?

GyanP:
> we can already set readonly on content profile for a field, wouldn't it work?

No, imo because if have any new field like here I have added the `"read_only": true` for cv **original_metadata**  in `vocabualries.json` but as it is not defined in the `vocabularies resource` we are not able to fetch and use it. so we have to add this here.

GyanP:
> we can already set readonly on content profile for a field, wouldn't it work?

sorry I have updated my comment and to address this https://github.com/superdesk/superdesk-client-core/pull/3958#issuecomment-893376096. I did changes.

petrjasek (approved):

PR #2251, opened 2021-08-10
FIX: Content filters with more than one condition do not work - corresponding article does not match neither content filter nor product filter [SDESK-6127]

GyanP:


apps/content_filters/filter_condition/filter_condition_field.py
```diff
@@ -150,11 +150,11 @@ def get_value(self, article):
 class FilterConditionGenreField(FilterConditionField):
     def __init__(self, field):
         self.field = FilterConditionFieldsEnum.genre
-        self.entity_name = "genre.name"
+        self.entity_name = "genre.qcode"
```
MarkLark86, line 153: Does this mean the `Genre` `Filter Condition` has not been working for quite some time?

GyanP: Yep, if the genre `name` and its `qcode` is different, then it is not working because whenever we create any filter condition related to genre, it stores `qcode` not `name`.

PR #2252, opened 2021-08-11
[SDESK-6123] Use query to delete audits in purge command

MarkLark86:


GyanP (approved):

superdesk/audit/commands.py
```diff
@@ -125,26 +126,14 @@ def purge_orphaned_item_audits(self):
...
-                logger.info("Finished purging audit logs at {}".format(utcnow()))
-                return
-            logger.info("Found {} audit items at {}".format(len(items), utcnow()))
-            service.delete_ids_from_mongo(items)
-        logger.warning("Audit purge didn't finish in 100 iterations.")
+        service = superdesk.get_resource_service("audit")
+        time_start = time()
+        service.delete_from_mongo({"_id": {"$lt": ObjectId.from_datetime(self.expiry)}})
+        time_diff = time() - time_start
+        logger.info(f"Finished purging audit logs. Took {time_diff:.4f} seconds")
```
petrjasek, line 136: fyi we have this [timer](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/timer.py) so it could be

```py
with timer("audit expiry"):
    service.delete_from_mongo(...)
```

PR #2258, opened 2021-08-23
[CI] Install dependencies before CodeQL intialisation

jerome-poisson:
CodeQL auto python setup was failing, this patch install missing
dependencies before doing it.

SDESK-6148

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -35,7 +35,6 @@ jobs:
     - name: Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get upgrade -y
         sudo apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl
 
     # Initializes the CodeQL tools for scanning.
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -32,11 +32,18 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v2
 
+    - name: Python Dependencies
+      run: |
+        sudo apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl
```
petrjasek, line 37: it's good to add apt-get update first otherwise this breaks when those packages get updated

.github/workflows/codeql-analysis.yml
```diff
@@ -32,6 +32,12 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v2
 
+    - name: Dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get upgrade -y
```
petrjasek, line 38: I wouldn't do the upgrade here, that could potentially upgrade the whole system including kernel etc

petrjasek (approved):

PR #2261, opened 2021-08-24
fix extra fields handling on rewrite

petrjasek:
- don't override extra fields when empty
- sync extra dict first to populate any missing values

SDCP-561

### DIFF
--- a/superdesk/editor_utils.py
+++ b/superdesk/editor_utils.py
@@ -10,6 +10,7 @@
 """This module contains tools to manage content for Superdesk editor"""
 
 import re
+from typing import Dict
 import uuid
 import logging
 import lxml.etree as etree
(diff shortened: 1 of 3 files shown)


superdesk/editor_utils.py
```diff
@@ -819,3 +820,15 @@ def render_fragment(elem) -> str:
         # client renders empty paragraph as `<p><br></p>`
         etree.SubElement(elem, "br", nsmap=None, attrib=None)
     return str(lxml_html.tostring(elem, encoding="unicode"))
+
+
+def is_empty_content_state(item: Dict, field: str) -> bool:
```
jerome-poisson, line 825: Could you add a docstring here?

superdesk/editor_utils.py
```diff
@@ -819,3 +820,15 @@ def render_fragment(elem) -> str:
...
+    content_state = get_field_content_state(item, field)
+    return content_state is None or not any([block.get("text", "").strip() for block in content_state["blocks"]])
+
+
+def copy_fields(source: Dict, dest: Dict, ignore_empty=False):
```
jerome-poisson, line 830: Same, please as a docstring here. Even if we can guess with names, it would be useful to know what `ignore_empty` does exactly (is it that key is missing, set to `None` or something else?).

jerome-poisson (approved):

PR #2269, opened 2021-09-07
fix(build): Upgrade feedparser to 6.0.x

MarkLark86:
Upgrading `feedparser` to `6.0.x` so that the build does not fail with the latest `setuptools`.
Setuptools changelog:
```
v58.0.2
06 Sep 2021

Misc
#2769: Build now fails fast when use_2to3 is supplied.
```

Which `feedparser 5.x` contained the `use_2to3` in their `setup.py` file (https://github.com/kurtmckee/feedparser/blob/5.2.1/setup.py#L6).


MarkLark86:
`fire:build` still fails because it is trying to `pip install` from `superdesk-core:develop` first, then this PR:

From the build.sh generated file:
```bash
cd /opt/superdesk/server
time pip install -r requirements.txt

cd /opt/superdesk/server-core
time pip install -Ur requirements.txt
```

petrjasek (approved):

PR #2278, opened 2021-09-28
[SDESK-6146] Ability to authenticate using cookies or auth header

MarkLark86:
We can use https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/auth/auth.ts#L233 watches to send `POST` request to `/api/auth_cookie` to either create or delete the `session_token` cookie (PR to come soon with other changes)


apps/auth/auth_cookie.py
```diff
@@ -0,0 +1,42 @@
...
+from datetime import timedelta
+from flask import make_response, current_app as app, request, g
+from superdesk import Blueprint
+from superdesk.utc import utcnow
+
+auth_cookie_bp = Blueprint("auth_cookie_bp", __name__)
+
+
+@auth_cookie_bp.route("/auth_cookie", methods=["POST", "OPTIONS"])
+def get_auth_test_post():
```
petrjasek, line 20: imo this method can use the `blueprint_auth` decorator as well
and we could use `DELETE` method to delete cookie

PR #2279, opened 2021-09-29
[SDESK-6206] Ability to authenticate using flask session cookies

MarkLark86:
Requires https://github.com/superdesk/superdesk-client-core/pull/4001 otherwise cookies won't be stored in the browser

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -434,8 +434,10 @@ The number of days an activation token is valid, env ``ACTIVATE_TTL``.
 ^^^^^^^^^^^^^^
 
 .. versionadded:: 1.5
+.. versionchanged:: 2.4
+    Now required as ``flask.session`` will be used for some authentication
 
-Default: ``''``
+Default: Hardcoded value, which should not be used in production!
 
 This value should be set to a unique, unpredictable value. It is used for auth token signing.
 
--- a/apps/auth/auth.py
+++ b/apps/auth/auth.py
@@ -13,7 +13,7 @@ import logging
 import superdesk
 
 from datetime import timedelta
-from flask import request, current_app as app, json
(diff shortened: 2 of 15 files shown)


docs/settings.rst
```diff
@@ -434,8 +434,10 @@ The number of days an activation token is valid, env ``ACTIVATE_TTL``.
 ^^^^^^^^^^^^^^
 
 .. versionadded:: 1.5
+.. versionchanged:: 2.4
+    Now required as ``flask.session`` will be used for some authentication
 
-Default: ``''``
+Default: Hardcoded value, which should not be used in production!
```
petrjasek, line 440: I think we shouldn't have any default here, this must be set via environment.

MarkLark86: If we don't put a default, tests/fireq/customer instances will all start failing.

I can change this if required, but will need to update all the affected tests/instances as well (probably under a separate ticket as it will be a big task)

petrjasek: I can handle fireq, that's not a big deal, for tests we can set something hardcoded, but otherwise we should require this to be set

petrjasek: but we can probably keep it, seems consistent with newsroom

MarkLark86: I actually copied this code from NewsHub (except generated a new secret key), as I wanted to see how we handle it there.

I do agree it's not the best to have a hard-coded default for this, and not defining this value is an invalid config which should not proceed to run.

What if we merge this for now and create a ticket to tackle both Superdesk & NewsHub together?

petrjasek: yep makes sense

petrjasek (approved):

PR #2292, opened 2021-10-28
FIX: Embed field's description not sent if empty [SDESK-6217]

GyanP:


petrjasek (commented):
I'm not sure it's a bug actually, but if it's really needed would rather handle it in the output. what does `"embed" in value means anyway? if we have extra text field and type there embed it would raise TypeError probably

GyanP:
> I'm not sure it's a bug actually, but if it's really needed would rather handle it in the output. what does `"embed" in value mean anyway? if we have an extra text field and type their embed it would raise TypeError probably

Yep it is not a bug, instead of it we can say that it is a kind of improvement. "embed is the key to identifying that field". Any suggestion from your side?

petrjasek:
you could check if the value is dict and has an embed key, but would still rather do it in the ninjs formatter so we have it working for existing items as well

apps/archive/archive.py
```diff
@@ -439,6 +439,12 @@ def on_update(self, updates, original):
             publish_from_personal = flask.request.args.get("publish_from_personal") if flask.request else False
             update_state(original, updates, publish_from_personal)
 
+        # set description for custom embed field
+        for key, value in updates.get("extra", {}).items():
+            if type(value) == dict and "embed" in value:
+                value.setdefault("description", "")
+                break
```
petrjasek, line 446: not sure about this break, if there are more embed field it would only handle 1, why?

GyanP: My bad, forget to remove this.

petrjasek:
I would still rather handle it in the output, it's kind of hard to figure out why do we do it if we don't need it

petrjasek:
meaning in the ninjs formatter

petrjasek (approved):

PR #2294, opened 2021-11-03
FIX: For scheduled items sent to internal destination, unexpected duplicate is created in internal destination [SDESK-6239]

GyanP:
It will also resolve the issue of #SDBELGA-594

### DIFF
--- a/features/internal_destinations.feature
+++ b/features/internal_destinations.feature
@@ -90,6 +90,8 @@ Feature: Internal Destinations
         """
         Then we get OK response
 
+        When we get "/archive"
+        Then we get list with 0 items
         When we get "/published"
         Then we get list with 1 items
         """
@@ -702,6 +704,8 @@ Feature: Internal Destinations
         }
         """
         Then we get OK response
+        When we get "/published?where=%7B%22processed_from%22%3A%22123%22%7D"
+        Then we get list with 0 items
         When we get "/published"
         Then we get list with 1 items
 
--- a/features/internal_destinations.feature
+++ b/features/internal_destinations.feature
@@ -90,22 +90,6 @@ Feature: Internal Destinations
         """
         Then we get OK response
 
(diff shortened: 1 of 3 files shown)


petrjasek:
how does this create a duplicate item? I guess it's rather some code using the signal doing it

GyanP:
> how does this create a duplicate item? I guess it's rather some code using the signal doing it

IMO, the internal destination should work only while publishing API calls. I investigated and found whenever we schedule an item for publishing. firstly that publish API calls and create new items using internal destination. I noticed that part works only when we schedule the item here https://github.com/superdesk/superdesk-core/blob/develop/apps/publish/enqueue/__init__.py#L121 so again it calls internal destination and created duplicate item.

petrjasek:
so that signal is called 2x?

GyanP:
> so that signal is called 2x?

on normal publishing(when publishing schedule is not), that signal not calls but when we set publish schedule then it calls. 
so if any item is scheduled for publishing
1. while we click on publish, an internal destination calls and create a new item, at that signal is calls
2. while actually that item is published once again internal destination calls and create more item which is kind of duplicate
for 2 points that signal call is responsible for creating a duplicate item.

petrjasek:
might be better to avoid that first signal then when item is scheduled so we trigger internal destination only when the item is really published

petrjasek (approved):

features/internal_destinations.feature
```diff
@@ -90,22 +90,6 @@ Feature: Internal Destinations
         """
         Then we get OK response
 
-        When we get "/archive"
-        Then we get list with 1 items
```
petrjasek, line 94: can you rather change it to `we get list with 0 items` so it's obvious this was intended

petrjasek:
can it still create an item when the item is scheduled if enabled via the config?

PR #2296, opened 2021-11-08
update docs requirements

petrjasek:


jerome-poisson (approved):
Looks good, we just need to be sure that plantuml is still there (we use it for some schematics), but according to https://gitmotion.com/readthedocs-org/15818359/support-for-uml-via-plantuml-sphinxcontrib-plantuml it's supported since `Ubuntu 18.04` so it's probably fine.

PR #2298, opened 2021-11-19
As a user I want to have the Coming Up field active when I create a new story. [SDBELGA-466]

GyanP:


petrjasek:
imo it's not exactly clear what this does from template perspective:
```
{{ date|iso_datetime(5) }}
```
might be better something like:
```
{{ date|timedelta(minutes=5)|iso_datetime }}
```

GyanP:
> ```
> {{ date|timedelta(minutes=5)|iso_datetime }}
> ```

Doesn't this will mean that we are applying two filters: 1) timedelta(min=5) 2) iso_datetime?

petrjasek:
yep that would be 2 filters

apps/templates/filters.py
```diff
@@ -63,10 +63,15 @@ def first_paragraph_filter(input_string):
     return ""
 
 
-def iso_datetime(date, initial_offset):
+def add_timedelta(date, minutes):
```
petrjasek, line 66: it could be with `**kwargs` so it's not limited to use minutes

apps/templates/filters.py
```diff
@@ -63,9 +63,9 @@ def first_paragraph_filter(input_string):
     return ""
 
 
-def add_timedelta(date, minutes):
+def add_timedelta(date, **kwargs):
     try:
-        return date + timedelta(minutes=minutes)
+        return date + timedelta(minutes=kwargs.get("minutes"))
```
petrjasek, line 68: it should be rather `timedelta(**kwargs)` here so there can be hours/seconds/whatever

GyanP: yep, I will do the changes.

features/templates.feature
```diff
@@ -736,7 +736,7 @@ Feature: Templates
                 "test": "{{ user.sign_off }}",
                 "test2": "{{ item.something.missing }}",
                 "test3": "{% if something %}",
-                "test4": "{{ now|iso_datetime }}"
+                "test4": "{{ now|add_timedelta(minutes=45)|iso_datetime }}"
```
petrjasek, line 739: what/how does it actually test it?

GyanP: I think we don't require it now before that test case is failing here on [first commit](https://github.com/superdesk/superdesk-core/pull/2298#issuecomment-973882915), and later you suggested creating two filters so thought we should add another filter also there. is it ok if I remove that **add_timedelta** filter from here?

petrjasek: I think it would be good to test the filter, I'm just not sure what are we testing here exactly

petrjasek (approved):

PR #2301, opened 2021-11-24
[SDESK-6312] fix: Use Lax for cookie samesite

MarkLark86:
This allows cookies to be sent along with requests initiated by third party services.
This was causing issues with Google OAuth


jerome-poisson (approved):
I've tested this PR, GMail ingest is working fine with it.

PR #2303, opened 2021-11-24
[FEEDING SERVICE][GMAIL]Parse GMail labels

jerome-poisson:
Labels are retrieved using GMail IMAP Extensions. Apparently there is no
standard lib method to parse the RFC 3501 astring, thus a minimal regex
parsing is done.

SDESK-6219

### DIFF
--- a/superdesk/io/feeding_services/gmail.py
+++ b/superdesk/io/feeding_services/gmail.py
@@ -6,7 +6,9 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
+import re
 import imaplib
+from typing import List
 from bson import ObjectId
 from os.path import join
 import time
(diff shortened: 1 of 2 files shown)


superdesk/io/feeding_services/gmail.py
```diff
@@ -99,6 +103,32 @@ def authenticate(self, provider: dict, config: dict) -> imaplib.IMAP4_SSL:
...
+            if match_labels_str is None:
+                raise ValueError(f"Can't find the expected label string in data: {data_str:r}")
+            labels_str = match_labels_str.group(1)
+            labels = [
+                (m.group("quoted") or m.group("unquoted")).replace('\\"', '"') for m in RE_LABEL.finditer(labels_str)
+            ]
+            for parsed_item in parsed_items:
+                subjects = parsed_item.setdefault("subject", [])
+                for label in labels:
+                    subjects.append({"name": label, "qcode": label, "scheme": "label"})
```
petrjasek, line 128: `label` seems kind of generic scheme and potentially conflicting, I would call it `gmail_label`

petrjasek (approved):

superdesk/io/feeding_services/gmail.py
jerome-poisson: I actually did it on purpose, I thought a generic label would be a good thing if similar labels are used with several ingest formats. But it's not a problem to change if you think it's better, I'll update it.

PR #2305, opened 2021-12-17
fix href for media items

petrjasek:
assuming the media can be access via app storage, so it's not doing
migration of media only fixing links.

SDESK-6330

### DIFF
--- a/superdesk/services.py
+++ b/superdesk/services.py
@@ -122,6 +122,9 @@ class BaseService:
             req.projection = json.dumps(projection)
         return self.backend.get_from_mongo(self.datasource, req=req, lookup=lookup)
 
+    def get_all(self):
+        return self.get_from_mongo(None, {}).sort("_id")
+
     def find_and_modify(self, query, update, **kwargs):
         res = self.backend.find_and_modify(self.datasource, query=query, update=update, **kwargs)
         return res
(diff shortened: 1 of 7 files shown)


superdesk/services.py
```diff
@@ -122,6 +122,9 @@ def get_from_mongo(self, req, lookup, projection=None):
             req.projection = json.dumps(projection)
         return self.backend.get_from_mongo(self.datasource, req=req, lookup=lookup)
 
+    def get_all(self):
```
MarkLark86, line 125: If the collection is too big, would this result in an OOM being raised? Would it be better to use generators that yield pages for consumers to iterate over all documents?

superdesk/storage/fix_href.py
```diff
@@ -0,0 +1,95 @@
+import copy
+import superdesk
+
+from flask import current_app as app
+from superdesk.editor_utils import generate_fields, get_field_content_state
+
+
+class FixHrefCommand(superdesk.Command):
+    """Fix href links in db."""
```
MarkLark86, line 9: It would be beneficial if we had a more descriptive docstring here including examples, like we have with other commands, so it is easier to understand from the `readthedocs` documentation.

superdesk/services.py
petrjasek: it's using batches internally in mongo cursor, I tested it on the bi media db and it works fine

MarkLark86 (approved):

PR #2308, opened 2021-12-22
[DATA MANIPULATION] api to restore records 

jerome-poisson:
This is the base implementation of the API to restore records.

SDESK-6311


jerome-poisson:
this PR implement a base API to restore records. It should be enough to start using new storage records from frontend for e2e tests.

petrjasek (approved):

PR #2310, opened 2021-12-28
As a SD admin I want to update the metadata section by uploading (or referencing) a settings file[SDESK-6331]

GyanP:


superdesk/vocabularies/vocabularies.py
```diff
@@ -520,6 +520,36 @@ def get_field_options(self, field) -> Dict:
         cv = self.find_one(req=None, _id=field)
         return cv and cv.get("field_options") or {}
 
+    def update_vocabulary_from_json(self, vocab_items):
```
petrjasek, line 523: would be good to have this on base service rather than here so we can reuse this

superdesk/upload.py
```diff
@@ -55,6 +56,31 @@ def get_upload_as_data_uri(media_id):
     raise SuperdeskApiError.notFoundError("File not found on media storage.")
 
 
+@bp.route("/upload/config-file", methods=["POST", "OPTIONS"])
+@blueprint_auth()
```
petrjasek, line 60: I think  we should check if the current user has manage rights on the resource

superdesk/services.py
```diff
@@ -209,6 +211,36 @@ def remove_from_search(self, item):
...
+                res = self.post([item]) if not cv else self.patch(cv["_id"], item)
+                if res:
+                    success.append(res)
+            except Exception as ex:
+                raise SuperdeskApiError.badRequestError(str(ex), exception=ex)
+
+        if success:
+            return {
+                "_status": "SUCCESS",
+                "_success": {"code": 200, "_message": "Vocabularies uploaded successfully."},
```
petrjasek, line 234: would be good not to use `Vocabularies` in those messages, either use the current datasource or just make that more generic

GyanP: my bad, I totally missed it, I will address this one.

superdesk/upload.py
```diff
@@ -55,6 +58,50 @@ def get_upload_as_data_uri(media_id):
...
+def upload_config_file():
+    if request.method == "OPTIONS":
+        # return headers to avoid CORS problems
+        response = make_response()
+        response.headers.add("Access-Control-Allow-Origin", "*")
+        response.headers.add("Access-Control-Allow-Headers", "*")
+        response.headers.add("Access-Control-Allow-Methods", "POST")
+        return response
+
+    if not current_user_has_privilege("vocabularies"):
```
petrjasek, line 72: imo it would be good to get the resource via request param so we can pick the service and privilege based on that.
but we should have some whitelist of resources which can be updated like this.

GyanP: For `VOCABULARIES MANAGEMENT` the privilege name is `vocabularies`, so here it is the privilege name, not the resource. however, we know that for vocabularies, the resource and privilege names are the same but it does not work for all privileges and resources.
Like we have one privilege called `UNPUBLISH` but do not have resources with the same name.

petrjasek: yep but we can use the resource to figure out its privileges right?

GyanP: Okay got it, we will send resource via request param and then find out its privileges and then based on that allow the user to do further processes so that we can reuse that functionality for other resources also right?

petrjasek: exactly

superdesk/services.py
```diff
@@ -209,6 +211,36 @@ def remove_from_search(self, item):
         """
         return self.backend.remove_from_search(self.datasource, item)
 
+    def update_data_from_json(self, items):
+        success = []
+        for item in items:
+            try:
+                cv = self.find_one(req=None, _id=item["_id"])
```
petrjasek, line 218: we can probably use something generic and not cv here

petrjasek (approved):

PR #2316, opened 2022-01-20
refresh session info every 30s max

petrjasek:
SDESK-6365

### DIFF
--- a/apps/auth/auth.py
+++ b/apps/auth/auth.py
@@ -154,7 +154,7 @@ class SuperdeskTokenAuth(TokenAuth):
             if method in ("POST", "PUT", "PATCH") or method == "GET" and not request.args.get("auto"):
                 now = utcnow()
                 auth_updated = False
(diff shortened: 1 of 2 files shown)


apps/auth/auth.py
```diff
@@ -154,7 +154,7 @@ def check_auth(self, token, allowed_roles, resource, method):
             if method in ("POST", "PUT", "PATCH") or method == "GET" and not request.args.get("auto"):
                 now = utcnow()
                 auth_updated = False
-                if auth_token[app.config["LAST_UPDATED"]] + timedelta(seconds=10) < now:  # update once every 10s max
+                if auth_token[app.config["LAST_UPDATED"]] + timedelta(seconds=30) < now:  # update once every 30s max
```
MarkLark86, line 157: It might be good if this was configurable. Environments with small amount of users can have quicker updates, where as environments with large amount of users can have slower updates

MarkLark86 (approved):

PR #2318, opened 2022-01-24
Accent removal option for search in superdesk [SDESK-6224]

GyanP:


superdesk/default_settings.py
```diff
@@ -212,8 +212,8 @@ def local_to_utc_hour(hour):
             "analyzer": {
                 "phrase_prefix_analyzer": {
                     "type": "custom",
-                    "filter": ["remove_hyphen", "lowercase"],
-                    "tokenizer": "keyword",
+                    "filter": ["remove_hyphen", "lowercase", "asciifolding"],
+                    "tokenizer": "standard",
```
petrjasek, line 216: not sure we want to affect slugline with this, imo we would need to rather add a new config for headline and body

GyanP: Here we are adding `assciifolding` and using the tokenizer `standard` to the existent analyzer **phrase_prefix_analyzer**. so that we can search non-english words.
For example - 

> Suppose our headline is like **još Neka Debata**

if we search "jos Neka Debata" then as a result we get that item.
I updated tokenzier from **keyword** to **standard** becuase in `keyword tokenzier` we have to search whole string like "jos Neka Debata" we will get resultant item but after using `standard tokenzier`, if  we will search only "jos" string then we will get resultant item.

GyanP: Okay, means for slugline don't need this type of searching?

petrjasek: exactly, the analyzer is ok, but I wouldn't apply it to slugline

petrjasek (commented):
looks good, but pls add some behave tests to demonstrate what it can do now

petrjasek (approved):

PR #2320, opened 2022-02-03
As an SD user manager, I want to indicate whether an author wants to be publicly visible or hidden [SDBELGA-605]

GyanP:


superdesk/users/users.py
```diff
@@ -88,6 +88,7 @@ def __init__(self, endpoint_name, app, service, endpoint_schema=None):
             "user_type": {"type": "string", "allowed": ["user", "administrator"], "default": "user"},
             "is_support": {"type": "boolean", "default": False},
             "is_author": {"type": "boolean", "default": True},
+            "public": {"type": "boolean", "default": True},
```
petrjasek, line 91: this wouldn't be set for existing users, might be better to have it `private` and default to `None` or `False`

GyanP: okay I will set its default value `None` and it means here it will show as `private` but in the **author field** it will visible as `public`?

GyanP: means we need a public flag right so in the **author field** we will have a flag called `public` and here in **users DB** will have field/key name is `private`

petrjasek: I'm not sure if we need to call it public in ui, that's kind of the default state, so having there private which you have to toggle might be ok too

GyanP: yep, it's ok to use a toggle name private instead of a public one.

superdesk/users/users.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, endpoint_name, app, service, endpoint_schema=None):
             "user_type": {"type": "string", "allowed": ["user", "administrator"], "default": "user"},
             "is_support": {"type": "boolean", "default": False},
             "is_author": {"type": "boolean", "default": True},
-            "public": {"type": "boolean", "default": True},
+            "private": {"type": "boolean", "default": True, "nullable": True},
```
petrjasek, line 91: default should be `False` or not set

PR #2322, opened 2022-02-10
[SDESK-6369] fix: Unable to correct an item after embargo has passed

MarkLark86:


MarkLark86:
@GyanP Looks like `black` is failing across a multitude of files in v2.3.x, and looks like it's been that way since October last year.

I would rather ignore `black` in v2.3.x than change a lot of files for the sake of linting/code style

apps/archive/archive.py
```diff
@@ -979,7 +980,7 @@ def validate_embargo(self, item):
 
                     if (
                         item[ITEM_STATE] not in {CONTENT_STATE.KILLED, CONTENT_STATE.RECALLED, CONTENT_STATE.SCHEDULED}
-                    ) and embargo <= utcnow():
+                    ) and embargo <= utcnow() and item[ITEM_OPERATION] != ITEM_OPERATION_CORRECT:
                         raise SuperdeskApiError.badRequestError(_("Embargo cannot be earlier than now"))
```
petrjasek, line 984: I wonder why we wouldn't allow the embargo to be expired, it makes sense to check this if it wasn't set before but if we're just looking at the current state that's not invalid per se

MarkLark86: This could happen if the embargo was added, saved, then at a later time the item is published which this validation would pick up (as only a state transition would occur in this scenario).

Though I don't particularly like having strict business rules such as checking embargo time, in my eyes it's more of a warning than a bad request.

petrjasek: yep so the embargo was already validated when saved, but it can expire meanwhile, and that's imo ok. what we enforce is that once the embargo is set user must publish the item before it expires, and I'm just not sure why

MarkLark86: Fair point. At the time of publishing, after embargo expires, it is a valid request. So we can safely ignore that validation (maybe remove the embargo as it's not longer forced under embargo conditions).

I'll update this tomorrow

petrjasek: I would just ignore it, no need to remove it imo

MarkLark86:
@petrjasek I'm closing this one, and migrating the changes to https://github.com/superdesk/superdesk-core/pull/2323

PR #2325, opened 2022-02-17
[SDESK-6384] fix: Pass updates to item_publish signal

MarkLark86:
Also include `item_resend` and `item_resent` signals


MarkLark86:
I've also updated the PR in TGA's repo for this signal to make use of `updates`

petrjasek (approved):

PR #2336, opened 2022-04-12
fix media expiry from mongo

petrjasek:
removing of expired media items didn't work
as we were not using ObjectId but str.

also remove media_references when deleting
an item so those won't prevent published
media from being removed.

SDESK-6432

### DIFF
--- a/apps/archive/common.py
+++ b/apps/archive/common.py
@@ -470,7 +470,7 @@ def fetch_item(doc, desk_id, stage_id, state=None, target=None):
     return dest_doc
 
 
-def remove_media_files(doc):
+def remove_media_files(doc, published=False):
     """Removes the media files of the given doc.
 
     If media files are not references by any other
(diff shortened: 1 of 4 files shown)


apps/archive/common.py
```diff
@@ -502,16 +505,23 @@ def remove_media_files(doc):
...
-                logger.exception("Failed to remove Media Id: {} from item: {}".format(media, doc.get(config.ID_FIELD)))
+                logger.exception("Failed to remove Media Id: %s from item: %s", media, doc.get("guid"))
 
     for attachment in doc.get("attachments", []):
         lookup = {"_id": attachment["attachment"]}
         get_resource_service("attachments").delete_action(lookup)
 
 
+def remove_media_references(item_id, published):
+    get_resource_service("media_references").delete_action({"item_id": item_id, "published": published})
```
MarkLark86, line 521: Would this need an index for `published`, or do you think this should be fast enough?

MarkLark86 (approved):

apps/archive/common.py
petrjasek: there is `item_id` index and then it's only a few items to be checked so should be fine

PR #2337, opened 2022-04-13
FIX: Export action from 3-dots-menu does not work on bel12 [back-end] [SDBELGA-618]

GyanP:
we need to do is to make sure we generate a link with https on the server.


superdesk/download.py
```diff
@@ -40,7 +40,7 @@ def download_file(id, folder=None):
 
 
 def download_url(media_id):
-    return url_for("download_raw.download_file", id=media_id, _external=True)
+    return url_for("download_raw.download_file", id=media_id, _external=True, _scheme="https")
```
petrjasek, line 43: imo that scheme should be in a config so we can change it if we need to (like when running in docker)

petrjasek (changes requested):

superdesk/download.py
MarkLark86: We have a config available already `SESSION_COOKIE_SECURE`, might be good to copy/rename this one to a more generic config:
https://github.com/superdesk/superdesk-core/blob/develop/superdesk/default_settings.py#L155

petrjasek: flask has a config for it [PREFERRED_URL_SCHEME](https://flask.palletsprojects.com/en/2.1.x/config/#PREFERRED_URL_SCHEME) which is used by url_for when there is no request. we can populate it using same logic as for the cookies config and then use it here explicitly

GyanP: means should we add `PREFERRED_URL_SCHEME = https` in https://github.com/superdesk/superdesk-core/blob/develop/superdesk/default_settings.py and then here we can do something like this
`url_for(
"download_raw.download_file", id=media_id, _external=True, _scheme= app.config.get("PREFERRED_URL_SCHEME", "http")
)` right?

petrjasek: yep that's right, but the config should be rather:
```
PREFERRED_URL_SCHEME = env('PREFERRED_URL_SCHEME', 'https')
```
so we can change it via env variables

petrjasek (approved):

PR #2342, opened 2022-05-12
implement output formats api

petrjasek:
listing registered output formats

SDESK-6212


superdesk/publish/formatters/newsml_1_2_formatter.py
```diff
@@ -44,6 +44,9 @@ class NewsML12Formatter(Formatter):
     }
     ENCODING = "UTF-8"
 
+    name = None
```
GyanP, line 47: Any reason for `name = None` here?

petrjasek: not really, I was probably testing something. will set it

GyanP (approved):

PR #2343, opened 2022-05-16
FIX: Existing media contacts are not appearing in the list [SDESK-6412]

GyanP:


petrjasek (changes requested):
pls update the comment to reflect those changes

GyanP:
According to SDESK-6412, 
whenever we create a new event then there it should be possible to add any Media Contact to an Event or Coverage, regardless of the “Media Contact”s state (“active”, “public”) means whether media contact is public or private they should visible and we can add those media contacts in the event.

petrjasek (approved):

PR #2344, opened 2022-05-16
improve elastic apm setup

petrjasek:
- set environment based on CLIENT_URL
- allow local monitoring
- use service name as prefix

CPNHUB-119


petrjasek:
syncing the changes with https://github.com/superdesk/newsroom-core/pull/153 and eventually will use these methods also in newsroom

PR #2345, opened 2022-05-25
Fix NTB Currency Macros in Editor3 [SDNTB-764]

GyanP:


GyanP:
https://github.com/superdesk/superdesk-ntb/pull/434

petrjasek:
I wonder now why does it raise the error? seems like all the cells are there actually, it should be able to render it

GyanP:
> I wonder now why does it raise the error? seems like all the cells are there actually, it should be able to render it

I thought the same but it seems like maybe we are missing something so I think putting an exception would be good.

petrjasek:
seems to me the issue is we're using int numbers for `row_idx` and `col_idx` but there are string indices like `"0", "1"` , not sure where it's coming from, we either need to make sure it's integers or work with strings

GyanP:
> seems to me the issue is we're using int numbers for `row_idx` and `col_idx` but there are string indices like `"0", "1"` , not sure where it's coming from, we either need to make sure it's integers or work with strings

I see, okay, let me do that too.

tests/editor_utils_test.py
```diff
@@ -1923,3 +1923,152 @@ def test_client_compatibility(self):
...
+                        },
+                    },
+                },
+            },
+            "entity": {"type": "TABLE"},
+            "children": " ",
+        }
+
+        DraftJSHTMLExporter.render_table(self, data)
+        self.assertRaises(KeyError)
```
petrjasek, line 2074: imo it shouldn't raise an error, tha payload looks ok

GyanP: Good point

tests/editor_utils_test.py
```diff
@@ -1923,3 +1923,152 @@ def test_client_compatibility(self):
...
+                        },
+                    },
+                },
+            },
+            "entity": {"type": "TABLE"},
+            "children": " ",
+        }
+
+        item = DraftJSHTMLExporter.render_table(self, data)
+        self.assertFalse(item.items(), [])
```
petrjasek, line 2074: why would it be false? I still don't get why this won't render actual table.

GyanP: We can make it `self.assertEqual(item.items(), [])`.

GyanP: >  render actual table.

By actual table, you mean it should return `<table></table>`

tests/editor_utils_test.py
```diff
@@ -2071,6 +2071,8 @@ def test_for_key_error(self):
             "children": " ",
         }
 
-        expected = "<table><tbody><tr><td/><td/><td/></tr><tr><td/><td/><td/></tr><tr><td/><td/><td/></tr></tbody></table>"
+        expected = (
+            "<table><tbody><tr><td/><td/><td/></tr><tr><td/><td/><td/></tr><tr><td/><td/><td/></tr></tbody></table>"
```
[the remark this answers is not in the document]
GyanP: I think the issue was I was taking it as `int` instead of `str`, I have resolved it and now I think everything is good.

petrjasek (approved):

PR #2350, opened 2022-06-06
initial rundown show/template api

petrjasek:
SDESK-6491

apidocs: https://app.swaggerhub.com/apis-docs/petrjasek/superdesk-rest_api/2.5

### DIFF
--- a/apps/rundowns/__init__.py
+++ b/apps/rundowns/__init__.py
@@ -4,21 +4,26 @@ from flask_babel import lazy_gettext
 
 from superdesk.factory.app import SuperdeskEve
 
-from . import shows, templates, privileges
+SCOPE = "rundowns"
+
+from . import shows, templates, privileges, create  # noqa: E402
 
 
 def init_app(app: SuperdeskEve) -> None:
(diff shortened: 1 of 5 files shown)


petrjasek:
I'm thinking about calling it `scope` rather than `context` not to conflict with flask app context

superdesk/factory/app.py
```diff
@@ -123,6 +123,25 @@ def init_indexes(self, ignore_duplicate_keys=False):
                     if not ignore_duplicate_keys:
                         raise
 
+    def item_context(self, name, schema=None):
+        """Register item context."""
+        self.config.setdefault("item_context", {})[name] = {
+            "schema": schema,
+        }
+
+        def update_resource_schema(resource):
```
MarkLark86, line 132: What are your thoughts about having a check here to make sure a new `scope` cannot override existing schema fields, only add to them?

MarkLark86:
@petrjasek Looks good so far. `scope` does sound better, to reduce confusion

MarkLark86 (approved):

PR #2354, opened 2022-06-14
[SDESK-6441] Ability to register new routing rule handlers

MarkLark86:


apps/rules/rule_handlers.py
```diff
@@ -0,0 +1,185 @@
...
+    def apply_rule(self, rule, ingest_item, routing_scheme):
+        raise NotImplementedError()
+
+
+registered_routing_rule_handlers: Dict[str, RoutingRuleHandler] = {}
+
+
+def register_routing_rule_handler(routing_handler: RoutingRuleHandler):
+    if registered_routing_rule_handlers.get(routing_handler.NAME):
+        raise AlreadyExistsError(f"Ingest Publisher: {routing_handler.NAME} already registered")
```
petrjasek, line 37: imo we're loosing ability to override some core behaviour for client if needed, not sure that's needed

apps/rules/routing_rules.py
```diff
@@ -78,6 +78,7 @@ class RoutingRuleSchemeResource(Resource):
                 "type": "dict",
                 "schema": {
                     "name": {"type": "string"},
+                    "handler": {"type": "string", "required": True, "nullable": False, "empty": False},
```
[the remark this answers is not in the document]
MarkLark86: Agree'd, considering I am using the current one as default anyway in the rest of the code

apps/rules/rule_handlers.py
MarkLark86: Understood, I'll remove this tomorrow

petrjasek (approved):

PR #2355, opened 2022-06-15
fix flask request scope error in private filter

petrjasek:
use explicit req param


GyanP:
> use explicit req param

Any reason for Explicit param?

petrjasek:
was getting errors when using flask.request and it will be also simpler to call that api with context manually

GyanP (approved):

PR #2360, opened 2022-06-27
Planning: As a SD admin I like to ingest events from the Onclusive API [SDCP-627]

GyanP:


superdesk/default_settings.py
```diff
@@ -1064,3 +1064,6 @@ def local_to_utc_hour(hour):
 #: .. versionadded:: 2.6
 #:
 RUNDOWNS_TIMEZONE = DEFAULT_TIMEZONE
+
+#: default offset limit which can processed during onclusive API ingest
+ONCLUSIVE_MAX_OFFSET = 10000
```
MarkLark86, line 1069: I don't think we need this in core.

For other Planning related configs, we place a util function here https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/common.py#L152

Also, for documentation purposes we place Planning configs in the githubs README file:
https://github.com/superdesk/superdesk-planning/blob/develop/README.md?plain=1#L128

PR #2363, opened 2022-07-26
ingest and distribute BTA news agency as 2 seperate sources [SDBELGA-660]

devketanpro:
Fixed mapping of content fields to aliases fields.

### DIFF
--- a/tests/io/feeding_services/rss_test.py
+++ b/tests/io/feeding_services/rss_test.py
@@ -658,6 +658,32 @@ class CreateItemMethodTestCase(RssIngestServiceTest):
             items = self.instance._update(provider, None)[0]
         self.assertEqual("tag:www.nrk.no:finnmark:stanset-ikke-for-fotgjenger-1.13362571", items[0]["guid"])
 
+    def test_aliases_fields_are_mapped_to_data(self):
+        """
+        Test the aliases fields are mapped to the content fields.
+        """
+        data = dict(
+            guid="http://news.com/rss/1234abcd",
+            published_parsed=struct_time([2015, 2, 25, 16, 45, 23, 2, 56, 0]),
+            updated_parsed=struct_time([2015, 2, 25, 17, 52, 11, 2, 56, 0]),
+            title="Breaking News!",
+            body_text="This is body_text.",
+            summary="This is summary text.",
+        )
+        field_aliases = [{"summary": "body_html"}]
+        item = self.instance._create_item(data, field_aliases)
+
+        self.assertIsNone(item.get("abstract"))  # because summary is aliased
+        self.assertEqual(item.get("headline"), "Breaking News!")
+        # summary is mapped to body_html
+        self.assertEqual(
+            item.get("body_html"),
+            (
+                '<p><a href="http://news.com/rss/1234abcd" target="_blank">source</a></p>'
+                "This is summary text.This is body_text."
+            ),
+        )
+
 
 class CreateImageItemsMethodTestCase(RssIngestServiceTest):
     """Tests for the _create_image_items() method."""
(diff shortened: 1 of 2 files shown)


petrjasek:
pls describe the issue it's solving and ideally provide a test case @devketanpro

tests/io/feeding_services/rss_test.py
```diff
@@ -658,6 +658,32 @@ def test_guid_not_set(self):
...
+        item = self.instance._create_item(data, field_aliases)
+
+        self.assertIsNone(item.get("abstract"))  # because summary is aliased
+        self.assertEqual(item.get("headline"), "Breaking News!")
+        # summary is mapped to body_html
+        self.assertEqual(
+            item.get("body_html"),
+            (
+                '<p><a href="http://news.com/rss/1234abcd" target="_blank">source</a></p>'
+                "This is summary text.This is body_text."
```
petrjasek, line 683: not sure this is actually something we would want. can you pls elaborate how does it fix the issue? we get the content split in summary and body text from bta?

devketanpro: As per my conversation with **Gideon**, He told me when we used **Rss/Atom** feedparser to ingest items coming from **BTA**,
so here the `body` data is parsed in the `abstract` field in **superdesk**. For **superdesk** it is ok, as we know **Belga** is not using `abstract`, so here before applying any specific changes for **Belga** , we have a feature in the **Rss/Atom** feedparser called `Alias`, so when we using a `Alias` feature here we set `summary` as a **Content field** and `body_html` as a **Alias field**, so in this case `summary` data would we mapped to the `body_html` and `abstract` goes to `None`.
but previously when we using this feature so `abstract` fields is set to `None` but in the `body_html` we don't show any data because we don't have any mapping of aliases fields. so I have fixed it. 
Please let me know if i am missing something.

petrjasek: so what would we get in the body without your changes?

devketanpro: you mean body_html ?

petrjasek: yep in the body_html

devketanpro: https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feeding_services/rss.py#L323

petrjasek: so if you set that sommary -> body_html alias it doesn't do it right?
so imo there is an issue with the code in https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feeding_services/rss.py#L274-L289
looking at the code it might be that on https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feeding_services/rss.py#L274 it should be rather using `.keys()` because there it's looking for the `name_in_data` which is the rss field name. but that's just me guessing.
in general I would rather try to fix the aliasing code than doing another round later overriding what was set before

devketanpro: > so if you set that sommary -> body_html alias it doesn't do it right? so imo there is an issue with the code in https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feeding_services/rss.py#L274-L289 looking at the code it might be that on https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feeding_services/rss.py#L274 it should be rather using `.keys()` because there it's looking for the `name_in_data` which is the rss field name. but that's just me guessing. in general I would rather try to fix the aliasing code than doing another round later overriding what was set before

Ok I am looking into that code https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feeding_services/rss.py#L274-L289, If there are any updates, I'll let you know.
Thanks

PR #2374, opened 2022-09-01
Fix : item updates if it's a new version.

devketanpro:


superdesk/io/commands/update_ingest.py
```diff
@@ -665,10 +665,13 @@ def ingest_item(item, provider, feeding_service, rule_set=None, routing_scheme=N
...
             new_version = is_new_version(item, old_item)
             updates = deepcopy(item)
-            ingest_service.patch_in_mongo(old_item[superdesk.config.ID_FIELD], updates, old_item)
+            if new_version:
+                ingest_service.patch_in_mongo(old_item[superdesk.config.ID_FIELD], updates, old_item)
+                item.update(updates)
+
             item.update(old_item)
-            item.update(updates)
             items_ids.append(item["_id"])
```
petrjasek, line 673: imo this should also happen only if it's a new version because that will be used later for syncing with elastic, and if there were no changes there is no need to do it

petrjasek (approved):

PR #2376, opened 2022-09-09
handle item missing in elastic when updating

petrjasek:
as long as it's in mongo instead of failing insert it there

SDESK-6590

### DIFF
--- a/superdesk/storage/fix_links.py
+++ b/superdesk/storage/fix_links.py
@@ -1,9 +1,13 @@
 import copy
+import logging
 import superdesk
 
 from flask import current_app as app
 from superdesk.editor_utils import generate_fields, get_field_content_state
 from superdesk.errors import SuperdeskApiError
+
+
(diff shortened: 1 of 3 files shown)


superdesk/storage/fix_links.py
```diff
@@ -56,8 +57,11 @@ def run(self, prefix, resource, dry_run):
                     if dry_run:
                         print("update", item["_id"], updates.keys())
                     else:
-                        print(".", end="")
-                        service.system_update(item["_id"], updates, orig)
+                        try:
+                            service.system_update(item["_id"], updates, orig)
+                            print(".", end="")
+                        except SuperdeskApiError:
+                            print("x", end="")
```
MarkLark86, line 64: It might be good for debugging purposes if the error were to be logged to the console for further investigation (with resource name/id and reason for error included in the output)

MarkLark86 (approved):

PR #2381, opened 2022-09-16
fix(gmail) token expiry calculation is wrong

marwoodandrew:
I'm surprised no one else seems to have this problem!

We get a timestamp time from google and save it in expires_at as time zone naïve in the oauth2_token collection, using datetime.fromtimestamp with tz parameter so it returns local time naïve.

It is assumed by mongo to be UTC.

When we retrieve it in timezone UTC+10 the expires_at is effectively well in the future so don't refresh the token when required. 

So now we just treat it as timezone naïve when testing if near expiry.


petrjasek:
might be because our servers are close to utc timezone (only +1 or +2)

petrjasek (approved):

PR #2392, opened 2022-10-27
Enable search highlights on /ingest and /published apis [SDESK-6611]

devketanpro:


petrjasek:
it would be better probably to create some `HighlightsSearchMixin` and use that in both published/ingest/archive instead of inheriting published in ingest, that could have some side effects

petrjasek (commented):
looks good, pls add a behave test case for those

petrjasek (approved):

PR #2394, opened 2022-11-08
[NHUB-195] improve(ws): Ability to filter consumers using message filters

MarkLark86:


superdesk/websockets_comms.py
```diff
@@ -192,6 +204,52 @@ def _client_loop(self, websocket):
...
+            return clients
+
+        def _filter(websocket: WebSocketClientProtocol) -> bool:
+            if filters["include"] and not websocket.path.startswith(self.subscribe_prefix):
+                # If ``filter.include`` is defined, client must provide url args in websocket path
+                # as we're explicitly including only clients that have args in this list
+                return False
+
+            try:
+                url_args: Dict[str, str] = {
```
petrjasek, line 235: I suppose it's not a big deal, but this could be done once when client connects and not every time there is a new message

petrjasek (approved):

superdesk/websockets_comms.py
MarkLark86: Good optimisation, I'll do that tomorrow (as to keep ws service running as fast as possible)

petrjasek (approved):

PR #2395, opened 2022-11-08
PNG image upload fails (EWTN) [SDESK-6662]

devketanpro:


petrjasek (changes requested):
we don't want to backport all those changes to the previous release, only the pillow related ones

PR #2397, opened 2022-11-10
fix(index_from_mongo) index collections with string id's

marwoodandrew:


MarkLark86:
If you run the following, it will fix the lint issues automatically for you:
```bash
pip install black
black .
```

marwoodandrew:
yeah! I've got it installed, just haven't gotten used to running it!

petrjasek (approved):

PR #2398, opened 2022-11-11
Fix : admin user can't clear their own session but can clear other user sessions [SDESK-6696].

devketanpro:


apps/auth/service.py
```diff
@@ -146,13 +146,9 @@ def __can_clear_sessions(self, user):
...
             1. Check if the user exists.
-            2. Check if the user is clearing his/her own sessions.
 
         :return: error message if invalid.
         """
 
         if not user:
             return "Invalid user to clear sessions."
-
-        if str(user[config.ID_FIELD]) == str(flask.g.user[config.ID_FIELD]):
```
petrjasek, line 157: just not sure if we should drop the current session, how is that handled?

devketanpro: Actually the problem was that, logged user can not clear their own season, so when I investigated I found we have a prebuilt functionality logged user can not clear their own session. so here I remove that condition.

petrjasek: that part is fine, I'm just not sure if it should drop all user session but the current one so the user won't end up logged out of the system

devketanpro: So you mean we have to clear all session of active user except current session, so user would not logged out of the system.

petrjasek: exactly

devketanpro: Actually, I looked in to the code and I couldn't distinguish between the current and other sessions of user? can you please provide me some suggestions on it. :)

apps/auth/service.py
```diff
@@ -130,9 +130,11 @@ def delete(self, lookup):
         if error_message:
             raise SuperdeskApiError.forbiddenError(message=error_message)
 
-        # Delete all the sessions
+        # Delete all the sessions except current session
         for session in sessions:
-            get_resource_service("auth").delete_action({config.ID_FIELD: str(session[config.ID_FIELD])})
+            current_session_id = auth.get_auth().get("_id")
```
petrjasek, line 135: not that it would be a big deal, but this should be done before the loop, it won't change inside so makes me wonder why it's there

devketanpro: sorry my bad, I got your point.

petrjasek (approved):

PR #2411, opened 2022-12-16
Fix : Picture correction not being saved [SDESK-6735]

devketanpro:


apps/publish/content/common.py
```diff
@@ -723,7 +723,7 @@ def _set_updates_for_media_items(self, doc, updates):
             return
 
         for key in DEFAULT_SCHEMA.keys():
-            if doc.get(key) and not updates.get(key):
+            if doc.get(key) and doc.get(key) == updates.get(key):
                 updates[key] = doc[key]
```
petrjasek, line 727: this seems to me like it would never change the `updates` if it only sets the values which are already equal

petrjasek:
pls add/update a test case, I'm not sure I understand the fix

devketanpro:
> pls add/update a test case, I'm not sure I understand the fix

ok I will

apps/publish/content/common.py
```diff
@@ -723,7 +723,7 @@ def _set_updates_for_media_items(self, doc, updates):
             return
 
         for key in DEFAULT_SCHEMA.keys():
-            if doc.get(key):
+            if doc.get(key) and updates.get("operation") == "publish":
```
petrjasek, line 726: might be more readable if that condition for publish operation is part of that previous one as it's not really key related so we won't need it in the loop:
```python
if doc.get("type") not in MEDIA_TYPES or updates.get("operation") == "publish":
    return
```

petrjasek (approved):

PR #2415, opened 2023-01-03
As a Belga AD admin I want to restrict the sending of items to a Desk [SDBELGA-702]

devketanpro:
https://github.com/superdesk/superdesk-client-core/pull/4171

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -49,7 +49,7 @@ desks_schema = {
     "working_stage": Resource.rel("stages", True),
     "content_expiry": {"type": "integer"},
     "source": {"type": "string"},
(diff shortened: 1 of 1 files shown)


apps/desks.py
```diff
@@ -49,6 +49,7 @@ class DeskTypes(SuperdeskBaseEnum):
     "working_stage": Resource.rel("stages", True),
     "content_expiry": {"type": "integer"},
     "source": {"type": "string"},
+    "restriction": {"type": "boolean"},
```
petrjasek, line 52: can you pls name it in a way that it's obvious what does it restrict?

PR #2418, opened 2023-01-10
Add image suggestion feature to the iMatrics service

imatricsOpenSource:
Hi! These changes adds support for an image suggestion widget in the fronted that suggests images based on the tags on the article.

Two new environment variables are required. We can discuss them in our shared slack channel.

Happy reading!

### DIFF
--- a/superdesk/text_checkers/ai/__init__.py
+++ b/superdesk/text_checkers/ai/__init__.py
@@ -88,7 +88,6 @@ class AIService(BaseService):
         docs[0].update({"analysis": analyzed_data})
         return [0]
 
-
 class AIDataOpResource(Resource):
     schema = {
         "service": {
@@ -157,7 +156,6 @@ class AIDataOpService(BaseService):
         result = service.data_operation("POST", operation, name, data)
         docs[0].update({"result": result})
         return [0]
-
 
 
 class AIImageResource(Resource):
--- a/superdesk/text_checkers/ai/__init__.py
+++ b/superdesk/text_checkers/ai/__init__.py
@@ -164,13 +164,13 @@ class AIImageResource(Resource):
             "type": "string",
             "required": True,
         },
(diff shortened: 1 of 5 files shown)


superdesk/text_checkers/ai/__init__.py
```diff
@@ -156,6 +158,70 @@ def create(self, docs, **kwargs):
         return [0]
 
 
+class AIImageResource(Resource):
+    schema = {
+        "service": {
+            "type": "string",
+            "required": True,
+        },
+        "item": {
```
petrjasek, line 167: would just call this `items` if it's a list and not a single one

petrjasek: or make it `item` which won't be a list but a dictionary like in the other services

hanstchou: Thank you for the feedback. We have now adjusted this and refactored from "item" to "items".

superdesk/text_checkers/ai/imatrics.py
```diff
@@ -208,6 +216,27 @@ def _analyze(self, data, **params):
...
 
+    def search_images(self, items: list) -> dict:
+        """fetch image suggestions"""
+        if not self.base_url or not self.user or not self.key:
+            logger.warning("IMatrics is not configured properly, can't fetch images")
+            return {"result": []}
+        data = items
+        try:
+            r_data = self._search_images(data)
+        except:
```
petrjasek, line 227: would be useful to log the exception

petrjasek (approved):

superdesk/text_checkers/ai/imatrics.py
hanstchou: We were told to add filtering of images without src in the server instead, therefore we made a small change after the approval.

PR #2419, opened 2023-01-11
Rename send_to_desks_not_allowed

thecalcc:


petrjasek:
what was the motivation for this?
it will be undefined by default but that means it's actually allowed, so explicitly blocking it makes more sense imo

superdesk/data_updates/00030_20230113-141048_desks.py
```diff
@@ -0,0 +1,25 @@
...
+    resource = 'desks'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        print(
+            mongodb_collection.update_many({}, {"$set": {"send_to_desk_allowed": True}})
+        )
+
+    def backwards(self, mongodb_collection, mongodb_database):
+        print(
+            mongodb_collection.update_many({}, {"$unset": {"send_to_desk_allowed": True}})
```
tomaskikutis, line 24: This should remove `send_to_desk_allowed` here. `backwards` is executed when there is an issue with a deployment and a rollback is done to a previous version

tomaskikutis:
> now all desk related tests will need to be fixed when this is a required field.

Would it work if we also updated the prepopulate script? I saw there are around 5 usages. Or is it also used in back-end tests with different data than e2e tests use?

PR #2420, opened 2023-01-16
[TGA-39] Improve author handling in ContentAPI

MarkLark86:


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -580,6 +580,9 @@ def _format_authors(self, article):
                 "biography": user.get("biography", author.get("biography", "")),
             }
 
+            if user.get("_id"):
+                author["uri"] = "domain:user:" + str(user["_id"])
```
petrjasek, line 584: I suppose domain here should be based on `SERVER_DOMAIN` config

MarkLark86: I had thought of that. But `SERVER_DOMAIN` is a config and could change (although shouldn't), and this is used by the 360info CAPI to search for authors. So I thought here `domain` could simply refer to `self` (aka the provider).
Otherwise I would have to change the 360info code to use `SERVER_DOMAIN` in the search query as well.

Because of this, do you think this should be okay to leave as is, or do you want me to change it?

petrjasek: I see, in core imo it should use a config, we might introduce a new one which might not necessarily be the same as `SERVER_DOMAIN` so this wouldn't have to change. something like `URN_DOMAIN`

petrjasek: and maybe we should also prefix this with `urn:`

MarkLark86: Sounds good. I'll add the config tomorrow and update the PRs

petrjasek (approved):

PR #2421, opened 2023-01-24
Improve data structure in MediaEditorService

thecalcc:


superdesk/media/media_editor.py
```diff
@@ -50,37 +50,40 @@ class MediaEditorResource(Resource):
...
+    def transform(self, im, operation):
         """Apply image transformation
 
         :param Image im: image to transform
         :param str operation: name of the operation to do
         :param param: parameters of the operation
         :return Image: resulting image
         """
-        if operation == "rotate":
+        operationType = operation[0]
```
petrjasek, line 61: in python we rather use `operation_type`

petrjasek (commented):
looks good, can you fix the failing tests or would you need some help there?

petrjasek (approved):

PR #2427, opened 2023-02-06
Add External user type in the backend [SDESK:6824]

devketanpro:


superdesk/users/services.py
```diff
@@ -229,8 +229,15 @@ def get_avatar_renditions(self, doc):
         renditions = get_resource_service("upload").find_one(req=None, _id=doc)
         return renditions.get("renditions") if renditions is not None else None
 
+    def is_email_required(self, doc):
```
petrjasek, line 232: this would be better handled with custom validation, like we have for `username_pattern`, so it's also checked for updates

petrjasek (commented):
we should also prevent those users from logging in, they might have username/password populated but with that switch active they shouldn't be able to authenticate

MarkLark86:
> we should also prevent those users from logging in, they might have username/password populated but with that switch active they shouldn't be able to authenticate

@petrjasek In this scenario it might be good to remove the `password` on update, if `user_type` is changed to `external`.

Likewise if the `user_type` is changed **from** `external` to something else, should we send a `reset_password` email?

devketanpro:
Hey @MarkLark86 @petrjasek , we can also check if user is external on login, if it is external so in this case we raise an `ExternalUser` error. could you please provide your opinion on it.

apps/auth/db/db.py
```diff
@@ -24,6 +24,12 @@ def authenticate(self, credentials, ignore_expire=False):
         if not user:
             raise CredentialsAuthError(credentials)
 
+        _user = get_resource_service("users").find_one(req=None, username=credentials.get("username"))
+        if _user.get("user_type") == "external":
+            raise ExternalUserError(
+                message="Oops!This account has been changed to External. External accounts have no login capability."
```
petrjasek, line 30: I guess you only need 1 place to specify the message, and would be good to use gettext for it

apps/auth/db/db.py
```diff
@@ -12,10 +12,11 @@
 from flask import g
 from apps.auth.service import AuthService
 from superdesk import get_resource_service
-from apps.auth.errors import CredentialsAuthError, PasswordExpiredError
+from apps.auth.errors import CredentialsAuthError, PasswordExpiredError, ExternalUserError
 from flask import current_app as app
 from superdesk.utc import utcnow
 import datetime
+from superdesk.text_utils import get_text
```
MarkLark86, line 19: This is the incorrect version of `get_text` (this utility returns the plain text version of the provided HTML/XML value).

We need gettext to flag this error message for translation. On the server side this is done using the `flask_babel` module.

i.e.
```python
from flask_babel import _

def test():
    return _("Flags me for translation")
```

MarkLark86 (approved):

PR #2428, opened 2023-02-07
[SDAAP-45] The AP Media v2.2 includes the nitf namespace

marwoodandrew:
Rather than dealing with a mapping that include the namespace definitions, I took the simpler route of just removing the namespace.


petrjasek (approved):
just wondering - would this work also with v2.1?

marwoodandrew:
@petrjasek Yes it's fine with v2.1, there is no namespace defined in the nitf so just skips the attempt to remove it!

PR #2432, opened 2023-02-23
Add multi-line quote support

thecalcc:


superdesk/editor_utils.py
```diff
@@ -435,6 +437,9 @@ def render_embed(self, props):
         return div
 
     def render_table(self, props):
+        print("---------------------------------")
+        print(props)
```
tomaskikutis, line 441: this is probably test code that we shouldn't commit? :upside_down_face:

PR #2436, opened 2023-03-13
cache filters/products/conditions when publishing

petrjasek:
to improve enqueue speed.

using cache with tags for invalidation, but also keeping the loaded data on `flask.g` to avoid multiple loading from the cache during single task execution.

SDESK-6840

### DIFF
--- a/apps/products/product_test.py
+++ b/apps/products/product_test.py
@@ -9,6 +9,8 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 import logging
+import warnings
+
 from superdesk import get_resource_service
 from superdesk.resource import Resource
 from superdesk.services import BaseService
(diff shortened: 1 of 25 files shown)


apps/products/product_test.py
```diff
@@ -61,10 +60,9 @@ def create(self, docs, **kwargs):
         doc["_items"] = results
         return [article_id]
 
-    def test_products(self, article, lookup=None):
-        req = ParsedRequest()
+    def test_products(self, article):
```
MarkLark86, line 63: Planning module will also need to be updated (to remove `lookup=None`): https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/output_formatters/utils.py#L42

MarkLark86 (approved):

apps/products/product_test.py
petrjasek: I see, will keep that param for then to make it easier to backport it

PR #2447, opened 2023-04-26
highlights endpoint fix

thecalcc:


tomaskikutis:
@thecalcc didn't we talk that we need to revert this?

tomaskikutis:
or that was some other PR?

PR #2448, opened 2023-04-28
Fix : pictures of de- and rescheduled stories do not get published [SDCP-686]

devketanpro:


apps/archive/archive.py
```diff
@@ -870,8 +870,6 @@ def deschedule_item(self, updates, original):
         :param original: original is document.
         """
         updates[ITEM_STATE] = CONTENT_STATE.PROGRESS
-        updates[PUBLISH_SCHEDULE] = original[PUBLISH_SCHEDULE]
-        updates[SCHEDULE_SETTINGS] = original[SCHEDULE_SETTINGS]
```
petrjasek, line 874: I think we want to unset it, but only if that item was auto published

devketanpro: I see, but how can be identify item auto published.

petrjasek: I think we set `auto_publish=True` on those

devketanpro: okay let me check.

petrjasek: I think you have to change it not only on the picture item itself but also in the associations of the text item

PR #2451, opened 2023-05-05
Fix:pictures of de- and rescheduled stories do not get published [SDCP-686]

devketanpro:


petrjasek (commented):
can you add some test for this?

petrjasek (approved):

PR #2459, opened 2023-07-03
SDAAP-75 Add function to remove embedded media content

marwoodandrew:


superdesk/publish/formatters/__init__.py
```diff
@@ -135,6 +136,38 @@ def set_destination(self, destination=None, subscriber=None):
...
+
+        # List of keys of the removed entities
+        keys = []
+
+        def not_embed(block):
+            if block.type.lower() == "atomic":
+                keys.extend([e.key for e in block.entities])
+                for idx, kk in enumerate(block.entities):
+                    del block.entities[idx]
+                del block.entities
```
MarkLark86, line 155: When a `Block` instance is constructed, the `entites` attribute is populated (https://github.com/superdesk/superdesk-core/blob/1050960444d26b9d845f0ecde78b910782c6e01e/superdesk/editor_utils.py#L186C25-L186C39).

As to ensure the `entities` attribute is always defined, would it be good to add a `clear` function in the `EntitySequence` class and use it here instead?

superdesk/publish/formatters/ninjs_ftp_formatter.py
```diff
@@ -47,48 +42,212 @@ def _transform_to_ninjs(self, article, subscriber, recursive=True):
...
+        if article.get("type") == "text" and recursive:
+            self.apply_product_filtering_to_associations(formatted_article, subscriber)
+
+        ninjs = super()._transform_to_ninjs(formatted_article, subscriber, recursive)
+
+        renditions = ninjs.get("renditions")
+        if renditions:
+            for name, rendition in renditions.items():
+                rendition["href"] = (
+                    self.path.lstrip("/")
```
MarkLark86, line 62: Why remove the `/` prefix? It was there in the original implementation

superdesk/publish/formatters/ninjs_ftp_formatter.py
```diff
@@ -47,48 +42,212 @@ def _transform_to_ninjs(self, article, subscriber, recursive=True):
...
+        return ninjs
+
+    def apply_product_filtering_to_associations(self, article, subscriber):
+        """
+        Remove the embedded items from the article that the subscriber has no matching product for.
+        :param article:
+        :param subscriber:
+        :return:
+        """
+        kill_keys = []
```
MarkLark86, line 76: `kill_` might be a confusing naming here, as we have the `Kill` action/state and may confuse developers.

superdesk/publish/formatters/ninjs_ftp_formatter.py
```diff
@@ -47,48 +42,212 @@ def _transform_to_ninjs(self, article, subscriber, recursive=True):
...
+        for key, item in article.get("associations", {}).items():
+            if key.startswith("editor_"):
+                result = get_resource_service("product_tests").test_products(item, lookup=None)
+                matching_products = set(p["product_id"] for p in result if p.get("matched", False))
+                if not matching_products.intersection(permitted_products):
+                    kill_keys.append(key)
+
+        self.remove_embeds(article, kill_keys)
+
+    def remove_embeds(self, article, kill_keys):
```
MarkLark86, line 88: A lot of similarities here between this and the parent `remove_all_embeds` function.

Would it be possible if the `remove_all_embeds` (once moved to `editor_utils`) was a more generic version that allows to provide a custom filter function/side-affects instead?

superdesk/publish/formatters/ninjs_ftp_formatter.py
```diff
@@ -47,48 +42,212 @@ def _transform_to_ninjs(self, article, subscriber, recursive=True):
...
+            for name, rendition in renditions.items():
+                rendition["href"] = (
+                    self.path.lstrip("/")
+                    + ("/" if not self.path.endswith("/") and self.path else "")
+                    + get_rendition_file_name(rendition)
+                )
+
+        return ninjs
+
+    def apply_product_filtering_to_associations(self, article, subscriber):
```
MarkLark86, line 69: Until we have a way to configure this in the front-end, it would be good to have a system wide config to control this (and default to `False - do not filter associations`, so this becomes an opt-in feature as not to affect existing customers)

MarkLark86 (changes requested):

superdesk/publish/formatters/ninjs_ftp_formatter.py
marwoodandrew: The idea is to make the paths relative rather than absolute on the FTP server

superdesk/publish/formatters/__init__.py
```diff
@@ -135,6 +136,38 @@ def set_destination(self, destination=None, subscriber=None):
         self.destination = destination
         self.subscriber = subscriber
 
+    def remove_all_embeds(self, article):
```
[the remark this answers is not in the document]
marwoodandrew: Done!

superdesk/publish/formatters/__init__.py
marwoodandrew: Done

superdesk/publish/formatters/ninjs_ftp_formatter.py
marwoodandrew: Done, renamed to remove_keys

superdesk/publish/formatters/ninjs_ftp_formatter.py
marwoodandrew: Done!

superdesk/publish/formatters/ninjs_ftp_formatter.py
marwoodandrew: Not that keen! the remove_all_embeds is a sweet simple thing. What's going on the the formatter is more complex, I'd prefer not to mix them.

superdesk/publish/formatters/__init__.py
```diff
@@ -15,6 +15,7 @@
 from superdesk.metadata.item import ITEM_TYPE, CONTENT_TYPE, FORMATS, FORMAT
 from superdesk.etree import parse_html
 from superdesk.text_utils import get_text
+from superdesk.editor_utils import get_content_state_fields, filter_blocks
```
MarkLark86, line 18: This import is no longer used

MarkLark86 (approved):

superdesk/publish/formatters/ninjs_ftp_formatter.py
MarkLark86: @marwoodandrew An issue I have with the change to the FTP path, will this not require a change to the config after updating?

I understand your reasons, but we will have to make sure this is communicated as a breaking change when we update clients that use this formatter/transmitter, so that their configs are updated.

superdesk/publish/formatters/__init__.py
marwoodandrew: Sorted!

MarkLark86:
@petrjasek Would be good to get your review as well please

superdesk/publish/formatters/ninjs_ftp_formatter.py
petrjasek: I don't think anyone else is using ninjs ftp formatter, so as long as we add some notice we should be fine

PR #2465, opened 2023-08-10
SDAAP-83 Update to a story published with Crops not matching the image

marwoodandrew:
Addresses both SDAAP-83 and SDAAP-91, if the association had changed it was still using the metadata for the old one.


MarkLark86:
@marwoodandrew It would be good to have tests around this, so we don't get any regressions here

MarkLark86 (approved):

PR #2470, opened 2023-09-15
Fix : NTB articles scheduled to publishing are not included in the NTB to NPK internal routing [SDNTB-837]

devketanpro:


superdesk/internal_destinations.py
```diff
@@ -63,9 +63,12 @@ def handle_item_published(sender, item, desk=None, **extra):
...
                 continue
-
-        if dest.get("send_after_schedule", False) and item.get("state") != "published":
-            # if send_after_schedule is set to True and item state is other than published
+        if not extra.get("after_scheduled") and dest.get("send_after_schedule"):
+            # if after_schedule is set to False and send_after_schedule is True
+            # then don't execute
+            continue
+        if extra.get("after_scheduled") and not dest.get("send_after_schedule"):
+            # if after_schedule is set to True and send_after_schedule is False
```
petrjasek, line 71: I think you have to also handle the case when item is not scheduled and it's called with `after_scheduled=False` , otherwise it would create 2 items imo

devketanpro: do you mean that the article is published? Not scheduled; if so, do we want to avoid creating a duplicate item through internal destination ?
and IMO it is called with `after_scheduled=False` only when item is published.

petrjasek: yep if it's just published and not scheduled, how many items do you get at the end?

devketanpro: one item on the desk where it will be send through internal_destination and one item on the desk where it will be published.

petrjasek: I see, ok then. can you pls add some test so we have this fix covered

devketanpro: Okay I will

features/internal_destinations.feature
```diff
@@ -705,6 +705,104 @@ Feature: Internal Destinations
         """
         Then we get OK response
         When we get "/published?where=%7B%22processed_from%22%3A%22123%22%7D"
+        Then we get list with 1 items
+        When we get "/published"
+        Then we get list with 2 items
```
petrjasek, line 710: why are there 2 published items created?

features/internal_destinations.feature
```diff
@@ -705,6 +705,104 @@ Feature: Internal Destinations
...
+                "time_zone": "Europe/London",
+                "utc_publish_schedule": "2099-05-19T10:15:00+0000"
+            }
+        }
+        """
+        Then we get OK response
+        When we get "/archive"
+        Then we get list with 0 items
+        When we get "/published"
+        Then we get list with 2 items
```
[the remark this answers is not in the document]
devketanpro: yep it is auto publish

features/internal_destinations.feature
```diff
@@ -705,6 +705,104 @@ Feature: Internal Destinations
...
+
+    @auth
+    Scenario: Send item to the destination desk on scheduled if send_after_schedule is enabled.
```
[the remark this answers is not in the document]
devketanpro: we need to develop logic based on the initial span, and if we find the first span, then display the text together with the previous and next siblings, and if there are any more highlighted span in this paragraph, then they will also display.

features/internal_destinations.feature
devketanpro: because of `Internal_Destination_Auto_Publish` macro

features/internal_destinations.feature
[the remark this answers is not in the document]
devketanpro: my bad

features/internal_destinations.feature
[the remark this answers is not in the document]
petrjasek: I see

petrjasek (approved):

PR #2474, opened 2023-09-21
fix error when parsing exif on media upload

petrjasek:
SDCP-712


devketanpro (commented):
I'm not sure how we resolved this issue. Would it be a good idea to include error tests?

petrjasek:
without that change the new test case fails

devketanpro (approved):

PR #2477, opened 2023-09-27
Fix: Update logic for publish internal destination [SDNTB-837]

devketanpro:
Updated logic with the addition of `item[PUBLISH_SCHEDULE]`  so that when an item is published, it is not dependent on the config `send_after_schedule`.
`item[PUBLISH_SCHEDULE]` is `None` when item is published.

### DIFF
--- a/superdesk/internal_destinations.py
+++ b/superdesk/internal_destinations.py
@@ -63,8 +63,8 @@ def handle_item_published(sender, item, desk=None, **extra):
                 continue
             if not filters_service.does_match(content_filter, item):
                 continue
-        if not extra.get("after_scheduled") and dest.get("send_after_schedule"):
-            # if after_schedule is set to False and send_after_schedule is True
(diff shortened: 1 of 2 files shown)


superdesk/internal_destinations.py
```diff
@@ -63,8 +63,8 @@ def handle_item_published(sender, item, desk=None, **extra):
                 continue
             if not filters_service.does_match(content_filter, item):
                 continue
-        if not extra.get("after_scheduled") and dest.get("send_after_schedule"):
-            # if after_schedule is set to False and send_after_schedule is True
+        if not extra.get("after_scheduled") and item[PUBLISH_SCHEDULE] is not None and dest.get("send_after_schedule"):
+            # if after_schedule is set to False and item[PUBLISH_SCHEDULE] is not None and send_after_schedule is True
```
petrjasek, line 67: would be good to describe it in a more user friendly way, it seems pretty much like the code above

petrjasek (approved):

superdesk/internal_destinations.py
devketanpro: yep

PR #2478, opened 2023-10-12
[SDESK-7062] Allow registering new resources for autocomplete suggestions

MarkLark86:


features/archive_autocomplete.feature
```diff
@@ -34,3 +34,14 @@ Feature: Archive autocomplete
             {"value": "PUBLISHED-C"}
         ]}
         """
+        When we get "/archive_autocomplete?field=slugline&language=en&resources=archive"
```
petrjasek, line 37: I'm not sure if we should add that resources param, afaik there should be always suggestions from all configured resources

features/archive_autocomplete.feature
```diff
@@ -34,3 +34,14 @@ Feature: Archive autocomplete
...
+        Then we get list with 3 items
+        """
+        {"_items": [
+            {"value": "PUBLISHED-A"},
+            {"value": "PUBLISHED-B"},
+            {"value": "PUBLISHED-C"}
```
petrjasek, line 43: might be useful to add the document count to the payload so we could sort it also based on that

features/archive_autocomplete.feature
MarkLark86: I was just forward thinking there. I will remove it if you don't foresee any use cases

petrjasek: It might be useful to limit it, but imo would be also useful to get them all together by default

MarkLark86: It does that now. If no argument is provided, it uses all registered resources

apps/archive/autocomplete.py
```diff
@@ -16,62 +16,89 @@
...
-        query = {
-            "query": {
-                "bool": {
-                    "filter": [
-                        {"term": {"state": "published"}},
-                        {"term": {"language": language}},
-                        {"range": {"versioncreated": {"gte": versioncreated_min}}},
-                    ],
-                },
+        return ListCursor([{"value": key, "count": all_suggestions[key]} for key in sorted(all_suggestions.keys())])
```
MarkLark86, line 58: @petrjasek I kept the sorting based on bucket key to keep with same functionality.

If we need to, we can modify the sort from the front-end (such as hoisting those that have `count > 1` to the top, and leave the rest sorted as is).

petrjasek (approved):

PR #2483, opened 2023-10-31
Fix : hide http push security token in the subscribers api [SDESK-7093]

devketanpro:


superdesk/publish/subscribers.py
```diff
@@ -107,8 +107,9 @@ def get(self, req, lookup):
             req = ParsedRequest()
         if req.args and req.args.get("filter_condition"):
             filter_condition = json.loads(req.args.get("filter_condition"))
-            return ListCursor(self._get_subscribers_by_filter_condition(filter_condition))
-        return super().get_from_mongo(req=req, lookup=lookup)
+            return self.hideConfigField(self._get_subscribers_by_filter_condition(filter_condition), "secret_token")
+
+        return self.hideConfigField(list(super().get_from_mongo(req=req, lookup=lookup)), "secret_token")
```
petrjasek, line 112: would be good to handle more fields, like `apikey` or `password`

devketanpro: okay

superdesk/publish/subscribers.py
```diff
@@ -103,12 +103,14 @@ class SubscribersService(CacheableService):
     cache_lookup = {"is_active": True}
 
     def get(self, req, lookup):
+        fields = ("secret_token", "password", "apiKey", "access_key_id", "secret_access_key")
```
petrjasek, line 106: just wondering why it's set here and not in the function itself, or as a property on the class

petrjasek (approved):

PR #2489, opened 2023-11-10
pls adapt the NINJS outputformatter to provide better references for all renditions [SDCP-723]

devketanpro:


superdesk/publish/formatters/ninjs_formatter.py
```diff
@@ -142,7 +142,7 @@ def __init__(self):
         self.format_type = "ninjs"
         self.can_preview = True
         self.can_export = True
-        self.internal_renditions = ["original"]
+        self.internal_renditions = ["original", "baseImage"]
```
petrjasek, line 145: if we send those 2 we can as well send them all based on the configured renditions in settings, but would be probably good to introduce a new config for this so we can revert to the current behaviour if needed:
```py
NINJS_COMMON_RENDITIONS = list(RENDITIONS.keys())
```

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -120,9 +120,10 @@ def _transform_from_ninjs(self, ninjs):
                     associated_item["versioncreated"] = self.datetime(associated_item["versioncreated"])
                 item["associations"][key] = deepcopy(associated_item)
 
-        if ninjs.get("renditions", {}).get("baseImage"):
-            item["renditions"] = {"baseImage": {"href": ninjs.get("renditions", {}).get("original", {}).get("href")}}
-
+        renditions = ninjs.get("renditions", {})
+        if renditions.get("baseImage") or renditions.get("original"):
+            href = ninjs.get("renditions", {}).get("original", {}).get("href")
+            item["renditions"] = {"baseImage": {"href": href}, "original": {"href": href}}
```
petrjasek, line 126: I'm wondering why not populate all renditions we get?

petrjasek: would just check if they have all the data we need

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -120,10 +120,9 @@ def _transform_from_ninjs(self, ninjs):
                     associated_item["versioncreated"] = self.datetime(associated_item["versioncreated"])
                 item["associations"][key] = deepcopy(associated_item)
 
-        renditions = ninjs.get("renditions", {})
-        if renditions.get("baseImage") or renditions.get("original"):
-            href = ninjs.get("renditions", {}).get("original", {}).get("href")
-            item["renditions"] = {"baseImage": {"href": href}, "original": {"href": href}}
+        if ninjs.get("renditions"):
+            item["renditions"] = ninjs["renditions"]
```
petrjasek, line 124: I would still do some parsing there to only get relevant info and make sure it's all using correct types

petrjasek (commented):
some tests would be useful

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -146,6 +146,42 @@ def _transform_from_ninjs(self, ninjs):
 
         return item
 
+    def parse_renditions(self, renditions):
+        rend = []
```
petrjasek, line 150: it should return dict not a list

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -179,7 +179,7 @@ def parse_renditions(self, renditions):
                 parsed_rendition["media"] = media
 
             if parsed_rendition:
-                rend.append({rendition_name: parsed_rendition})
+                rend.update({rendition_name: parsed_rendition})
```
petrjasek, line 182: what about just doing `rend[rendition_name] = parsed_rendition`

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -159,8 +159,9 @@ def parse_renditions(self, renditions):
             # Parse width and height
             width = rendition_data.get("width")
             height = rendition_data.get("height")
-            if isinstance(width, int) and isinstance(height, int):
```
petrjasek: I think this might be better, either get both or none, otherwise later in the code we would have to check always if we have both

devketanpro: got it

petrjasek (approved):

PR #2495, opened 2023-11-24
add a new migration script to update width for existing profile fields [SDESK-7099]

devketanpro:


petrjasek:
would be better to make this a data migration not a standalone script, that should run everywhere shouldn't it?

petrjasek (approved):

PR #2505, opened 2024-01-02
increase publish transmit delay timeout on each retry [SDESK-7151]

devketanpro:


superdesk/publish/publish_content.py
```diff
@@ -186,12 +186,13 @@ def transmit_item(queue_item_id, is_async=False):
         retry_attempt_delay = app.config.get("TRANSMIT_RETRY_ATTEMPT_DELAY_MINUTES")
         try:
             orig_item = publish_queue_service.find_one(req=None, _id=queue_item["_id"])
+            timeout = min(2 ** min(6, orig_item.get("retry_attempt", retry_attempt_delay)), 60)
```
petrjasek, line 189: I don't think we need 2x min, 2 ** 6 is 64 so it won't get bigger than that

petrjasek (approved):

PR #2511, opened 2024-01-19
avoid search lookup in find_one call

petrjasek:
this is taking quite some time in NH during push:

find-one


MarkLark86 (approved):

MarkLark86:
@petrjasek I noticed too that in Newshub Agenda push, we're using `app.data.find_one("agenda", req=None, guid=item["guid"])`. `guid` has no mongo index defined, so we could improve Mongo lookup by using `_id` instead:
* https://github.com/superdesk/newsroom-core/blob/develop/newsroom/push.py#L89
* https://github.com/superdesk/newsroom-core/blob/develop/newsroom/push.py#L93

PR #2513, opened 2024-02-08
avoid extra push notifications when editing desk members

petrjasek:
- only updated visible stages for users added/removed from desk
- don't push `resource:updated` notifications when there are no fields updated.

SDANSA-526

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -298,7 +298,7 @@ class DesksService(BaseService):
                 users_service.update_stage_visibility_for_user(user)
 
             for removed_user in removed:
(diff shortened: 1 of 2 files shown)


apps/desks.py
```diff
@@ -294,8 +294,12 @@ def __send_notification(self, updates, desk):
                     desk=desk.get("name"),
                 )
                 push_notification("activity", _dest=activity["recipients"])
+                get_resource_service("users").update_stage_visibility_for_user(user)
```
MarkLark86, line 297: I see `get_resource_service("users")` 4 times in this function, might be good to store the service locally use it thereafter.

MarkLark86 (approved):

PR #2525, opened 2024-02-19
change the NINJS ingest parser to use the translations of the respective language from the respective CV [SDESK-7169]

devketanpro:


MarkLark86 (commented):
Some unit tests would be good too

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -182,16 +186,35 @@ def parse_renditions(self, renditions):
...
+        self, items: List[Dict[str, Any]], language: Optional[str] = None, cv_name: Optional[str] = None
+    ) -> List[Dict[str, Any]]:
         subjects = []
         for item in items:
-            subject = {"name": item.get("name"), "qcode": item.get("code")}
+            if cv_name and not item.get("translations"):
+                cv_data = self.find_cv_item(cv_name, item.get("code"))
+                item = cv_data if cv_data else item
+
+            subject = {
```
MarkLark86, line 198: After looking more into this, I realised that setting the `name` attribute to the appropriate translated value happens on publish (in the [NINJSFormatter](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/publish/formatters/ninjs_formatter.py#L77))

So what we should be doing here instead, is retrieving the `name`, `qcode`, `scheme` and `translations` attributes from the CV, falling back to the ingest source if not found.

devketanpro: hi yep, you are right, but Canadians don't publish audio files from Superdesk but export them from another system, so maybe their translation value is missing.

devketanpro: that's why I'm implementing this functionality here to retrieve the translation value from the CV if it is not found in the file.

MarkLark86: The `translations` field is not being populated on ingest. That is issue no. 1.
Issue 2. is we should use the CV entry first, and fallback to ingested value.

If a translation is missing, then it would be required in Superdesk's CV as well anyways

MarkLark86: And leave `name` as it is, and let the NINJS Formatter do the work (as it currently does with all other content types)

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -52,6 +54,8 @@ class NINJSFeedParser(FeedParser):
 
     items = []
 
+    _cv_items = {}
```
MarkLark86, line 57: This is not used anywhere

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -182,16 +186,31 @@ def parse_renditions(self, renditions):
...
         return rend
 
-    def _format_qcodes(self, items):
+    def _format_qcodes(
+        self, items: List[Dict[str, Any]], language: Optional[str] = None, cv_name: Optional[str] = None
+    ) -> List[Dict[str, Any]]:
         subjects = []
         for item in items:
-            subject = {"name": item.get("name"), "qcode": item.get("code")}
+            if cv_name:
```
MarkLark86, line 194: It would be better to retrieve the CV items before this for loop, and store the items there to use.
i.e.
```python
cv = get_resource_service("vocabularies").get_from_mongo(req=None, lookup={"_id": cv_name}) or {}
cv_items = {
    item["qcode"]: item
    for item in cv.get("items")
}
```

Then inside the for loop, you can use the following:
```python
if cv_item.get(item.get("code")):
    subject = cv_item[item["code"]
else:
    subject = {"name": item.get("name"), "qcode": item.get("code")}

if not subject.get("translations") and item.get("translations"):
    subject["translations"] = item["translations"]
if not subject.get("scheme") and item.get("scheme"):
    subject["scheme"] = item["scheme"]

subjects.append(subject)
```

superdesk/io/feed_parsers/ninjs.py
```diff
@@ -17,6 +17,9 @@
 from superdesk.io.feed_parsers import FeedParser
 from superdesk.utc import utc
 from superdesk.metadata.utils import generate_tag_from_url
+from typing import Optional, Dict, List, Any
+from superdesk import get_resource_service
+import collections
```
MarkLark86, line 22: Unused import

MarkLark86 (approved):

PR #2555, opened 2024-04-09
[SDESK-7156] Allow FeedingServices to provide request kwards for association downloading

MarkLark86:
This is used by AP Media API to provide x-api-key header, among others.

Also adds a new config `AP_MEDIA_API_VERIFY_SSL`, which defaults to True. This is to help prevent MITM style attacks, but can be turned off by env. variable if SSL verification fails for some reason.

### DIFF
--- a/superdesk/io/commands/update_ingest.py
+++ b/superdesk/io/commands/update_ingest.py
@@ -628,12 +628,11 @@ def ingest_item(item, provider, feeding_service, rule_set=None, routing_scheme=N
             _ingest_cancel(item, feeding_service)
 
         rend = item.get("renditions", {})
-        request_kwargs = feeding_service.get_request_kwargs()
         if rend:
             baseImageRend = rend.get("baseImage") or next(iter(rend.values()))
             if baseImageRend and not baseImageRend.get("media"):  # if there is media should be processed already
                 href = feeding_service.prepare_href(baseImageRend["href"], rend.get("mimetype"))
-                update_renditions(item, href, old_item, request_kwargs)
+                update_renditions(item, href, old_item, feeding_service=feeding_service)
 
         # if the item has associated media
         for key, assoc in item.get("associations", {}).items():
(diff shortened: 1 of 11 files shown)


superdesk/io/feeding_services/http_base_service.py
```diff
@@ -184,8 +184,13 @@ def get_url(self, url=None, **kwargs):
...
             kwargs["params"] = params
 
+        # Let the provided ``kwargs`` override the feeding service's ``kwargs``
+        request_kwargs = self.get_request_kwargs()
+        request_kwargs.update(kwargs)
+        request_kwargs.setdefault("timeout", self.HTTP_TIMEOUT)
+
         try:
-            response = requests.get(url, timeout=self.HTTP_TIMEOUT, **kwargs)
+            response = requests.get(url, **request_kwargs)
```
petrjasek, line 193: would be good to use a session here

superdesk/io/commands/update_ingest.py
```diff
@@ -628,11 +628,12 @@ def ingest_item(item, provider, feeding_service, rule_set=None, routing_scheme=N
             _ingest_cancel(item, feeding_service)
 
         rend = item.get("renditions", {})
+        request_kwargs = feeding_service.get_request_kwargs()
```
petrjasek, line 631: I think it might be useful eventually to pass the service instance to those methods, there could be different ways of downloading renditions which we wouldn't be able to handle via request args only

petrjasek (approved):

MarkLark86:
@petrjasek I have updated the implementation to use the feeding service like you suggested. Can you have another look for me please?

superdesk/media/renditions.py
```diff
@@ -403,7 +406,10 @@ def transfer_renditions(renditions, request_kwargs=None):
                 rend["href"] = app.media.url_for_media(rend["media"], local.content_type)
                 continue
 
-        content, filename, content_type = download_file_from_url(rend.get("href"), request_kwargs)
+        if feeding_service is not None and getattr(feeding_service, "download_file"):
+            content, filename, content_type = feeding_service.download_file(rend.get("href"), **request_kwargs or {})
+        else:
+            content, filename, content_type = download_file_from_url(rend.get("href"), request_kwargs)
```
petrjasek, line 412: those 2 sections with `if feeding service ...` seems like duplicates, could be extracted to another method

marwoodandrew (approved):

PR #2556, opened 2024-04-12
[SDBELGA-789] feature: New HTTP Proxy system

MarkLark86:


superdesk/http_proxy.py
```diff
@@ -0,0 +1,160 @@
...
+        return f"/{url_prefix}/{self.base_url}"
+
+    def process_request(self, path: str) -> FlaskResponse:
+        """The main function used for processing requests from the client"""
+
+        self.authenticate()
+        if request.method == "OPTIONS":
+            return self.construct_cors_response()
+
+        result = self.session.request(**self.get_proxy_request_kwargs())
```
petrjasek, line 85: might be useful to have this in a separate method so we can override this in a subclass but would still have that auth/cors handling there

tests/http_proxy_tests.py
```diff
@@ -0,0 +1,183 @@
...
+from urllib.parse import urlencode
+from flask import json, url_for
+
+from superdesk import __version__ as superdesk_version
+from superdesk.tests import TestCase, setup_auth_user
+from superdesk.http_proxy import HTTPProxy, register_http_proxy
+
+
+class HttpProxyTestCase(TestCase):
+    proxy = HTTPProxy("test_proxy", internal_url="test/proxy", external_url="http://localhost:5001/api")
```
petrjasek, line 14: I wonder if we need the ability to specify that `internal_url`, maybe using one route with: `/proxy/<endpoint_name>/<path>` would be enough, otherwise it will be possible to override internal apis

petrjasek (approved):

PR #2558, opened 2024-04-15
[SDCP-743] Manual editing of the events

devketanpro:


superdesk/io/commands/update_ingest.py
```diff
@@ -586,6 +586,9 @@ def ingest_item(item, provider, feeding_service, rule_set=None, routing_scheme=N
                 f"updating items is disabled on provider '{provider['name']}'."
             )
             return False, []
+        elif old_item and old_item.get("version_creator"):
```
petrjasek, line 589: I wonder if this would somehow affect the articles as well, might be good to call here the respective service and keep the logic there so we can handle it specifically for events, eg.:
```py
elif old_item and not ingest_service.should_update(old_item, item, provider):
```
and have this `old_item.get("version_creator")` in events ingest service
and add some default implementation in the core `IngestService`:
```py
def should_update(self, old_item, new_item, provider):
    return True
```

MarkLark86:
@petrjasek @devketanpro I added a comment on the CP PR, https://github.com/superdesk/superdesk-cp/pull/229#discussion_r1566553735, if this helps here

superdesk/io/commands/update_ingest.py
```diff
@@ -586,7 +586,7 @@ def ingest_item(item, provider, feeding_service, rule_set=None, routing_scheme=N
                 f"updating items is disabled on provider '{provider['name']}'."
             )
             return False, []
-        elif old_item and old_item.get("version_creator"):
+        elif old_item and not ingest_service.should_update(old_item):
```
petrjasek, line 589: I would pass there more than just the old item, we might need it in other cases

superdesk/io/commands/update_ingest.py
```diff
@@ -587,7 +587,7 @@ def ingest_item(item, provider, feeding_service, rule_set=None, routing_scheme=N
             )
             return False, []
         elif old_item and not ingest_service.should_update(old_item, item, provider):
-            logger.warning(f"Resource '{ingest_collection}' " f"item '{item[GUID_FIELD]}' is already edited in SD .")
+            logger.info(f"Resource '{ingest_collection}' " f"item '{item[GUID_FIELD]}' is already edited in SD .")
```
petrjasek, line 590: actually here the reason is unknown, so you can change it to something like `item should not be updated.` or you can move the logging to events service where you know why it was rejected

petrjasek (approved):

PR #2570, opened 2024-05-15
SDAAP-105 Cherry pick SDESK-7156 to pass token in header for AP Media…

marwoodandrew:
… ingest

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -14,7 +14,7 @@ pytest
 pytest-env
 python3-saml>=1.9,<1.10
 typing_extensions>=3.7.4
-moto[sqs]
+moto[sqs]<5.0
 
 -e .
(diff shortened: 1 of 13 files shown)


MarkLark86:
@marwoodandrew If you cherry-pick this commit as well, this should fix the build (https://github.com/superdesk/superdesk-core/commit/bbe63a766f719a49659c7a4f8856dd1576360a5f)

petrjasek:
for this older version it might be better to pick this one https://github.com/superdesk/superdesk-core/pull/2560

marwoodandrew:
@MarkLark86 @petrjasek I'm not sure that the lxml was the problem, I wound up limiting moto[sqs] and pulling in part of a PR to fix the wordpress related test (no idea why it broke!) . Documentation is still busted.

petrjasek (approved):

PR #2580, opened 2024-05-21
update urllib3

petrjasek:
1.25 is not supported anymore and does not get security fixes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,11 +13,7 @@ from setuptools import setup, find_packages
 LONG_DESCRIPTION = "Superdesk Server Core"
 
 install_requires = [
-    # temporary requirement to get urllib in a version compatible with requests
-    # to be kept until requests update its requirements
-    # (cf. https://github.com/psf/requests/issues/5654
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -13,11 +13,7 @@
 LONG_DESCRIPTION = "Superdesk Server Core"
 
 install_requires = [
-    # temporary requirement to get urllib in a version compatible with requests
-    # to be kept until requests update its requirements
-    # (cf. https://github.com/psf/requests/issues/5654
-    # and https://github.com/psf/requests/pull/5651)
-    "urllib3<1.26",
+    "urllib3>=1.26,<3",
```
MarkLark86, line 16: We could probably remove `urllib3` completely from our requirements, as the `requirements` lib already defines it in theirs (https://github.com/psf/requests/blob/main/setup.py#L64)

petrjasek: it's actually dependency also for elasticsearch and botocore, but we also import it so would keep this one explicit

MarkLark86 (approved):

PR #2601, opened 2024-06-10
update elastic mapping of item subject field [SDESK-7268]

devketanpro:
https://github.com/superdesk/superdesk-analytics/pull/151

### DIFF
--- a/superdesk/metadata/item.py
+++ b/superdesk/metadata/item.py
@@ -220,7 +220,12 @@ metadata_schema = {
     },
     "subject": {
         "type": "list",
-        "mapping": {"type": "object", "dynamic": False, "properties": {"qcode": not_analyzed, "name": not_analyzed}},
+        "mapping": {
+            "type": "nested",
+            "include_in_parent": True,
+            "dynamic": False,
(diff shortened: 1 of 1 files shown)


superdesk/metadata/item.py
```diff
@@ -220,7 +220,12 @@ class Formats(NamedTuple):
     },
     "subject": {
         "type": "list",
-        "mapping": {"type": "object", "dynamic": False, "properties": {"qcode": not_analyzed, "name": not_analyzed}},
+        "mapping": {
+            "type": "nested",
+            "include_in_parent": True,
```
MarkLark86, line 225: @petrjasek Unfortunately for Analytics we would need nested mapping here, so aggregation on scheme -> qcode/name can be done.

`include_in_parent` is there so non-nested queries for subject still works, but I fear adding nested here will increase the resource requirements within elastic as we're adding a nested doc for each element in this list.

Do you believe this would be OK?

petrjasek: I think it's ok, in general in superdesk there should be less content than in eg. newshub with big archive and there we are using it too

MarkLark86 (approved):

PR #2608, opened 2024-06-12
[SDESK-7276] New core app/module framework

MarkLark86:


petrjasek (approved):

docs/requirements.txt
```diff
@@ -6,48 +6,58 @@
...
+    # via uvicorn
 idna==3.7
-    # via requests
+    # via
+    #   anyio
+    #   requests
 imagesize==1.4.1
     # via sphinx
-jinja2==3.1.4
+jinja2==2.11.3
```
eos87, line 37: Why are we going down in the version here?

tests/core/app_test.py
eos87, line 1: I see that each test case is instantiating the app with something like `SuperdeskAsyncApp(WSGI(config={...}))`. I would suggest that we have a `setUp` method and then just override the `config` in each case if it's needed. Something like 

```python
class SuperdeskAsyncAppTestCase(TestCase):
    def setUp(self):
        self.config = {}
        self.app = SuperdeskAsyncApp(WSGI(config=self.config))
``` 

then the cases could be simplified. One idea would be to split `test_construction` into something like:

```python
def test_app_initial_state_is_not_running(self):
    self.assertFalse(self.app.running, "App should not be running immediately....")

def test_starting_app_changes_state_to_running(self):
    self.app.start()
    self.assertTrue(self.app.running, "App should be running after start() is called")

def test_starting_running_app_raises_error(self):
    self.app.start()
    with self.assertRaises(RuntimeError, msg="Already running app should raise RuntimeError"):
        self.app.start()
```

Something similar could be done to `test_loading_modules`. What do you think?

eos87 (commented):
It looks good, I just left a couple of comments. Please take a look and let me know your thoughts.

docs/requirements.txt
MarkLark86: Causes a version conflict, see https://github.com/superdesk/superdesk-core/pull/2607. This should be fixed when we upgrade Flask etc

tests/core/app_test.py
MarkLark86: @eos87 I have done this in my next PR

PR #2617, opened 2024-06-24
Languagetool spellchecker support

antonborgstrom:
## What does this PR do?
- Adds support for the spellchecker LanguageTool (https://languagetool.org)

### How to test?
Add LanguageTool env-variables to settings.py:
- LANGUAGETOOL_API_URL = env('LANGUAGETOOL_API_URL', 'https://api.languagetoolplus.com/v2/')
- No need to specify API-key for testing against public url above 

### Dependencies
- Additional languages must be added to superdesk-client-core in `scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx` 
- Dependent on superdesk-client-core [PR 4560](https://github.com/superdesk/superdesk-client-core/pull/4560)

### DIFF
--- a/superdesk/text_checkers/spellcheckers/languagetool.py
+++ b/superdesk/text_checkers/spellcheckers/languagetool.py
@@ -1,0 +1,186 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013-2019 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+import os
+import logging
+import requests
+import json
+from superdesk.errors import SuperdeskApiError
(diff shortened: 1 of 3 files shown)


superdesk/text_checkers/spellcheckers/languagetool.py
```diff
@@ -0,0 +1,166 @@
...
+            'language': language or 'auto',
+        }
+
+        # Add apiKey and additional options from the environment variable
+        additional_options = self.languagetool_config
+        payload.update(additional_options)
+
+        try:
+            # Send the POST request to the LanguageTool API
+            response = requests.post(self.api_url + '/check', data=payload)
```
petrjasek, line 79: would be good to use a session, that will allow it to reuse connections

petrjasek (approved):

superdesk/text_checkers/spellcheckers/languagetool.py
antonborgstrom: Good point! Added now.

PR #2622, opened 2024-06-26
[SDESK-7291] Feature: New config system

MarkLark86:


MarkLark86:
@petrjasek @eos87 Updated docs can be found here: https://superdesk--2622.org.readthedocs.build/en/2622/#new-core-framework. I have improved the layout of the docs for his new core.

docs/core/modules.rst
```diff
@@ -0,0 +1,132 @@
...
+    class ModuleConfig(ConfigModel):
+        # Make sure to provide defaults for all fields
+        # otherwise validation would fail
+        uri: str = "localhost"
+        port: int = 27017
+        options: Optional[Dict[str, Any]] = None
+
+    # Create a module local instance of the class,
+    # to be used at runtime
+    config = ModuleConfig()
```
petrjasek, line 74: I'm wondering if it wouldn't be better usage to call it with app so when it's initialized it will actually have the values already populated, eg.:
```python3
app = get_current_app()
config = ModuleConfig(app)
```
otherwise this could in theory lead to usage without proper initialization.

docs/core/modules.rst
```diff
@@ -0,0 +1,132 @@
...
+        # The module's config is loaded BEFORE executing the
+        # modules `init` function
+        if config.uri == "localhost":
+            print("Using a local instance")
+        else:
+            print("Using a remote instance")
+
+        # You can also import configs from other modules
+        # all configs are loaded BEFORE executing module's
+        # init function
```
petrjasek, line 87: I'm not sure if we should try to expose module config to other modules, that kind of breaks module encapsulation. Not that we should try to prevent it, but would probably not document that part.
Maybe we should have some section in general what to do and what not to do.

docs/core/modules.rst
MarkLark86: I'd rather have a direct dependence of App for Config. At the moment it's pure python/pedantic.

App is only available after app has been built.

We can place in a function on the ConfigModel class that does the loading, and sets an attribute saying it's loaded. Then when trying to access a variable, we can throw an exception if it's not loaded.

docs/core/modules.rst
MarkLark86: Why can't one module depend on another? Configs are global anyway, this way it's typed and validated on load

petrjasek: I don't consider it a public interface of a module, but that might be just my view

docs/core/modules.rst
petrjasek: I think it should actually depend on it rather explicitly than implicitly. Using the config before the app was loaded is always wrong anyhow, isn't it?

petrjasek: but might be more complicated to test different configs

petrjasek: I see

docs/core/modules.rst
MarkLark86: One reason why I made it a module local instance, is so other files in that module can import it directly and use, with strict typing.

The Configs are frozen by default, so they cannot be changed once loaded. Would that suffice for now, that we document do's and don'ts

MarkLark86:
@petrjasek I have updated the PR with a RuntimeError raised if accessing a config attribute before it's been loaded.
What do you think?

petrjasek (approved):

PR #2624, opened 2024-07-01
[SDESK-7278] New Elasticsearch datalayer

MarkLark86:


MarkLark86:
Latest docs: https://superdesk--2624.org.readthedocs.build/en/2624/

docs/core/elastic.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Elasticsearch
+=============
+
+.. module:: superdesk.core.elastic
+
+ElasticResources Instance
+-------------------------
+The :class:`ElasticResources <resources.ElasticResources>` instance provides access to the Elasticsearch clients. This instance is available under the
+:attr:`SuperdeskAsyncApp.mongo <superdesk.core.app.SuperdeskAsyncApp.mongo>`.
```
petrjasek, line 11: I think there is typo `s/mongo/elastic/`

setup.py
```diff
@@ -16,7 +16,7 @@
     "urllib3>=1.26,<3",
     "eve>=1.1.2,<=2.1.0",
     "eve-elastic>=7.4.0,<7.5.0",
-    "elasticsearch<7.14",  # we are using oss version on test server
+    "elasticsearch[async]<7.14",  # we are using oss version on test server
```
petrjasek, line 19: we're using 7.17 now on test server so if needed you can use more recent version

superdesk/core/elastic/resources.py
```diff
@@ -0,0 +1,398 @@
...
+        """Find a document in multiple registered resources and Elasticsearch indexes
+
+        :param item_id: The id of the item to find
+        :param resource_names: A list of names of the registered resources to search in
+        :return: The document if found, else None
+        :raises KeyError: If any provided resource name is not registered for use with Elasticsearch
+        """
+
+        for resource_name in resource_names:
+            doc = self.get_client(resource_name).find_by_id(item_id)
```
petrjasek, line 343: in general it can be searching multiple indexes at the same time, as long as they are using the same connection

petrjasek (approved):

superdesk/core/resources/fields.py
```diff
@@ -0,0 +1,188 @@
...
+        return core_schema.int_schema()
+    elif class_type == float:
+        return core_schema.float_schema()
+    elif class_type == bool:
+        return core_schema.bool_schema()
+    elif class_type == bytes:
+        return core_schema.bytes_schema()
+    elif class_type == datetime:
+        return core_schema.datetime_schema()
+    raise RuntimeError(f"Unsupported base class type: {class_type}")
```
eos87, line 46: May I suggest an alternative version to improve readability. What do you think?

```suggestion
def get_core_schema_from_type(class_type: Type) -> core_schema.CoreSchema:
    schema_mapping = {
        str: core_schema.str_schema,
        int: core_schema.int_schema,
        float: core_schema.float_schema,
        bool: core_schema.bool_schema,
        bytes: core_schema.bytes_schema,
        datetime: core_schema.datetime_schema,
    }

    schema_function = schema_mapping.get(class_type)
    if schema_function:
        return schema_function()
    
    raise RuntimeError(f"Unsupported base class type: {class_type}")
```

tests/core/elastic_sync_test.py
```diff
@@ -0,0 +1,189 @@
...
+    code=Keyword("my codes"),
+    categories=[
+        Category(qcode="sports", name="Sports", scheme="category"),
+        Category(qcode="swimming", name="Swimming", scheme="sports"),
+    ],
+    related_items=[
+        RelatedItems(id=ObjectId(), link_type=Keyword("text"), slugline=HTML("sports-results")),
+        RelatedItems(id=ObjectId(), link_type=Keyword("photo"), slugline=HTML("sports-results")),
+    ],
+)
```
eos87, line 24: We should put this into a separate function instead of having it statically setup, specially because it is used in multiple tests.

tests/core/elastic_sync_test.py
eos87, line 1: I'd say we should have a "cleanup" step to reset the state of Elastic between tests? At the moment the tests seem dependant between each other and future changes to one of them would break others. What do you think?

eos87 (commented):
It looks good. I just left a few minor comments. Please have a look when you have some time.

superdesk/core/elastic/resources.py
MarkLark86: Looks like the underlying request uses the [docs-get](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-get.html) endpoint of Elasticsearch, which by the look from their API reference (both in python code and that website) the index param only allows 1.

I'm thinking because Elastic is not actually searching, but returning a document directly based on ID. This should be fast enough. We also have the `app.elastic.search` function that allows searching across indexes of the same connection, if required.

tests/core/elastic_sync_test.py
MarkLark86: This is already happening. In `superdesk.tests.asyncio.AsyncTestCase.asyncTearDown` we're calling `self.app.elastic.drop_indexes()` which deletes all registered resource indexes.

superdesk/core/app.py
```diff
@@ -111,3 +118,5 @@ def get_current_app() -> SuperdeskAsyncApp:
 
 from .module import Module  # noqa: E402
 from .mongo import MongoResources  # noqa: E402
+from .elastic import ElasticResources  # noqa: E402
+from .resources import Resources  # noqa: E402
```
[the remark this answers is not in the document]
MarkLark86: Looks like Flake8 won't implement per file ignores for 1 specific rule. For now I think this will suffice

superdesk/commands/index_from_mongo.py
```diff
@@ -48,21 +50,36 @@ def run(self, collection_name, all_collections, page_size, last_id, string_id):
         elif all_collections:
             app.data.init_elastic(app)
             resources = app.data.get_elastic_resources()
+            resources_processed = []
+            for resource_config in get_current_app().resources.get_all_configs():
```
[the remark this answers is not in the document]
MarkLark86: Using `app.async_app.resources` would not give us type checked code. Also I would prefer to keep Flask specifics out of the new async code as much as possible, hence a helper function to specifically get the async app only.

I have renamed that helper function as you suggested.

superdesk/commands/index_from_mongo.py
```diff
@@ -89,7 +106,11 @@ def get_mongo_items(cls, mongo_collection_name, page_size, last_id, string_id):
         bucket_size = int(page_size) if page_size else cls.default_page_size
         print("Indexing data from mongo/{} to elastic/{}".format(mongo_collection_name, mongo_collection_name))
 
-        db = app.data.get_mongo_collection(mongo_collection_name)
+        try:
+            db = get_current_app().mongo.get_collection(mongo_collection_name)
+        except KeyError:
+            db = app.data.get_mongo_collection(mongo_collection_name)
```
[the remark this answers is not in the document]
MarkLark86: These `try/except` would only be in these low level parts of the async app. Most devs would not need to use the mongo/elastic clients directly, but rather through a resource service like we do now.

This will be in my next PR, which allows a dev to interact with resources without knowing the underlying data layer(s).
```python
from my_module.users import UsersService, User

async def test_users():
    service = UsersService()  # automatic singleton instance
    user = User(id="user_1", first_name="Monkey", last_name="Magic")
    await service.create([user])
    my_user = await service.get_by_id("user_1")
    ...
```

superdesk/commands/rebuild_elastic_index.py
```diff
@@ -41,7 +42,21 @@ def run(self, resource_name=None, requests_per_second=1000):
             resources = [resource_name]
         elif resource_name:
             raise ValueError("Resource {} is not configured".format(resource_name))
+
+        resources_processed = []
+        new_app = get_current_app()
+        for config in new_app.resources.get_all_configs():
```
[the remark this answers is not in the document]
MarkLark86: I don't foresee many places where we will have to work on all registered resources. If this turns out to not be the case, we can revisit this

tests/core/elastic_sync_test.py
MarkLark86: I will add a fixtures folder to `tests/core` that we can store model instances for use throughout test cases (we do something similar in Newshub, so test cases work against a common group of fixtures)

superdesk/commands/index_from_mongo.py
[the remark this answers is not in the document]
eos87: got it!

PR #2629, opened 2024-07-08
introduce a specific settings in the user profile to opt in or out for assignement and Mark for user mails [SDBELGA-818]

devketanpro:
Related PR's

- https://github.com/superdesk/superdesk-client-core/pull/4566
- https://github.com/superdesk/superdesk-planning/pull/2018

### DIFF
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -1276,7 +1276,7 @@ class ArchiveService(BaseService, HighlightsSearchMixin):
         """
 
         # No notification sent if user is not enabled mark for user notification
-        filter_user_list = [
+        users_with_enabled_notifications = [
             user
             for user in user_list
             if superdesk.get_resource_service("preferences").mark_for_user_notification_is_enabled(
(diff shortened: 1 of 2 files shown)


apps/archive/archive.py
```diff
@@ -1275,6 +1275,13 @@ def _send_mark_user_notifications(
         :param data: kwargs
         """
 
+        user = get_user()
```
petrjasek, line 1278: it should check users from user list, not the current user imo

devketanpro: got it

apps/archive/archive.py
```diff
@@ -1275,6 +1275,15 @@ def _send_mark_user_notifications(
         :param data: kwargs
         """
 
+        # No notification sent if user is not enabled mark for user notification
+        filter_user_list = [
```
petrjasek, line 1279: could use a better name like `users_with_enabled_notifications`

petrjasek (approved):

PR #2632, opened 2024-07-09
[SDESK-7281] Upgrade Flask to v3

eos87:
This PR is far from being ready to be merged. It has been only opened to allow collaboration with other team members. 

[SDESK-7281]

[SDESK-7281]: https://sofab.atlassian.net/browse/SDESK-7281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


setup.py
```diff
@@ -17,9 +17,8 @@
     "eve>=1.1.2,<=2.1.0",
     "eve-elastic>=7.4.0,<7.5.0",
     "elasticsearch[async]<7.18",  # we are using oss version on test server
-    "flask>=1.1,<1.2",
+    "flask>=3.0",
```
MarkLark86, line 20: We will need to include the `async` extra for Flask.
i.e. `flask[async]>=3.0`

MarkLark86:
@eos87 I'm okay with merging this PR when you're ready. Our current focus is Newshub (which works with this branch), and we will need more PRs against superdesk-core. Merging this will help there.

As we already have tickets to fix deprecations, we can use that ticket to make Superdesk run and tests pass etc.

setup.py
eos87: done

eos87:
> @eos87 I'm okay with merging this PR when you're ready. Our current focus is Newshub (which works with this branch), and we will need more PRs against superdesk-core. Merging this will help there.
> 
> As we already have tickets to fix deprecations, we can use that ticket to make Superdesk run and tests pass etc.

I agree. I will merge it in a little while.

PR #2635, opened 2024-07-11
[SDESK-7289] AsyncResourceSerivce and async validation

MarkLark86:


superdesk/core/app.py
```diff
@@ -105,15 +111,55 @@ def stop(self):
...
+        from flask import current_app
+
+        try:
+            setattr(current_app, "async_app", None)
+        except RuntimeError:
+            # Flask context not available
+            pass
+
+        global _global_app
+        _global_app = None
```
eos87, line 139: What is the reason for the global here? Just curious. Normally what I've read in the past is that the usages of globals are not encouraged.

eos87 (approved):
Looks good

superdesk/core/app.py
MarkLark86: The `_global_app` is stored in the module's root scope (as it has to be stored somewhere, for later retrieval).
Without the use of `global`, assignment to that variable won't work inside a function scope.
I agree I don't like it much, and ATM is mainly there for tests.

We can improve this later (I also have ideas I'd like to implement, regarding module loading, circular references, module file structure, app passing etc).

PR #2645, opened 2024-07-23
Fix: Marked for user desktop notifications [SDBELGA-818]

devketanpro:
Previously, when users disabled email notifications, desktop notifications were also turned off.

### DIFF
--- a/apps/preferences.py
+++ b/apps/preferences.py
@@ -421,11 +421,13 @@ class PreferencesService(BaseService):
         send_email = preferences.get("assignment:notification", {}) if isinstance(preferences, dict) else {}
         return send_email and send_email.get("enabled", False)
 
-    def mark_for_user_notification_is_enabled(self, user_id=None, preferences=None):
+    def check_notification_is_enabled(self, preference_name, user_id=None, preferences=None):
         """
         This function checks if email notification is enabled or not based on the preferences.
         """
+        send_email = {}
         if user_id:
             preferences = self.get_user_preference(user_id)
-        send_email = preferences.get("mark_for_user:notification", {}) if isinstance(preferences, dict) else {}
+        if preference_name:
+            send_email = preferences.get(f"{preference_name}:notification", {}) if isinstance(preferences, dict) else {}
         return send_email and send_email.get("enabled", False)
--- a/apps/preferences.py
+++ b/apps/preferences.py
@@ -213,7 +213,7 @@ class PreferencesResource(Resource):
     )
 
     superdesk.register_default_user_preference(
-        "mark_for_user:notification",
+        "email:notification:mark_for_user",
         {
             "type": "bool",
             "enabled": True,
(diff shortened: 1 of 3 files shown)


superdesk/activity.py
```diff
@@ -228,7 +228,9 @@ def add_activity(
     return activity
 
 
-def notify_and_add_activity(activity_name, msg, resource=None, item=None, user_list=None, **data):
+def notify_and_add_activity(
+    activity_name, msg, resource=None, item=None, user_list=None, users_with_enabled_email_notifications=None, **data
```
petrjasek, line 232: would be good to rename `user_list` as well it's not obvious who are those users

superdesk/activity.py
```diff
@@ -242,8 +244,8 @@ def notify_and_add_activity(activity_name, msg, resource=None, item=None, user_l
         **data,
     )
 
-    if activity_name == ACTIVITY_ERROR or user_list:
-        recipients = get_recipients(user_list, activity_name)
+    if activity_name == ACTIVITY_ERROR or users_with_enabled_email_notifications:
+        recipients = get_recipients(users_with_enabled_email_notifications, activity_name)
```
petrjasek, line 248: actually I'm wondering why we need to pass users with enabled notifications if it checks that in the `get_recipients` method?

devketanpro: In the `get_recipients` method, we check `email_notifications`, and in the `users_with_enabled_email_notifications`, we have users whose `mark_for_user` notifications are enabled, both preferences are different.

petrjasek: I see, it would make sense then probably to have this code somehow generic and do the preferences based on activity name, so we don't have to implement some extra code for pre-filtering users each time we want to have it configurable. so that `get_recipients` method would look at `f'email_notification_{activity_name}'` preference

devketanpro: got it, Thanks

apps/preferences.py
```diff
@@ -421,11 +421,13 @@ def assignment_notification_is_enabled(self, user_id=None, preferences=None):
...
+    def check_notification_is_enabled(self, preference_name, user_id=None, preferences=None):
         """
         This function checks if email notification is enabled or not based on the preferences.
         """
+        send_email = {}
         if user_id:
             preferences = self.get_user_preference(user_id)
-        send_email = preferences.get("mark_for_user:notification", {}) if isinstance(preferences, dict) else {}
+        if preference_name:
+            send_email = preferences.get(f"{preference_name}:notification", {}) if isinstance(preferences, dict) else {}
```
petrjasek, line 432: imo we should have a specific `email_notification` later we might have separate in app notification setting, it's not clear which one is that

superdesk/activity.py
```diff
@@ -229,7 +229,7 @@ def add_activity(
 
 
 def notify_and_add_activity(
-    activity_name, msg, resource=None, item=None, user_list=None, users_with_enabled_email_notifications=None, **data
+    activity_name, msg, resource=None, item=None, user_list=None, preference_notification_name=None, **data
```
petrjasek, line 232: why not use the activity_name for the notification settings?

devketanpro: I thought the `activity_name`(`item:marked`, `item:unmarked`) is used in emails

apps/preferences.py
devketanpro: this one is for mark_for_user mails, it's a child of `"email:notification" `, if the parent `email:notification` is disabled, none of email is send

superdesk/activity.py
petrjasek: I see, makes sense then

apps/preferences.py
petrjasek: I'm not sure how that logic would work if we make it configurable for app notifications, so imo there should be some indication this is about email. if we have some concept of parent child, it should be more obvious imo.
I suppose `desktop:notification` is not child of `email:notification`. maybe it should be then `mark_for_user:email:notification` or `email:notification:mark_for_user` if we want to have some hierarchy

devketanpro: makes sense

petrjasek (approved):

PR #2648, opened 2024-07-29
[SDESK-7285] Create Async GridFS MediaStorage client

eos87:
### Purpose
Introduce the GridFSMediaStorageAsync class, which provides an asynchronous interface for storing, retrieving, and managing files in MongoDB's GridFS. The class includes methods for storing files with metadata, retrieving files by ID or filename, searching for files based on criteria, checking the existence of files, and deleting files. Documentation and usage examples have been also provided.

### What has changed
- Added GridFSMediaStorageAsync class for asynchronous file operations in GridFS.
- Implemented methods for storing, retrieving, searching, checking existence, and deleting files.
- Updated documentation with examples for each method.
- Added Sphinx documentation to include the new class and its methods.

Resolves: [SDESK-7285]

[SDESK-7285]: https://sofab.atlassian.net/browse/SDESK-7285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


superdesk/storage/async_storage/media_storage.py
```diff
@@ -0,0 +1,265 @@
...
+
+from eve.io.media import MediaStorage
+from superdesk.core.app import SuperdeskAsyncApp
+from superdesk.storage.desk_media_storage import format_id
+from superdesk.storage.mimetype_mixin import MimetypeMixin
+
+logger = logging.getLogger(__name__)
+
+
+class GridFSMediaStorageAsync(MediaStorage, MimetypeMixin):
```
MarkLark86, line 28: We don't want to use any `eve` at all for new code. One of the goals is to completely remove Eve from all our code.

superdesk/storage/async_storage/media_storage.py
```diff
@@ -0,0 +1,265 @@
...
+
+    .. versionadded:: 3.0
+    """
+
+    _fs: Dict[str, AsyncIOMotorGridFSBucket]
+
+    app: SuperdeskAsyncApp
+
+    def __init__(self, app: SuperdeskAsyncApp):
+        self.app = app
```
MarkLark86, line 41: It would be good to not rely on any app specifics, making this storage interface generic, only relying on the PyMongo library. We can then add async GridFS functionality into the existing async mongo module (under `superdesk.core.mongo`).

superdesk/storage/async_storage/media_storage.py
```diff
@@ -0,0 +1,265 @@
+# -*- coding: utf-8; -*-
```
MarkLark86, line 1: It might also be good to have this module under `superdesk.core` somewhere, to differentiate between the old and the new.

superdesk/storage/async_storage/media_storage.py
eos87: Removed inheritance from `MediaStorage`

superdesk/storage/async_storage/media_storage.py
eos87: done.

superdesk/storage/async_storage/media_storage.py
eos87: done

eos87:
Hi @MarkLark86, I've implemented your suggestions and also updated [the documentation](https://superdesk--2648.org.readthedocs.build/en/2648/). Please take another look when you have some time. Thanks!

superdesk/core/storage.py
```diff
@@ -0,0 +1,261 @@
...
+
+        :param resource: The resource to use. Defaults to "upload".
+        :return: The GridFS bucket.
+        """
+        resource = resource or "upload"
+        mongo = get_current_async_app().mongo
+        px = mongo.get_resource_config(resource).prefix
+
+        if px not in self._fs:
+            _, db = mongo.get_client_async(resource)
```
MarkLark86, line 239: I was thinking here, if `resource` is not registered with the new system, should we fallback to the default mongo connection config (like we're now doing with the [MongoLock](https://github.com/superdesk/superdesk-core/blob/async/superdesk/lock.py#L46))?

We could easily do this by wrapping this call into a `try...except KeyError` block, and in the exception code block get the default config using `get_mong-client_config`. This would also mean you don't need the upload resource in the tests.

Thoughts?

MarkLark86 (approved):
Looks good. I do have concerns around async streaming vs consuming the entire request, but that can be looked at later when we upgrade to Quart

superdesk/core/storage.py
MarkLark86: An improvement might be to add a `get_default_client_async` function onto the [MongoResources](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/mongo.py#L172) class that will return a mongo client/db, and use that here (and in the superdesk.lock module)

petrjasek: I would rather raise an error in that case, this could be error prone when you have a typo in resource you might get different connection, most of the time we get the same but in those few times when it matters it might be doing something unexpected

eos87: I think I agree with the idea of raising an error instead and let the developer handle it.

PR #2650, opened 2024-07-30
[SDESK-7325] Move flask imports to superdesk module

MarkLark86:
Also replace usage of `eve.utils.config` with `superdesk.core.get_app_config`


eos87 (approved):
Looks good

PR #2651, opened 2024-07-30
[SDESK-7348] fix: Use MongoClient directly for MongoLocks

MarkLark86:


eos87:
Hi @MarkLark86, I think I've missed the conversation about this one. Just curious, what is the reason for this change? Thanks!

MarkLark86:
For some reason it was always trying to connect to localhost.

So I changed it to use the default configuration, from the same config system used by the new core

eos87 (approved):

PR #2655, opened 2024-08-06
Register notifications object

thecalcc:
SDBELGA-818


petrjasek:
seems like I made a mistake here https://github.com/superdesk/superdesk-core/blob/develop/apps/preferences.py#L219 , can you remove it?

petrjasek (approved):

PR #2658, opened 2024-08-07
[SDESK-7287] Support async tasks in Celery

eos87:
_**Note: work-in-progress**_ 

### Purpose
This pull request introduces enhanced support for both synchronous and asynchronous tasks in our Celery setup, encapsulated within the Flask application context. 

### What has changed
- Introduced a new class `HybridAppContextTask` that extends celery.Task to handle asynchronous task execution alongside traditional synchronous tasks.
- Reworked the celery serializer methods into a separate class factory. The `ContextAwareSerializerFactory` simplifies the creation of serializers that automatically work within the Flask application context, using the provided get_current_app function. It will helps avoid code duplication like the one in https://github.com/superdesk/newsroom-core/blob/cd39f5eb5a0dfc5d4ef415fc6feb1f57d252bf7f/newsroom/celery_app.py#L44-L81
- Added and improved tests

Resolves: [SDESK-7287]


tests/celery_app/context_test.py
```diff
@@ -0,0 +1,48 @@
...
+from unittest.mock import patch
+
+from superdesk.errors import SuperdeskError
+from superdesk.celery_app import HybridAppContextTask
+from superdesk.tests.asyncio import AsyncFlaskTestCase
+
+
+class TestHybridAppContextTask(AsyncFlaskTestCase):
+    async def test_sync_task(self):
+        @self.app.celery.task(base=HybridAppContextTask)
```
MarkLark86, line 11: Is `base=HybridAppContextTask` required here, when we're setting `celery.Task = HybridAppContextTask`?

eos87: I added it for two reasons: 
- It is explicit therefore easier to understand what we're testing
- If for some reason the `celery.Task = ` value is different in the config as some point, the test of the `HybridAppContextTask` class will still work/pass.

superdesk/celery_app/context_task.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        """
+        Executes the task function, determining if it should be run synchronously or asynchronously.
+
+        Args:
+            args: Positional arguments passed to the task function.
+            kwargs: Keyword arguments passed to the task function.
+        """
+        with self.get_current_app().app_context():
```
MarkLark86, line 40: This will need changes once we go to Quart, as the `app_context` in Quart is async. But should be good for now

superdesk/celery_app/context_task.py
```diff
@@ -0,0 +1,96 @@
...
+        # all exceptions are managed and logged regardless of where they occur within the event loop
+        async def wrapper():
+            try:
+                return await task_func(*args, **kwargs)
+            except self.app_errors as e:
+                self.handle_exception(e)
+                return None
+
+        if not loop.is_running():
+            return loop.run_until_complete(wrapper())
```
MarkLark86, line 81: Does the Celery app would use this code path, using `run_until_complete`, where as tests use the `create_task` response?

Reason I ask is, because `create_task` returns an instance of [asyncio.Task](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task) and not the actual methods response.

MarkLark86: I know the issue here, and a big pain point with Python async, is that once an event loop is already running, there doesn't seem to be an easy way to add a task to the event loop and await it's response from within a sync method

MarkLark86: If the Celery app uses `run_until_complete` when running the celery task, then this should be fine (with knowing that tests may fail to run properly)

eos87: > I know the issue here, and a big pain point with Python async, is that once an event loop is already running, there doesn't seem to be an easy way to add a task to the event loop and await it's response from within a sync method

Indeed it is a big pain 😅 Actually that is the reason why I decided to go with `asyncio.create_task`, because if the event loop is already running, it schedules the execution of the task and returns an awaitable object. My initial intention was that I wanted to do the same when the event loop is _**not**_ running (to return an awaitable) but it was pretty complex to try to handle the loop creation, then scheduling the execution of the task, returning the awaitable object and then closing the loop. So the only alternative I had to make sure the task will be scheduled and actually executed, was to use `run_until_complete` which makes sure to run the event loop until the Future is done. 

**The downsides**
- Consistency in return type: Ideally, I wanted all execution paths to return the same type (asyncio.Future), but due to the differences in how tasks are handled based on the event loop's state, this wasn't feasible.
- Blocking execution: If the event loop is not running, the use `run_until_complete` (which is blocking as far as I understand) would make us sacrifice the async benefits.

**The benefits**
- Guaranteed tasks execution: the current approach makes sure to execute the tasks no matter if the event loop is already running or not.
- It allowed me to cover the module with tests .

Unfortunately, I could not write a proper test case for the scenario when the event loop is _**not**_ running as stopping or mocking the event loop was not possible but, I was able to test it manually and in fact the tasks get executed.

Thoughts?

MarkLark86: From within the Celery app, does the task get executed with run_until_complete?

From memory, we don't actually use the response of the task from the Web API, it's more a set and forget (send task request, and move on without getting the result).

Tests is one thing, and we can probably just use mocking to provide unit tests.

But in the Celery app, we should use run til complete, otherwise celery might start processing the next task request before the current one is complete.

The main goal with supporting async in celery tasks execution, is about reusing the same Async code that the Web API uses (otherwise we would need sync code for resource CRUD operations, in the new core)

eos87: > From within the Celery app, does the task get executed with run_until_complete?

As long as `CELERY_TASK_ALWAYS_EAGER` is not enabled (False), yes, it does. Although it is not really clear to me how celery workers handle the event loop. From testing, I can tell that the loop is not running and always uses `run_until_complete`.

> From memory, we don't actually use the response of the task from the Web API, it's more a set and forget (send task request, and move on without getting the result).

Indeed. I think we don't even care about enabling tasks results if I'm not mistaken (no `CELERY_RESULT_BACKEND`)

> But in the Celery app, we should use run til complete, otherwise celery might start processing the next task request before the current one is complete.

This depends on how we're running our celery workers. I think we're using the default worker model (Prefork) which as I understand it only handles one task at a time. I think this is a risk if we were using Gevent or Eventlet, where tasks could overlap as it is supposed to run tasks concurrently. Sticking only with `run_until_complete` would still worry me as if for some reason the loop is already running, we still need to schedule the async task for execution. 

> The main goal with supporting async in celery tasks execution, is about reusing the same Async code that the Web API uses (otherwise we would need sync code for resource CRUD operations, in the new core)

Right.

superdesk/celery_app/context_task.py
eos87: I added a TODO comment here so we don't forget :)

MarkLark86 (approved):

PR #2662, opened 2024-08-16
update get_user_notification_preferences function [SDESK-7366]

devketanpro:


petrjasek (commented):
can you add a test case? not sure what exactly it's fixing

devketanpro:
previously, the function wasn’t searching for values inside the notifications field; it was always checking `user["user_preferences"]["notification_name"]`. I’ve now updated the function so that if a notification name is provided, it will check `user["user_preferences"]["notifications"]["notification_name"]...` instead.

tests/preferences/preference_tests.py
```diff
@@ -161,16 +161,18 @@ def test_get_user_notification_preferences_with_email_prefix(self):
             username="foo",
             user_preferences={
                 "email:notification": {"enabled": True},
-                "email:notification:test": {"enabled": False},
+                "notifications": {"email:notification:test": {"email": False, "desktop": True}},
```
petrjasek, line 164: that's not how it should be working, having `email` prefixed notification inside `notifications`

devketanpro: currently, this is how it works

devketanpro: currently we store it in this way :- `"notifications": {"email:notification:assignments": {"email": False, "desktop": True}},` right @thecalcc ?

thecalcc: we store it with the notification id and then `email: boolean` & `desktop: boolean`

petrjasek: atm we don't store it inside `notifications` right. once we do we shouldn't have there `email:notification` in the name. I only had there this code with `email:notification` prefix to be backwards compatible, but it shouldn't be used in future and definitely not inside `notifications` dict

devketanpro:
don;t need to change anything in the core

PR #2665, opened 2024-08-19
fix validation error when saving preferences

petrjasek:


thecalcc (approved):
Maybe remove the @deprecated code as well

MarkLark86:
`guess we'll need to keep those forever since all clients have those notifications`
@petrjasek I'm not a fan of that idea. We can keep it as is for now to get it working, but might be good to have another ticket that addresses this with a migration script

petrjasek:
yep, eventually we should refactor how the preferences are stored because currently we put there everything including the defaults, it should be just key: value pairs

PR #2668, opened 2024-08-20
[SDESK-7384] fix: Store ObjectId in MongoDB from async resource services

MarkLark86:


eos87 (approved):
Tested it locally and works as expected. Thanks!

PR #2669, opened 2024-08-21
[SDESK-7323] Implement etag concurrency with async rest endpoints

MarkLark86:


eos87 (approved):
Looks good. Tested it locally with Newshub.

PR #2671, opened 2024-08-23
SDAAP-112 Stop the generation of html on the server throwing an exception

marwoodandrew:
The front end sends draftjs state blocks of type section, article etc. This causes the server to throw an exception when generating the html. 

To replicate the issue paste the html from the output panel on the page https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section the auto save will fail, also just saving the article.

Maybe a fix should be implemented on the front end, but this will at least stop the grief!

I think this issue exists in the current develop version as well.

@petrjasek @MarkLark86  I'm not sure if bundling the unhandled elements into a p tag is best? would a div be better?

### DIFF
--- a/superdesk/editor_utils.py
+++ b/superdesk/editor_utils.py
@@ -313,6 +313,13 @@ class DraftJSHTMLExporter:
                     TABLE: self.render_table,
                     IMAGE: self.render_image,
                 },
+                "block_map": dict(
+                    BLOCK_MAP,
+                    **{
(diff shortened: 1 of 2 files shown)


petrjasek:
I think falling back to `div` would be better than using `p`, `section` or `articles` are wrappers around other elements so there might be `p`s inside

petrjasek (approved):

PR #2678, opened 2024-08-30
[SDESK-7275] Support versioned async resources

MarkLark86:


eos87 (approved):
It looks good to me, though I feel I may not have the full context or understanding needed for a thorough review. It might be helpful to have another set of eyes on this. @petrjasek, could you take a quick look? Thanks!

superdesk/core/mongo.py
```diff
@@ -240,104 +246,138 @@ def stop(self):
...
+        indexes = mongo_config.indexes or []
+        if indexes:
+            self.create_collection_indexes(
+                self.get_collection(resource_name, versioning=False),
+                indexes,
+                ignore_duplicate_keys,
+            )
+
+        if mongo_config.versioning:
+            indexes += (mongo_config.version_indexes or []) + [
```
petrjasek, line 321: I'm not sure if we need to add any indexes from the main resource here, imo that `_id_document` should do in theory

superdesk/resource_fields.py
```diff
@@ -22,5 +23,6 @@
 ITEMS = "_items"
 LINKS = "_links"
 ETAG = "_etag"
-VERSION = "_current_version"
+VERSION = CURRENT_VERSION = "_current_version"
+LATEST_VERSION = "_latest_version"
```
petrjasek, line 27: I wonder what's the usecase for this field, seems like something that would require an update to all previous versions on each save, but at the end it's just duplicating the `_current_version` value from the main collection

petrjasek (commented):
would it register that extra `{resource}_version` resource as it does currently? I see it in a way useful that you can use it as any other resource, but it might not be actually what we should allow, maybe we can have some specific base class for those which would be readonly via the api

MarkLark86:
> would it register that extra `{resource}_version` resource as it does currently?

No it doesn't. Instead it provides an `version: int | None` argument to the [find_one](https://github.com/superdesk/superdesk-core/pull/2678/files#diff-7afef0e188cec0364bc5d9e62317fe6639b375a9889f923817c3ba2aed3777feR118) method on the AsyncResourceService, along with a `get_all_item_versions` and `get-item_version` methods. It is also handled by requests through the REST API via the same `version` URL argument we use now.

> maybe we can have some specific base class for those which would be readonly via the api

The AsyncResourceService doesn't publicly expose any interface to create, modify nor delete a version document. It is internally handled in the create, update delete methods on that service.

superdesk/core/mongo.py
MarkLark86: I just copied what Eve did. We should at a minimum have indexes for `_id_document_` and `_current_version`. I'll make this change tomorrow

superdesk/resource_fields.py
MarkLark86: It's populated on fetch (so no need to update previous items), and is also used in the front-end.

petrjasek: when fetching previous versions?

superdesk/core/mongo.py
petrjasek: yep, we need `_id_document` for sure, I'm just not sure about the other ones

superdesk/resource_fields.py
MarkLark86: Correct

superdesk/core/mongo.py
```diff
@@ -240,104 +247,138 @@ def stop(self):
...
+
+        if mongo_config.versioning:
+            indexes = (mongo_config.version_indexes or []) + [
+                MongoIndexOptions(
+                    name="_id_document_1",
+                    keys=[(VERSION_ID_FIELD, 1)],
+                    background=True,
+                    unique=False,
+                ),
+                MongoIndexOptions(
```
petrjasek, line 329: imo this one might be enough, that should be usable also for lookups using just `_id_document`

petrjasek (approved):

PR #2679, opened 2024-08-30
[NHUB-535] improve: Support module config in MODULES settings.py

MarkLark86:


superdesk/core/config.py
```diff
@@ -53,12 +53,15 @@ def create_from_dict(cls, config: Dict[str, Any], prefix: Optional[str] = None,
         config_instance._loaded = True
         return config_instance
 
-    def load_from_dict(self, config: Dict[str, Any], prefix: Optional[str] = None, freeze: bool = True):
+    def load_from_dict(
+        self, config: Dict[str, Any], prefix: Optional[str] = None, freeze: bool = True, additional: dict | None = None
```
eos87, line 57: nit:
```suggestion
        self, config: Dict[str, Any], prefix: Optional[str] = None, freeze: bool = True, additional: Optional[Dict] = None
```

eos87 (approved):
Looks good. I left a very minor suggestion.

superdesk/core/config.py
MarkLark86: Based on the [docs](https://docs.python.org/3/library/typing.html#typing.Optional): `Optional[X] is equivalent to X | None (or Union[X, None])`.

IMO we should use type definitions from the language without import from the typing module, where applicable.

PR #2680, opened 2024-08-30
[SDESK-7284] Fix async commands to build a Superdesk instance

eos87:
### Purpose
Migrate the following commands into using `Click` library
- [x] app:initialize_data
- [x] users:create
- [x] data:upgrade
- [x] data:downgrade
- [x] schema:migrate

### What has changed
- Introduce `async_to_sync` from `asgiref.sync` in order to run commands with async code from start to end in the terminal without having event loop issues
- Implemented a new async_cli module https://github.com/superdesk/superdesk-core/pull/2680/commits/89780b367f4a43b562597bafe135fb4649271ce5 to register a custom Blueprint and AppGroup so that we can easily wrap the commands with `async_to_sync` and add the to Quart's cli at the same time
- Rework the commands so they are not function based and using Click lib

Thanks for checking!

**_Notes:_** 
- This is most likely going to break Newshub but I guess we'll have solve that in a separate ticket.
- I'd recommend hiding the white space changes to reduce the noise at the moment to reviewing the code :)

### DIFF
--- a/superdesk/commands/async_cli.py
+++ b/superdesk/commands/async_cli.py
@@ -1,0 +1,71 @@
+from typing import Tuple
+from functools import wraps
+from asgiref.sync import async_to_sync
+
+from ..flask import Blueprint, AppGroup
+
+
+class AsyncAppGroup(AppGroup):
+    """
+    An extension of Quart's AppGroup to support registration of asynchronous command handlers.
+
+    This class provides a mechanism to register asynchronous functions as command line commands,
+    which are automatically handled synchronously using `asgiref.sync.async_to_sync` conversion.
+    """
+
+    def register_async_command(self, name=None, **click_kwargs):
+        """
+        A decorator to register an asynchronous command within the Quart CLI app group.
+
+        Args:
+            name (str): The name of the command. If None, the function's name will be used.
+            **click_kwargs: Arbitrary keyword arguments that are passed to the `AppGroup.command` decorator.
+
+        Returns:
+            function: A decorator that wraps the async function and registers it as a command.
+        """
+
+        def decorator(func):
+            @wraps(func)
+            # This wrapper will convert the async function to a sync function
+            # using async_to_sync when the command is actually called.
+            def sync_wrapper(*args, **kwargs):
+                return async_to_sync(func)(*args, **kwargs)
+
+            # register the sync wrapper as a click command
+            click_command = self.command(name, **click_kwargs)
+            return click_command(sync_wrapper)
+
+        return decorator
+
+
+class CommandsBlueprint(Blueprint):
+    """
+    Custom Blueprint that integrates AsyncAppGroup to register CLI commands that are asynchronous,
+    allowing them to be used in a synchronous context by Quart's command line interface.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.cli = AsyncAppGroup()
+
+
+def create_commands_blueprint(blueprint_name: str) -> Tuple[CommandsBlueprint, AsyncAppGroup]:
+    """
+    Create a Blueprint to organize all superdesk commands.
+
+    By setting cli_group=None, any new CLI commands added to this blueprint
+    will still be compatible with the existing `python manage.py <command-name>`
+    format.
+
+    Returns:
+        Tuple[Blueprint, AppGroup]: A tuple containing the configured Blueprint
+        object and its associated CLI AppGroup for command registration.
+    """
+    blueprint = CommandsBlueprint(blueprint_name, __name__)
+
+    return blueprint, blueprint.cli
+
+
+commands_blueprint, cli = create_commands_blueprint("superdesk")
--- a/superdesk/commands/async_cli.py
+++ b/superdesk/commands/async_cli.py
@@ -1,4 +1,4 @@
-from typing import Tuple
+from typing import Tuple, cast
 from functools import wraps
 from asgiref.sync import async_to_sync
 
@@ -65,7 +65,7 @@ def create_commands_blueprint(blueprint_name: str) -> Tuple[CommandsBlueprint, A
     """
     blueprint = CommandsBlueprint(blueprint_name, __name__)
 
-    return blueprint, blueprint.cli
+    return blueprint, cast(AsyncAppGroup, blueprint.cli)
 
 
 commands_blueprint, cli = create_commands_blueprint("superdesk")
--- a/superdesk/commands/async_cli.py
+++ b/superdesk/commands/async_cli.py
@@ -1,8 +1,8 @@
-from typing import Tuple, cast
 from functools import wraps
+from typing import Any, Callable, Optional, Tuple, cast
 from asgiref.sync import async_to_sync
 
-from ..flask import Blueprint, AppGroup
+from ..flask import Blueprint, AppGroup, Flask
 
 
 class AsyncAppGroup(AppGroup):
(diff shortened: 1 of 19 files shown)


superdesk/commands/async_cli.py
```diff
@@ -0,0 +1,71 @@
...
+
+    By setting cli_group=None, any new CLI commands added to this blueprint
+    will still be compatible with the existing `python manage.py <command-name>`
+    format.
+
+    Returns:
+        Tuple[Blueprint, AppGroup]: A tuple containing the configured Blueprint
+        object and its associated CLI AppGroup for command registration.
+    """
+    blueprint = CommandsBlueprint(blueprint_name, __name__)
```
MarkLark86, line 66: If you add the `cli_group=None` argument here, the command name would not require `superdesk ` before each command name. Otherwise we would have to change deployment scripts etc.

MarkLark86 (commented):
I tried `python manage.py app:initialize_data` locally and I get the following:
`Usage: manage.py [OPTIONS] COMMAND [ARGS]...
Try 'manage.py --help' for help.

Error: No such command 'users:create'.

When I add `app.register_blueprint(commands_blueprint)` in the [init_app](https://github.com/superdesk/superdesk-core/blob/2f3fe119ab0516c9b10b324b24570e75de235569/superdesk/commands/__init__.py#L25) I am able to see commands now.
```
Commands:
  data:downgrade  Runs the latest data update backward.
  data:upgrade    Runs all the new data updates available.
  routes          Show the routes for the app.
  run             Run a development server.
  schema:migrate  Migrate elastic schema if needed, should be triggered...
  shell           Run a shell in the app context.
  users:create    Create a user with given username, password and email.
```

I cannot see the `app:initialize_data` command there, and when trying to run `python manage.py users:create -u admin -p admin -e admin@example.com --admin` I get the following error:
```
  File "async/superdesk/server/manage.py", line 17, in <module>
    main()
  File "python3.10/site-packages/quart/cli.py", line 754, in main
    cli.main()
  File "python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "async/superdesk-core/superdesk/commands/async_cli.py", line 33, in sync_wrapper
    return async_to_sync(func)(*args, **kwargs)
  File "python3.10/site-packages/asgiref/sync.py", line 254, in __call__
    return call_result.result()
  File "python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "python3.10/site-packages/asgiref/sync.py", line 331, in main_wrap
    result = await self.awaitable(*args, **kwargs)
  File "async/superdesk-core/apps/auth/db/commands.py", line 47, in create_user_command
    return await create_user_command_handler(*args, **kwargs)
  File "async/superdesk-core/apps/auth/db/commands.py", line 62, in create_user_command_handler
    app = get_current_app().as_any()
  File "python3.10/site-packages/werkzeug/local.py", line 318, in __get__
    obj = instance._get_current_object()
  File "python3.10/site-packages/werkzeug/local.py", line 519, in _get_current_object
    raise RuntimeError(unbound_message) from None
RuntimeError: Not within an app context
```

MarkLark86:
@eos87 I'm wondering if it would be good to add to the github actions, a set of commands we use in fireq to setup an instance. In order, these are:

* python manage.py app:initialize_data
* python manage.py users:create -u admin -p admin -e admin@example.com --admin
* python manage.py data:upgrade
* python manage.py schema:migrate

Having them there would replicate building an environment ready for use

superdesk/commands/async_cli.py
eos87: yeah, I removed it for testing purposes and then forgot to add it back. Fixed

eos87:
> I tried `python manage.py app:initialize_data` locally and I get the following: `Usage: manage.py [OPTIONS] COMMAND [ARGS]... Try 'manage.py --help' for help.
> 
> Error: No such command 'users:create'.
> 
> When I add `app.register_blueprint(commands_blueprint)` in the [init_app](https://github.com/superdesk/superdesk-core/blob/2f3fe119ab0516c9b10b324b24570e75de235569/superdesk/commands/__init__.py#L25) I am able to see commands now.
> 
> ```
> Commands:
>   data:downgrade  Runs the latest data update backward.
>   data:upgrade    Runs all the new data updates available.
>   routes          Show the routes for the app.
>   run             Run a development server.
>   schema:migrate  Migrate elastic schema if needed, should be triggered...
>   shell           Run a shell in the app context.
>   users:create    Create a user with given username, password and email.
> ```
> 
> I cannot see the `app:initialize_data` command there, and when trying to run `python manage.py users:create -u admin -p admin -e admin@example.com --admin` I get the following error:
> 
> ```
> Traceback (most recent call last):
>   File "async/superdesk/server/manage.py", line 17, in <module>
>     main()
>   File "python3.10/site-packages/quart/cli.py", line 754, in main
>     cli.main()
>   File "python3.10/site-packages/click/core.py", line 1078, in main
>     rv = self.invoke(ctx)
>   File "python3.10/site-packages/click/core.py", line 1688, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "python3.10/site-packages/click/core.py", line 1434, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>   File "python3.10/site-packages/click/core.py", line 783, in invoke
>     return __callback(*args, **kwargs)
>   File "async/superdesk-core/superdesk/commands/async_cli.py", line 33, in sync_wrapper
>     return async_to_sync(func)(*args, **kwargs)
>   File "python3.10/site-packages/asgiref/sync.py", line 254, in __call__
>     return call_result.result()
>   File "python3.10/concurrent/futures/_base.py", line 451, in result
>     return self.__get_result()
>   File "python3.10/concurrent/futures/_base.py", line 403, in __get_result
>     raise self._exception
>   File "python3.10/site-packages/asgiref/sync.py", line 331, in main_wrap
>     result = await self.awaitable(*args, **kwargs)
>   File "async/superdesk-core/apps/auth/db/commands.py", line 47, in create_user_command
>     return await create_user_command_handler(*args, **kwargs)
>   File "async/superdesk-core/apps/auth/db/commands.py", line 62, in create_user_command_handler
>     app = get_current_app().as_any()
>   File "python3.10/site-packages/werkzeug/local.py", line 318, in __get__
>     obj = instance._get_current_object()
>   File "python3.10/site-packages/werkzeug/local.py", line 519, in _get_current_object
>     raise RuntimeError(unbound_message) from None
> RuntimeError: Not within an app context
> ```

Hi @MarkLark86, I've fixed this. Please take another look. The most important thing is in here https://github.com/superdesk/superdesk-core/tree/ffb14fafa06a6700138b1c846017b88f18ae24f0

In order for the commands to run properly, you need to update the `manage.py` with something like 
```python
from quart.cli import main

if __name__ == "__main__":
    main()
```

but either an application, or create_app method should be reachable by quart in app.py file in the same directory. I opened this small PR for reference https://github.com/superdesk/superdesk/pull/3443

eos87:
> @eos87 I'm wondering if it would be good to add to the github actions, a set of commands we use in fireq to setup an instance. In order, these are:
> 
> * python manage.py app:initialize_data
> * python manage.py users:create -u admin -p admin -e [admin@example.com](mailto:admin@example.com) --admin
> * python manage.py data:upgrade
> * python manage.py schema:migrate
> 
> Having them there would replicate building an environment ready for use

Regarding this, yes I agree but, I'd rather tackle it in a separate ticket. `app:initialize_data` and `users:create` work as expected but the ones of data migration need more adjustments as the migration files as broken because PyMongo has removed the `update` method. We need to check those carefully and update it to the new pymongo API.

MarkLark86:
Hey @eos87 I tested this locally, and the only command that didn't work was `python manage.py data:upgrade`. If we can maybe place a dummy command in for that one for now, and comment out the existing implementation. That way we could possibly have a superdesk async test instance running after this

MarkLark86:
Good news is, I ran every other command in sequence and now have a running async superdesk locally

eos87:
> Hey @eos87 I tested this locally, and the only command that didn't work was `python manage.py data:upgrade`. If we can maybe place a dummy command in for that one for now, and comment out the existing implementation. That way we could possibly have a superdesk async test instance running after this

That sounds good. I'll prepare the GitHub action then :)

eos87:
hey @MarkLark86, I made one small change that I think makes things better and easier. Take a look when you have some time https://github.com/superdesk/superdesk-core/pull/2680/commits/842cd1223ddb5f9835859724514b8c7974e783a1
Thanks!

superdesk/commands/async_cli.py
```diff
@@ -0,0 +1,107 @@
...
+                    # your command logic here
+        """
+
+        def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
+            @wraps(func)
+            # This wrapper will convert the async function to a sync function
+            # using async_to_sync when the command is actually called.
+            def sync_wrapper(*args, **kwargs) -> Any:
+                if with_appcontext:
+                    return async_to_sync(self.run_with_appcontext)(func, *args, **kwargs)
```
MarkLark86, line 53: We'll see how it goes using `async_to_sync`. My experience in the past, when we were on Flask3, is that this doesn't play nicely with PyMongo / Elasticsearch clients. Issues such as the connections to the DBs having a different event loop associated then the one currently being executed (assuming because it's not thread-safe, and not using the same context).

MarkLark86 (approved):

superdesk/commands/async_cli.py
eos87: Yeah, that was one of my concerns with the run_async_to_sync that I added on newsroom-core. I'm hoping that wrapping the command in the app context before executing it with async_to_sync would be enough to prevent some of those issues. My previous approach was instead, passing the app instance object directly to the command, and if the command needs the app then use that one instead of getting it from the `get_current_app` function, but still there is the risk that the connections are generated in a separate event loop before running the command. Anyways, we'll see how it goes.

PR #2684, opened 2024-09-03
[SDESK-7284] Launch superdesk test instance on Github Actions

eos87:
[Test Instance](https://github.com/superdesk/superdesk-core/actions/runs/10683439872/job/29611512169)

### Purpose
- Sets up a basic Superdesk server service using GitHub actions. 
- The basic configuration that has been taken from https://github.com/superdesk/superdesk repo. 
- The corresponding GitHub action workflow file has been added too. 
- Basic commands are run:
```bash
python manage.py app:initialize_data
python manage.py users:create -u admin -p admin -e admin@example.com --admin
python manage.py schema:migrate
# python manage.py data:upgrade 
```

`data:upgrade` is skipped until we know what to do with data migrations. 

This PR goes against `hg/SDESK-7284-commands-async`. 

Thanks for checking!

### DIFF
--- a/tests/test_instance/server/settings.py
+++ b/tests/test_instance/server/settings.py
@@ -1,0 +1,215 @@
+#!/usr/bin/env python
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013 to present Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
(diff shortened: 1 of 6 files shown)


tests/test_instance/server/settings.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python
```
MarkLark86, line 1: Probably don't need all these settings, might be good to stick with the default config

eos87: you're right. Removed.

MarkLark86 (approved):

PR #2685, opened 2024-09-04
[SDESK-7373] Support default sort config in async resources

MarkLark86:
Also made the following improvements:
* add `to_list` to ResourceCursorAsync
* allow kwargs to AsyncResourceService.find (so we execute find with `.find({}, sort=[("first_name", 1)])` without needing to import SearchRequest)
* support web endpoint with empty args, so we can use `def my_endpoint() -> Response: ...` when we don't need to request object.


MarkLark86:
The 2 `@overloads` for [service.find](https://github.com/superdesk/superdesk-core/pull/2685/files#diff-7afef0e188cec0364bc5d9e62317fe6639b375a9889f923817c3ba2aed3777feR419) method are documented here https://superdesk--2685.org.readthedocs.build/en/2685/core/resource_management.html#superdesk.core.resources.service.AsyncResourceService.find. This works with Mypy and the IDE to provide type errors, depending on the 1st argument to `.find`

superdesk/core/elastic/base_client.py
```diff
@@ -8,14 +8,14 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-from typing import Dict, Any, Optional, Iterator, List, Tuple, Union, TypedDict
+from typing import Dict, Any, Optional, Iterator, List, Tuple, Union, TypedDict, Literal
```
eos87, line 11: Is `Literal` needed here?

eos87 (approved):
Good improvement 👏!

PR #2688, opened 2024-09-06
[NHUB-534] Implement async cacheable service

eos87:
### Purpose
This service is an implementation based on `superdesk.services.CacheableService` although it is on top of our new async framework. 

### What has changed
- Implemented new service
- Added tests for it (old one lacked tests)
- Adjusted `SuperdeskCacheBackend` to avoid "App context not available" when running async coroutines with `@cache` decorator. The adjustment seems to do the trick but I'm not fully sure if it would work it all scenarios. We need to revisit it in order to determine if this fix is fine or if we should take another approach.

Thanks for checking!


superdesk/core/resources/service.py
```diff
@@ -564,4 +567,58 @@ def filter_ignore_fields(d, fields):
...
+        """
+        return getattr(g, self.cache_key, None)
+
+    def set_cache(self, value: Any) -> None:
+        """
+        Set the cached value in Flask's `g` object.
+        """
+        setattr(g, self.cache_key, value)
+
+    async def get_cached(self) -> List[Dict[str, Any]]:
```
MarkLark86, line 597: It would be good to document the methods of this class, and at least add API reference to the [docs](https://github.com/superdesk/superdesk-core/blob/async/docs/core/resource_management.rst?plain=1#L59)

MarkLark86 (approved):
Looks good to me

superdesk/core/resources/service.py
eos87: done https://superdesk--2688.org.readthedocs.build/en/2688/

PR #2693, opened 2024-09-10
Add minor improvements to core framework

eos87:
### What has changed
- Set default values to `status_code` and `headers` members in `Response` class
- Added two util methods (`to_dict` & `to_json`) to `ResourceModel` to avoid repeating `model_dump(by_alias=True, exclude_unset=True)`
- Fix minor `black` issue in `overload` definitions (latest version of black moves the `...` to same line and flake8 complains)

### DIFF
--- a/superdesk/core/resources/model.py
+++ b/superdesk/core/resources/model.py
@@ -110,25 +110,31 @@ class ResourceModel(BaseModel):
             obj["_type"] = item_type
         return instance
 
-    def to_dict(self) -> dict[str, Any]:
+    def to_dict(self, **kwargs) -> dict[str, Any]:
         """
         Convert the model instance to a dictionary representation with non-JSON-serializable Python objects.
 
+        :param kwargs: Optional keyword arguments to override the default parameters of model_dump.
         :rtype: dict[str, Any]
         :returns: A dictionary representation of the model instance with field aliases.
                 Only fields that are set (non-default) will be included.
         """
-        return self.model_dump(by_alias=True, exclude_unset=True)
+        default_params = {"by_alias": True, "exclude_unset": True}
+        default_params.update(kwargs)
+        return self.model_dump(**default_params)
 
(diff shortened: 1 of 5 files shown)


superdesk/core/resources/model.py
```diff
@@ -110,6 +110,26 @@ def model_validate(
             obj["_type"] = item_type
         return instance
 
+    def to_dict(self) -> dict[str, Any]:
```
MarkLark86, line 113: Might be good to allow providing options, such as [use_objectid](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/service.py#L257). This would allow us to define supported context attributes.

If any options are defined, then the `context` kwarg would be provided to the `model_dump` call.

eos87: Done!

superdesk/core/resources/service.py
```diff
@@ -237,19 +237,23 @@ async def validate_update(
...
+        :raises Pydantic.ValidationError: If any of the docs provided are not valid
         """
 
-        await self.on_create(docs)
         ids: List[str] = []
+
         for doc in docs:
+            if isinstance(doc, dict):
+                doc = self.get_model_instance_from_dict(doc)
+            await self.on_create([doc])
```
MarkLark86, line 256: With this change the `on_create` callback will only ever receive 1 item, therefor it doesn't make sense to have it's signature accept a list.

Rather than changing the signature of both `on_create` and `on_created` callbacks to receive 1 item (which would require changes in Newshub where these are used), we could instead convert the list of dicts before iterating over it.

I don't see this being much of a performance issue, as we rarely send more than 1 item to be created.

eos87: I adjusted it. I had to do some adjustments to the types so mypy won't complain. Please take another look.

MarkLark86 (approved):

PR #2695, opened 2024-09-12
Add cache parameter to does_match for selective caching control [SDBELGA-538]

devketanpro:
Issue:- 
In the content_filter service, caching results across different contexts (e.g., internal destinations and macros) can cause stale data to be reused when the same content filter is applied to different versions of an item. This leads to incorrect results, as the cache does not account for context-specific changes in the article data

### DIFF
--- a/apps/content_filters/content_filter/content_filter_service.py
+++ b/apps/content_filters/content_filter/content_filter_service.py
@@ -197,7 +197,9 @@ class ContentFilterService(CacheableService):
         if not content_filter:
             return True  # a non-existing filter matches every thing
         cache_id = _cache_id("filter-match", content_filter.get("_id") or content_filter.get("name"), article)
-        if not cache or not hasattr(flask.g, cache_id):
+        if not cache:
+            return self._does_match(content_filter, article, filters, cache)
+        elif not hasattr(flask.g, cache_id):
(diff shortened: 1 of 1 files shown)


apps/content_filters/content_filter/content_filter_service.py
```diff
@@ -193,34 +193,34 @@ def _get_elastic_query(self, doc, matching=True):
...
 
-    def does_match(self, content_filter, article, filters=None):
+    def does_match(self, content_filter, article, filters=None, cache=True):
         if not content_filter:
             return True  # a non-existing filter matches every thing
         cache_id = _cache_id("filter-match", content_filter.get("_id") or content_filter.get("name"), article)
-        if not hasattr(flask.g, cache_id):
-            setattr(flask.g, cache_id, self._does_match(content_filter, article, filters))
+        if not cache or not hasattr(flask.g, cache_id):
+            setattr(flask.g, cache_id, self._does_match(content_filter, article, filters, cache))
```
petrjasek, line 201: imo when `cache=False` it shouldn't set it neither, so rather do something like:
```py
if not cache:
    return self._does_match(content_filter, article, filters, cache)
elif not hasattr(...):
....
```

petrjasek: and that should be applied everywhere

petrjasek (approved):

PR #2697, opened 2024-09-13
Add cache parameter to does_match for selective caching control [SDBELGA-538] (#2695)

devketanpro:
* Add cache parameter to does_match for selective caching control [SDBELGA-538]

* refactore logic

[SDBELGA-538]: https://sofab.atlassian.net/browse/SDBELGA-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/content_filters/content_filter/content_filter_service.py
+++ b/apps/content_filters/content_filter/content_filter_service.py
@@ -193,34 +193,36 @@ class ContentFilterService(CacheableService):
             expressions_list.append({"bool": filter_conditions})
         return {"bool": expressions}
 
-    def does_match(self, content_filter, article, filters=None):
+    def does_match(self, content_filter, article, filters=None, cache=True):
         if not content_filter:
             return True  # a non-existing filter matches every thing
         cache_id = _cache_id("filter-match", content_filter.get("_id") or content_filter.get("name"), article)
-        if not hasattr(flask.g, cache_id):
-            setattr(flask.g, cache_id, self._does_match(content_filter, article, filters))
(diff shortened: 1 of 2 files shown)


apps/content_filters/content_filter/content_filter_service.py
```diff
@@ -230,18 +232,22 @@ def _does_filter_condition_match(self, content_filter, article, filters, express
...
+    def _does_content_filter_match(self, content_filter, article, filters, expression, cache=True) -> bool:
         for f in expression["expression"]["pf"]:
             cache_id = _cache_id("content-filter-match", f, article)
-            if not hasattr(flask.g, cache_id):
+            if not cache or not hasattr(flask.g, cache_id):
                 current_filter = (
                     filters.get("content_filters", {}).get(f, {}).get("cf") if filters else self.get_cached_by_id(f)
                 )
+                if not cache:
+                    return self.does_match(current_filter, article, filters=filters)
```
petrjasek, line 250: I think here you have to pass the `cache` param

petrjasek (commented):
would be good to cover that with some test

apps/content_filters/content_filter/content_filter_service.py
devketanpro: my bad I missed this

petrjasek (approved):

PR #2698, opened 2024-09-13
Hotfix/2.5.4

petrjasek:
backported celery update to make it installable again with recent pip,
not sure if we should merge that actually, it's more fyi @marwoodandrew if it might be useful to you


marwoodandrew (approved):

marwoodandrew:
Thanks @petrjasek

PR #2706, opened 2024-09-18
[SDESK-7379] Support projection in async resources

MarkLark86:
### Purpose
Add support to provide projection when requesting a resource, regardless if they're configured for Elastic and/or MongoDB.

Resolves: [SDESK-7379](https://sofab.atlassian.net/browse/SDESK-7379)

[SDESK-7379]: https://sofab.atlassian.net/browse/SDESK-7379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -56,7 +56,7 @@ Superdesk commands
 ``app:initialize_data``
 ^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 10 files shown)


eos87 (approved):
Looks good. I tested this locally with NH users to hide password and token. It works as expected. Only thing is, do you think we should have an example of usage in the documentation?

PR #2708, opened 2024-09-19
[SDESK-7381] fix: Return items from async post & patch REST requests

MarkLark86:
### Purpose
The response data from REST async resource endpoints are different than what Eve returns. This causes issues for code that expects the response to be the same as Eve.

### What has changed
Modify the ResourceRestEndpoints class to return the item from POST requests, the updates from PATCH requests and HATEOAS self link for both.

Resolves: [SDESK-7381](https://sofab.atlassian.net/browse/SDESK-7381)

[SDESK-7381]: https://sofab.atlassian.net/browse/SDESK-7381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


devketanpro (approved):
looks good !

PR #2715, opened 2024-09-24
Allow automatically setting `_id` attribute if not provided

eos87:
This will reduce the code duplication all over the app and dependent projects as we won't need to generate ids before creating registries. 

Blow an example of no longer needed code
```python
from bson import ObjectId

data = {
    "_id": ObjectId(),
   ...rest of data
}
```


MarkLark86:
@eos87 I've created a [PR](https://github.com/superdesk/superdesk-core/pull/2716) that implements this change, but via the ResourceModel directly. This way when any model is constructed, it will automatically generate an ID if one wasn't provided.

eos87:
Closing this in favor of #2716

PR #2716, opened 2024-09-25
improve(async): Auto-generate ID when constructing Model instances

MarkLark86:


superdesk/core/utils.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2024 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
```
eos87, line 10: We probably don't need to keep adding this as the License is in the repo and the link is broken anyways.

superdesk/core/resources/model.py
```diff
@@ -65,7 +66,7 @@ def type(self) -> str:
         return self.model_resource_name
 
     #: ID of the document
-    id: Annotated[Union[str, ObjectId], Field(alias="_id")]
+    id: Annotated[Union[str, ObjectId], Field(alias="_id", default_factory=lambda: generate_guid(type=GUID_NEWSML))]
```
eos87, line 69: Now I'm thinking we should also use `default_factory` for `created` and `updated` too. It would more clear that they are automatically set if not provided, therefore we don't need a service class do it for them. What do you think?

tests/core/resource_parent_links_test.py
```diff
@@ -93,12 +93,15 @@ async def test_patch_and_delete(self):
         folder_id = folder["_id"]
 
         # Update the users folder
+        from pprint import pprint
+
+        pprint(folder)
```
eos87, line 98: Intentional?

eos87 (approved):
Looks good to me. I left a couple of minor comments what do not affect the outcome of this PR.

superdesk/core/utils.py
MarkLark86: Good point. We can remove those all later from superdesk.core

tests/core/resource_parent_links_test.py
MarkLark86: Whoops

superdesk/core/resources/model.py
MarkLark86: I started to make this change. But, using `default_factory` here doesn't work because when you update the item, the `_updated` field already has a value and therefor does not get updated.

It also makes sense to set these values in the service, right before it is sent to the DB. So the `_created` and `_updated` fields are the date/time of when it was stored at rest (aka the DB)

PR #2722, opened 2024-10-01
[NHUB-546] Update FlushElasticIndex command 

eos87:
### What has changed
- This allows to flush elastic search indexes from registered async resources and reindexing them from mongo collections as well
- Added additional argument `index_prefix` to `drop_indexes` so when provided it would only remove the indices that contain such prefix 
- `analyze_wildcard` setting has been removed in recent versions of Elastic Search. There are some instances that use this setting and we need to address that in a separate ticket. In the meantime, we need this removed in order to be able to create indexes for async resources from `newsroom-core`. Some links with information below:
https://github.com/elastic/elasticsearch/pull/42706
https://github.com/elastic/elasticsearch/issues/20209
https://discuss.elastic.co/t/analyze-wildcard/5156 

The work in this PR is required by `newsroom-core` with ticket [NHUB-546]

[NHUB-546]: https://sofab.atlassian.net/browse/NHUB-546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eos87:
@MarkLark86 please ignore this for now. I realised that there some changes I would like to make before getting it reviewed.

MarkLark86:
@eos87 Please see https://github.com/superdesk/newsroom-core/pull/1104 for information about this `query_string.analyze_wildcard` setting. I found the culprit and fixed there.

eos87:
I think this is ready to be reviewed. @MarkLark86 please take a look when you get some chance. Thanks!

superdesk/default_settings.py
```diff
@@ -94,11 +94,6 @@ def local_to_utc_hour(hour):
 #:
 ELASTIC_DEFAULT_SIZE = 10
 
-#: allow wildcards in query string
-#:
-#: https://discuss.elastic.co/t/configuring-the-standard-tokenizer/8691/5
-ELASTIC_QUERY_STRING_ANALYZE_WILDCARD = False
```
MarkLark86, line 100: This will still be required, as it's used by search to fill in from a client search request (and not a creation of index & settings)

MarkLark86:
Looks good, just the one config change to be reverted.

superdesk/default_settings.py
eos87: Fixed.

PR #2723, opened 2024-10-02
improve(cli): Support async commands that wrap app context

MarkLark86:
### Purpose
Run cli commands in the same thread due to issues with Flask/Quart context in different threads/contexts.
I discovered an issue within Newshub when running a command, that raised a not within app context, and found this to be caused by running in a different thread/context.

### What has changed
1. Move core cli funcitonality from `superdesk.commands` to `superdesk.core.cli` module
2. Replaces `register_async_command` with standard `command` method
3. Defaults `with_appcontext` to `True`, as most commands will need it anyway
4. Implement a modified version of `quart.cli.with_appcontext` with our own, that allows cli functions to be async


eos87 (approved):
Nice!

PR #2728, opened 2024-10-09
support python 3.12

petrjasek:


MarkLark86 (approved):

MarkLark86:
@petrjasek The number of warnings in the pytest has me concerned.
Python v3.10:
`1239 passed, 53713 warnings`
Python v3.12:
`1239 passed, 12386573 warnings`

That's 53k vs 12m warnings, that's a huge increase.

petrjasek:
yep, looks like there is quite some core work right there :sweat_smile:

PR #2729, opened 2024-10-10
[SDESK-7372] New Auth mechanism

MarkLark86:
### Purpose
Implement a new Authentication and Authorization mechanism, for use with async resources and endpoints

### What has changed
* Move common superdesk.core types into superdesk.core.types, to reduce chance of cyclic imports
* Implement base Authentication mechanism
* Implement base Authorization rules
* Implement Superdesk specific Authentication mechanism (may require further work)
* Expose sync mongo interfaces from async resources, for scenario's where it would be easier to use sync resource
* Support custom error strings in field validators

Resolves: [SDESK-7372](https://sofab.atlassian.net/browse/SDESK-7372)

[SDESK-7372]: https://sofab.atlassian.net/browse/SDESK-7372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarkLark86:
This PR is missing the docs and tests. Rhe tests are covered by the Newshub PR that I will create, as it uses this new auth mechanism throughout it's tests.

I will add docs and tests for this core feature in a separate PR, just wanting to get this one in and reviewed.

superdesk/core/web/endpoints.py
```diff
@@ -0,0 +1,152 @@
...
+            url="<null>",
+            func=null_endpoint,
+        )
+
+
+NullEndpoint = NullEndpointClass()
+
+
+async def return_404() -> Response:
+    return Response("", 404)
```
eos87, line 78: do we need `async` here?

superdesk/core/web/endpoints.py
```diff
@@ -0,0 +1,152 @@
...
+                auth=auth,
+                parent=self,
+            )
+            self.endpoints.append(endpoint_func)
+            return endpoint_func
+
+        return fdec
+
+    async def __call__(self, args: dict[str, Any], params: dict[str, Any], request: Request):
+        return return_404()
```
eos87, line 130: if we need the async above, then we should have await here, right?
```suggestion
        return await return_404()
```

eos87 (approved):
Looks good to me, I just left a couple of minor comments.

superdesk/core/web/endpoints.py
MarkLark86: I've added support for non-async endpoint functions now.

PR #2740, opened 2024-10-28
[NHUB-540] Add async ResourceModel for ContentAPI items

MarkLark86:
### Purpose
Create async ResourceModel for ContentAPI items, and implement publish code to it as simple functions (for re-use with Newshub). Also small changes required by the Newshub Wire async upgrade task.

### What has changed
* Create async ResourceModel for ContentAPI items
* Create re-usable functions for publishing into the ContentAPI (for re-use with Newshub)
* Support `projection` argument to async `find_one_` based functions
* Catch more Elasticsearch search errors, due to invalid search request, and print the query as well
* Support nested `include_in_parent` elasticsearch config
* Create a new BaseModel from Pydantic, so we have common interface to all models, regardless of usage (resource model, view args etc).
* ElasticCursor: Support asycn next function, also reset elastic iterator index once iteration completed (so it can be iterated again, if needed)
* Support more elasticsearch mapping settings on fields/annotations
* Fix issue with async cacheable service, locking in threads was causing an issue (in tests only I believe). So provide fallback so the function will always work
* Add a TODO comment to validators, support `validate_iunique_value_async` without requiring to add `resource_name` and possibly `field_name`, making it easier to use. Will do this later
* Improve layout of the types module, moving out definitions to different files, making it easier to maintain
* Add `is_json_request` to Request instance
* Add GUID_FIELD to configured lookup fields (in superdesk.resource_fields)


superdesk/core/types/system.py
```diff
@@ -0,0 +1,98 @@
...
+
+    locators: Any
+
+    celery: Any
+
+    redis: Any
+
+    jinja_loader: Any
+
+    jinja_env: Any
```
eos87, line 59: are we going to provide types for these in the future?

eos87 (approved):
Looks good, just left two minor comments.

superdesk/core/types/system.py
MarkLark86: Ideally we would not use this at all, by replacing the required functionality into the new async app. For example, `data` will no longer be required as we have equivalent in the new async app.

PR #2748, opened 2024-11-06
[NHUB-537] Various core changes for Newshub Agenda async

MarkLark86:
### Purpose
Various changes needed by the Newshub Agenda async upgrade task.

### What has changed
* Fixed typo with `_inner_hits` when processing elasticsearch response
* Add support in elastic mapping for enum to keyword for nested types
* Raise `ElasticNotConfiguredForResource` exception instead of generic `KeyError` when Elastic not configured for a resource (allows to catch more errors by narrowing down the try...except statements in the resource service)
* Add elasticsearch field dynamic mapping helper function
* Add `find_by_ids` and `find_by_ids_raw` to async resource service
* When sending dicts to async service create method, the dict will be updated with the ID and Etag of the stored resource item (handy in tests)
* Support `set[str]` in search projection param
* Support `exclude_fields` in ESQuery, shorthand for `_source` field in elastic
* Add date helpers to `superdesk.core.utils`

### DIFF
--- a/content_api/items/model.py
+++ b/content_api/items/model.py
@@ -31,6 +31,7 @@ ContentAssociation = Annotated[
 class CVItem:
     qcode: fields.Keyword
     name: fields.Keyword
+    scheme: fields.Keyword | None = None
     schema: fields.Keyword | None = None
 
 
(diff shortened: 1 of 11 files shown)


eos87 (approved):
Looks good!

PR #2753, opened 2024-11-20
improve: Add get_service to ResourceModel, and allow any fields not defined

MarkLark86:
### Purpose
Improve usage of ResourceModel to make it easier for developers

### What has changed
* Implement our own Dataclass class, so we have common `from_dict`, `from_json`, `to_dict` and `to_json` at that level too
* Allow any field by default on the ResourceModel and Dataclass based data
* Default `ResourceModel.from_dict` to include unknown fields, if `model_config.extra != 'forbit'`
* Add `get_service` to `ResourceModel`, so we can import just the type and use it for CRUD operations (further reduces cyclic imports)

### DIFF
--- a/superdesk/core/resources/__init__.py
+++ b/superdesk/core/resources/__init__.py
@@ -17,6 +17,8 @@ from .model import (
     ModelWithVersions,
     ResourceConfig,
     dataclass,
+    Dataclass,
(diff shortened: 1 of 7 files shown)


eos87 (approved):
Looks good, nice improvements!

PR #2757, opened 2024-11-26
[PUBLISH]Cache `Formatter.format` result when suitable.

jerome-poisson:
In Enqueue Service, `Formatter.format` result is now cached when `subscriber` (e.g. for `publish_seq_num`) or `codes` is not used in the formatted item.

SDESK-7449


MarkLark86 (approved):
Looks good @jerome-poisson

jerome-poisson:
@MarkLark86 What about the failing tests? Are they expected to fail at this point? For a quick glance they don't look related to my changes.

MarkLark86:
> @MarkLark86 What about the failing tests? Are they expected to fail at this point? For a quick glance they don't look related to my changes.

Pytest is passing, I got that working last week. I'll be working on getting other tests to pass later.
So for now, yes it's fine

PR #2760, opened 2024-11-28
Improve add async signal system

MarkLark86:
### Purpose
Create a new signal system that supports both async callbacks as well as being strictly typed

### What has changed
* Add a new signal system, that supports both async callbacks and defining the function signature types
* Add tests
* Add docs (https://superdesk--2760.org.readthedocs.build/en/2760/core/signals.html)


eos87 (approved):
Nicely done!

PR #2761, opened 2024-11-29
fix unit tests

MarkLark86:
### Purpose
Previously most of the unit tests were failing

### What has changed:
* Deepcopy the app config when running the tests, so previous runs do not affect others
* Use our own `str_to_date` function instead of the one from eve
* Fix `CELERY_TASK_ALWAYS_EAGER` from a string to bool in test config
* Fix app context in a few tests

### DIFF
--- a/.github/workflows/nose-tests.yml
+++ b/.github/workflows/nose-tests.yml
@@ -20,9 +20,8 @@ jobs:
     - run: ./scripts/tests_setup
     - run: pip install -U pip wheel setuptools
     - run: pip install -r dev-requirements.txt
-    - run: ulimit -n 4096
-    - run: pytest --log-level=ERROR --disable-warnings tests/core
-    - run: pytest --log-level=ERROR --disable-warnings --ignore=tests/core
+    - run: pytest -p no:asyncio --log-level=ERROR --disable-warnings tests/core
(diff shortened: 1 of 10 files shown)


MarkLark86:
@eos87 not sure what's going on with the GitHub actions for pytest, it runs locally and passing within 10mins. I'll look into this later (I'm thinking memory/file descriptor leak with regards to db connections)

MarkLark86:
Hey @eos87 I finally found the culprit of unit tests failing...: 
https://github.com/superdesk/superdesk-core/blob/a39e6345e32e3a8375156424c00630736dda2ff7/superdesk/tests/__init__.py#L409

:cry: Took much longer than I would have liked, but I got there. I will work separately on another PR that will fix the remaining behave and unit tests (for ProdAPI). For now This PR is ready for review/merge

eos87:
> Hey @eos87 I finally found the culprit of unit tests failing...:
> 
> https://github.com/superdesk/superdesk-core/blob/a39e6345e32e3a8375156424c00630736dda2ff7/superdesk/tests/__init__.py#L409
> 
> 😢 Took much longer than I would have liked, but I got there. I will work separately on another PR that will fix the remaining behave and unit tests (for ProdAPI). For now This PR is ready for review/merge

😨 the kind of little things that scares me when it comes to debugging unexpected async issues. Well done, man!

eos87 (approved):

PR #2764, opened 2024-12-02
Improvements to unify the functionality from decorator & `Dataclass` in a single class

eos87:
### Purpose
The intention to avoid needing to have both a `Dataclass` and a `@dataclass` decorator

### What has changed
- Introduced `DataclassBase` for reusable utility methods (`to_dict`, `from_dict`, `ser_model`).
- Added `DataclassMeta` to handle automatic Pydantic dataclass wrapping and custom `Config` merging.
- Enabled per-class custom configuration via a `Config` dictionary in child classes.

### Example usage

```python
class RingBearer(Dataclass):
    name: str
    race: str

    # custom configuration that will merge/override default one
    Config = {
        "extra": "allow",
    }

frodo = RingBearer(name="Frodo Baggins", race="Hobbit")

# unexpected fields are allowed due to 'extra': 'allow' (allowed by default, this is just for tests)
frodo.unknown_item = "Mithril Coat" 
frodo.secret_mission = "Destroy the One Ring"
```

@MarkLark86 following up on our [conversation here](https://github.com/superdesk/superdesk-planning/pull/2131#discussion_r1863719653) what do you think? If you think it looks good I'd merge it and replace the usages of the decorator. Also extend the documentation a little bit to make it clear what is the intention.

Thanks for checking!

### DIFF
--- a/superdesk/core/resources/model.py
+++ b/superdesk/core/resources/model.py
@@ -107,17 +107,16 @@ class DataclassBase:
 class DataclassMeta(type):
     """Metaclass to enhance Dataclass functionality."""
 
-    def __new__(cls, name, bases, dct):
-        custom_config = dct.pop("Config", {})
+    def __new__(cls: type, name: str, bases: tuple[type, ...], attrs: dict[str, Any]) -> type:
+        custom_config = attrs.pop("Config", {})
 
         # merge with default configuration
         config = deepcopy(default_model_config)
         config.update(custom_config)
 
         # create the class and apply the Pydantic dataclass decorator
-        new_cls = super().__new__(cls, name, bases, dct)
-        new_cls = pydataclass(new_cls, config=config)
-        return new_cls
+        new_cls = super().__new__(cls, name, bases, attrs)  # type: ignore[misc]
+        return pydataclass(new_cls, config=config)
 
 
 class Dataclass(DataclassBase, metaclass=DataclassMeta):
--- a/superdesk/core/resources/model.py
+++ b/superdesk/core/resources/model.py
@@ -34,7 +34,7 @@ from pydantic import (
     SerializerFunctionWrapHandler,
     RootModel,
 )
-from pydantic_core import InitErrorDetails, PydanticCustomError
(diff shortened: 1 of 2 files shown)


MarkLark86:
@eos87 How does this work with the IDE predictions/suggestions when typing field names?

eos87:
> @eos87 How does this work with the IDE predictions/suggestions when typing field names?

It does. This is vscode below. I would assume it would also work on PyCharm and others. 
image

MarkLark86:
Testing this locally, and I found 2 things (1 of those is probably an issue with my original code).
* `from_dict` and `from_json` return an instance of `dict` and not an instance of the class
* mypy fails to recognise arguments to `__init__`.

```python
from superdesk.core.resources import Dataclass

class Person(Dataclass):
    name: str

assert type(Person(name="Monkeys")) == Person  # True
assert type(Person.from_dict(dict(name="Monkeys"))) == Person  # False
assert type(Person.from_dict(dict(name="Monkeys"))) == dict  # True
assert type(Person.from_json('{"name": "Monkeys"}')) == Person  # False
assert type(Person.from_json('{"name": "Monkeys"}')) == dict  # False

Person(name="Monkeys")  # mypy error: Unexpected keyword argument "name" for "Person"  [call-arg]
```

Might be good to add some tests with using this new datatype of ours

eos87:
> from_dict and from_json return an instance of dict and not an instance of the class

yeah, it was already a dict indeed. I'll take a look to return the proper model instance. 

> Person(name="Monkeys")  # mypy error: Unexpected keyword argument "name" for "Person"  [call-arg]

interesting, I've just tried this locally and it worked for me. Maybe we have a different version of mypy?

eos87:
Improvements added so editor will autocomplete the attributes inline when creating an instance of a class. Also added some tests to double check the functionality of the utility methods `from_dict`, `from_json`, `to_dict` and `to_json`.

MarkLark86 (approved):
Looks like a good improvement, thanks mate

PR #2797, opened 2024-12-26
New AFP Parser implementation [SDESK-7422]

devketanpro:


superdesk/io/feed_parsers/__init__.py
```diff
@@ -352,3 +352,4 @@ class EmailFeedParser(FeedParser, metaclass=ABCMeta):
 from superdesk.io.feed_parsers.ana_mpe_newsml import ANANewsMLOneFeedParser  # NOQA
 from superdesk.io.feed_parsers.bbc_ninjs import BBCNINJSFeedParser  # NOQA
 from superdesk.io.feed_parsers.ap_media import APMediaFeedParser  # NOQA
+from superdesk.io.feed_parsers.afp_newsml_2_0 import AFPNewsMLFeedParser  # NOQA
```
petrjasek, line 355: filename should be rather `newsml_1_2` it's not based on newsml 2.0 but 1.2

tests/io/feed_parsers/afpnewsml20_test.py
```diff
@@ -0,0 +1,57 @@
...
+    def setUp(self):
+        dirname = os.path.dirname(os.path.realpath(__file__))
+        fixture = os.path.join(dirname, "../fixtures", "afp20.xml")
+        provider = {"name": "Test"}
+        with open(fixture, "rb") as f:
+            self.item = AFPNewsMLFeedParser().parse(etree.fromstring(f.read()), provider)
+
+    def test_headline(self):
+        self.assertEqual(
+            self.item.get("headline"), "DÃ©cÃ¨s de Naomi Musenga: l'opÃ©ratrice du Samu dÃ©finitivement condamnÃ©e"
```
petrjasek, line 30: this looks like wrong encoding to me, should look like in the newsml

tests/io/feed_parsers/afpnewsml20_test.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_coreitemvalues(self):
+        self.assertEqual(self.item.get("type"), "text")
+        self.assertEqual(self.item.get("urgency"), None)
+        self.assertEqual(self.item.get("version"), "1")
+        self.assertEqual(self.item.get("versioncreated"), datetime.datetime(2024, 7, 10, 12, 59, 2, tzinfo=utc))
+        self.assertEqual(self.item.get("firstcreated"), datetime.datetime(2024, 7, 10, 12, 59, 2, tzinfo=utc))
+        self.assertEqual(self.item.get("pubstatus"), "usable")
+
+    def test_content_is_text(self):
+        self.assertIsInstance(self.item.get("body_html"), type(""))
```
petrjasek, line 57: would be good to check the actual text, I wonder if the closing tag is removed and if the closing tag is removed as well

tests/io/fixtures/afp20.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="iso-8859-1"?>
+<?xml version="1.0" encoding="UTF-8"?>
```
petrjasek, line 1: I guess they won't change that for us, we need to use what we get from them

devketanpro: then could you please give any idea how we handle it?

petrjasek: I guess the library will handle it, so it might be ok actually. to test it you could save the file in that encoding from the header, seems like atm it's encoded as utf8

tests/io/feed_parsers/afpnewsmlnew_test.py
```diff
@@ -0,0 +1,68 @@
...
+from superdesk.io.feed_parsers.afp_newsml_1_2_new import AFPNewsMLFeedParser
+
+
+class TestCase(unittest.TestCase):
+    def setUp(self):
+        dirname = os.path.dirname(os.path.realpath(__file__))
+        fixture = os.path.join(dirname, "../fixtures", "afp20.xml")
+        provider = {"name": "Test"}
+        with open(fixture, "rb") as f:
+            self.item = AFPNewsMLFeedParser().parse(etree.fromstring(f.read()), provider)
```
petrjasek, line 28: but maybe if you use mode `r` in open and then `f.read().encode("iso-8885-1")` it might work.
anyhow all the strings here must be readable, here it must be unicode

petrjasek (approved):

PR #2799, opened 2024-12-30
[SDESK-7464] - Create new async Users Resource and Service

BrianMwangi21:
### Purpose
This PR creates new async users resource and service:

- Implementing a new resource model for users
- Implementing new async resource service for users
- Implementing resource config for users and register it in users module

Solves: [SDESK-7464](https://sofab.atlassian.net/browse/SDESK-7464)

[SDESK-7464]: https://sofab.atlassian.net/browse/SDESK-7464


BrianMwangi21:
@MarkLark86 a few questions : 

1. For areas with `self.datasourse` in the previous service, do we have a similar attribute for new async resources/models ? 
2. For the `get` function in the new users async service, should it be a custom one or should it override the `find` or `search` in the AsyncResourceService ? 
3. Is there a current equivalent to `self._validator()` in the new AsyncResourceService ? Or is it to be accessed by a resource model like. It's used in the DBUsersService on line 603 : 

```
        validator = self._validator()  # TODO-ASYNC: Confirm on usage of validator
```

superdesk/types/users.py
```diff
@@ -0,0 +1,71 @@
...
+    job_title: str | None = None
+    biography: str | None = None
+    facebook: str | None = None
+    instagram: str | None = None
+    twitter: str | None = None
+    jid: Annotated[fields.Keyword, validate_unique_value_async("users", "jid")] | None = None
+    language: str | None = None
+    user_info: dict[str, Any] = Field(default_factory=dict)
+    picture_url: str | None = None
+    avatar: Annotated[ObjectId, validate_data_relation_async("uploads")] | None = None
```
MarkLark86, line 41: The old resource uses `upload` as the resource relation name

superdesk/types/users.py
```diff
@@ -0,0 +1,71 @@
...
+    avatar_renditions: dict[str, Any] | None = Field(default=None)
+    role: Annotated[ObjectId, validate_data_relation_async("roles")]
+    privileges: dict[str, Any] = Field(default_factory=dict)
+    workspace: dict[str, Any] = Field(default_factory=dict)
+    user_type: str = Field(default="user")
```
MarkLark86, line 46: This should be an enum, with possible values `["user", "administrator", "external"]`

superdesk/users/async_service.py
```diff
@@ -0,0 +1,619 @@
...
+                return get_resource_service("roles").find_one(_id=role_id, req=None)
+        return None
+
+    async def set_privileges(self, user: UsersResourceModel, role):
+        user_dict = user.to_dict()
+        user_dict["active_privileges"] = get_privileges(user, role)
+        user = UsersResourceModel(**user_dict)
+
+    # TODO-ASYNC: Confirm on change to `get` function
+    async def get(self, req, lookup) -> ResourceCursorAsync[UsersResourceModel]:
```
MarkLark86, line 405: This entire function can be skipped in this async service, as it is in the context of an API Request, so it belongs in the REST API service (which we will do at a later stage).

superdesk/users/async_service.py
```diff
@@ -0,0 +1,619 @@
...
+
+    async def on_delete(self, doc: UsersResourceModel) -> None:
+        """
+        Overriding clean up reset password tokens:
+        """
+
+        await super().on_deleted(doc)
+        get_resource_service("reset_user_password").remove_all_tokens_for_email(doc.to_dict().get("email"))
+
+    async def _process_external_data(self, _data: dict[str, Any], update=False):
```
MarkLark86, line 582: These next 3 aren't required for this PR (they're needed for auth code, when auto-creating users from an external system). Can be left out of this new service for now

MarkLark86:
> @MarkLark86 a few questions :
> 
> 1. For areas with `self.datasourse` in the previous service, do we have a similar attribute for new async resources/models ?
> 2. For the `get` function in the new users async service, should it be a custom one or should it override the `find` or `search` in the AsyncResourceService ?
> 3. Is there a current equivalent to `self._validator()` in the new AsyncResourceService ? Or is it to be accessed by a resource model like. It's used in the DBUsersService on line 603 :
> 
> ```
>         validator = self._validator()  # TODO-ASYNC: Confirm on usage of validator
> ```

1. `self.datasource` is the same as `self.resource_name` in the new async services
2. Don't worry about the `get` function, this will be done on the REST API later
3. Also don't worry about validator, as it's used by functions not required by this PR

superdesk/types/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .users import UsersResourceModel
```
MarkLark86, line 1: There is also a [superdesk.tests](https://github.com/superdesk/superdesk-core/blob/async/superdesk/types.py) file, which this module overrides causing an exception to be raised when running the app.

Could you move the content of that original file to this one, and delete the old one as well please

superdesk/users/async_service.py
```diff
@@ -0,0 +1,537 @@
...
+    if SIGN_OFF not in user or user[SIGN_OFF] is None:
+        set_sign_off(user)
+
+    return user[SIGN_OFF]
+
+
+class UsersAsyncService(AsyncResourceService[UsersResourceModel]):
+    _updating_stage_visibility = True
+
+    async def __is_invalid_operation(
```
MarkLark86, line 143: After further inspection, looks like this one is not needed here either, as it is in the context of an API request.

It's functions like these should not go into these new async resource services, but instead the REST API (when we implement the User one).

We want to keep the separation of concerns with the new code, by separating out the data and request layers specifically. This should help keep these two concerns smaller, easier to read etc

superdesk/users/async_service.py
```diff
@@ -0,0 +1,537 @@
...
+        resetService = get_resource_service("reset_user_password")
+        activate_ttl = get_app_config("ACTIVATE_ACCOUNT_TOKEN_TIME_TO_LIVE")
+        for doc in docs:
+            user_dict = doc.to_dict()
+            if self.user_is_waiting_activation(doc) and user_dict.get("user_type") != "external":
+                user_id = user_dict.get("_id")
+                email = user_dict.get("email")
+                username = user_dict.get("username")
+                tokenDoc = {"user": user_id, "email": email}
+                id = resetService.store_reset_password_token(tokenDoc, email, activate_ttl, user_id)
```
MarkLark86, line 485: `id` shadows the built-in function with the same name. It would be good not to use names that are built into the python language

MarkLark86 (commented):
Looks good, just a couple small things then this one should be good to go

superdesk/users/async_service.py
BrianMwangi21: @MarkLark86 for that function, as it is currently used in the `on_update` and `on_delete` functions in the service, do I remove it still in those function as it is going to be handled in the REST API before the functions are called ?

MarkLark86: Correct, we will validate the request in the REST API code

MarkLark86 (approved):

PR #2803, opened 2025-01-01
[PUBLISH] Push items directly to publish queue when possible.

jerome-poisson:
This patch add a `publish:push` command which directly put items to publish services queue without passing by the `published` collection.

SDESK-7448

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -8,73 +8,80 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-import logging
-import superdesk
-import superdesk.signals as signals
-
 from copy import copy, deepcopy
+import logging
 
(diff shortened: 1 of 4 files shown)


jerome-poisson:
**Note:** This is currently a work in progress; I am not entirely sure how to handle the correct publishing workflow for this scenario. I will need to discuss it further with @MarkLark86 .

jerome-poisson:
@MarkLark86 I've updated this PR following our discussion today. I'm keeping the `WIP` label because I still want to do some tests with frontend, but you and others can review the PR already.

Thanks!

MarkLark86 (approved):

MarkLark86:
> **Note:** This is currently a work in progress; I am not entirely sure how to handle the correct publishing workflow for this scenario. I will need to discuss it further with @MarkLark86 .

I wouldn't worry about publishing working at the moment, I think there's a few more kinks in the chain we need to fix before async publishing works.

PR #2809, opened 2025-01-07
[DOC] Add `plantuml` as apt package for ReadTheDocs.

jerome-poisson:


MarkLark86 (approved):
Seems to be working
https://superdesk--2809.org.readthedocs.build/en/2809/auth_server.html

PR #2810, opened 2025-01-07
[DOC] Add `plantuml` as apt package for ReadTheDocs.

jerome-poisson:


jerome-poisson:
We need to check that https://github.com/superdesk/superdesk-core/pull/2809 works and restores PlantUML diagram before merging this one.

MarkLark86 (approved):

PR #2811, opened 2025-01-09
Add Content profile name in Production API and Content API [SDESK-7440]

devketanpro:


apps/content_types/content_types.py
```diff
@@ -108,6 +108,7 @@ class ContentTypesResource(superdesk.Resource):
         "created_by": superdesk.Resource.rel("users", nullable=True),
         "updated_by": superdesk.Resource.rel("users", nullable=True),
         "init_version": {"type": "integer"},
+        "output_name": {"type": "string", "nullable": True, "mapping": not_analyzed},
```
petrjasek, line 111: we don't store that in elastic so there is no need for mapping

prod_api/content_types/service.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class ContentTypesService(ProdApiService):
+    excluded_fields = ProdApiService.excluded_fields
+
+    def get_output_name(self, profile):
+        try:
+            _id = bson.ObjectId(profile)
+            item = superdesk.get_resource_service("content_types").find_one(req=None, _id=_id) or {}
+            name = item.get("output_name") or item.get("label", str(_id))
```
petrjasek, line 24: would be better to update logic in https://github.com/superdesk/superdesk-core/blob/develop/apps/content_types/content_types.py#L239-L245 and use that method here instead of copying it

devketanpro: because when I tried to import functionality from the `superdesk_app`, I got some errors like `key_error`, so that's why I added a new service in the `prod_api` related to `content_types`.

petrjasek: would be good to reuse that to avoid that code duplication. not sure if we were doing so far

prod_api/content_types/service.py
```diff
@@ -0,0 +1,27 @@
...
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+from ..service import ProdApiService
+import bson
+import re
+import superdesk
+
+
+class ContentTypesService(ProdApiService):
```
petrjasek, line 17: I wonder why it's extending that base service?

devketanpro: to use predefined functions like `find_one`

prod_api/content_types/service.py
devketanpro: I tried to use this superdesk service here, but it's not working; may be I am missing something.

petrjasek: what exactly is not working when using the `ContentTypesService` ?

petrjasek: it would be also an option to put that code without `find_one` into some helper method and reuse that one at least

devketanpro: actually its gives us a keyerror, so we have to define a new `ContentTypesService` in the `production_api` app.

petrjasek: what keyerror exactly

devketanpro: So when we try to use a superdesk api `content_types` service, it fetches this `prod_api` service.
and when I tried to change the name of the `prod_api` service, it gave `None` in this case

petrjasek: when you add it to `INSTALLED_APPS` ?

petrjasek (approved):

PR #2815, opened 2025-01-14
SDAAP-126 Fix Height and width not always being defined for the custo…

marwoodandrew:
…m crop renditions.


marwoodandrew:
@MarkLark86 Would you mind have a look?

MarkLark86 (approved):

PR #2812, opened 2025-01-09
[SDESK-7465] - Create new async Desk resource and service

BrianMwangi21:
### Purpose
This PR creates new async desks resource and service:

- Implementing a new resource model for desks
- Implementing new async resource service for desks
- Implementing resource config for users and register it in desks module
- Add the module to the default `MODULES` config

Solves: [SDESK-7465](https://sofab.atlassian.net/browse/SDESK-7465)

[SDESK-7465]: https://sofab.atlassian.net/browse/SDESK-7465

PR #2816, opened 2025-01-14
[Vocabularies] Create new async Vocabularies resource, service & REST…

jerome-poisson:
… API.

SDESK-7468


jerome-poisson:
Not fully finished, I have a few points that I need to check (see FIXME).

MarkLark86 (commented):
There are a couple more tasks that are left here, such as creating the module, registering the resource, disabling the REST API from old resource service, and updating the default_settings.MODULE.

See the [docs](https://superdesk.readthedocs.io/en/async/core/resources.html#registering-resources) or this [PR](https://github.com/superdesk/superdesk-core/pull/2812/files) for an example

superdesk/vocabularies/async_vocabularies.py
```diff
@@ -0,0 +1,486 @@
...
+    initial_offset_minutes: int
+
+
+class VocabulariesResourceModel(ResourceModel):
+    display_name: str
+    description: str | None = None
+    helper_text: Annotated[str | None, validate_maxlength(120)] = None
+    tags: list[Tag] | None = None
+    popup_width: int | None = None
+    type_: str  # Use type_ instead of type due to reserved keyword conflict
```
MarkLark86, line 88: We need to make this an alias, so the DB and front-end still stores the item in the same dict key. We can then use a name that provides context too. Something like 
```python
from enum import Enum, unique

@unique
class CVAccessType(str, Enum):
    MANAGEABLE ="manageable"
    UNMANAGEABLE = "unmanageable"

class VocabulariesResourceModel(ResourceModel):
    ...
    management_type: Annotated[CVAccessType, Field(alias="type")]
```

superdesk/vocabularies/async_vocabularies.py
```diff
@@ -0,0 +1,486 @@
...
+# TODO(petr): add api to specify vocabulary schema
+vocab_schema = {
+    "crop_sizes": {
+        "width": {"type": "integer"},
+        "height": {"type": "integer"},
+    }
+}
+
+
+@dataclass
```
MarkLark86, line 57: For dataclasses and Pydantic models, we have our own implementation that provides a common set of functions. So instead we would do:

```python
from superdesk.core.resources import Dataclass, BaseModel # Use our BaseModel class

class Tag(Dataclass):
    ...
```

We then have access to functions, such as `to_dict` and `from_dict` etc

superdesk/vocabularies/async_vocabularies.py
```diff
@@ -0,0 +1,486 @@
...
+class Tag:
+    text: str
+
+
+class Item(BaseModel):
+    model_config = ConfigDict(extra="allow")
```
MarkLark86, line 63: We also allow extra by default with our own BaseModel

superdesk/vocabularies/async_vocabularies.py
```diff
@@ -0,0 +1,486 @@
...
+
+class VocabulariesResourceModel(ResourceModel):
+    display_name: str
+    description: str | None = None
+    helper_text: Annotated[str | None, validate_maxlength(120)] = None
+    tags: list[Tag] | None = None
+    popup_width: int | None = None
+    type_: str  # Use type_ instead of type due to reserved keyword conflict
+    items: list[Item]
+    selection_type: str | None = None
```
MarkLark86, line 90: It would be good to use an Enum here, to provide validation (as we had `allowed` list of choices in old resource)

MarkLark86 (commented):
Model looks good to me, just a couple improvement suggestions

jerome-poisson:
@MarkLark86 thanks for your feedbacks, I should have addressed all of them, let me know if it's fine now.

jerome-poisson:
My resource doesn't appear in the client, I'm not sure why.
Also, how are we supposed to convert `privilege`?
https://github.com/superdesk/superdesk-core/blob/7afb864add621f39dabcee151e8353e1b8076fe3/superdesk/vocabularies/vocabularies.py#L173

MarkLark86:
> My resource doesn't appear in the client, I'm not sure why.

I'm guessing the client is using the home endpoint to get the list of endpoints, and using that. This endpoint is populated by [eve](https://github.com/superdesk/eve/blob/async/eve/endpoints.py#L111). I will look into adding our async resources to that home endpoint. I will comment here once that's done

> Also, how are we supposed to convert `privilege`?
> https://github.com/superdesk/superdesk-core/blob/7afb864add621f39dabcee151e8353e1b8076fe3/superdesk/vocabularies/vocabularies.py#L173

Helmy has just created a [PR](https://github.com/superdesk/superdesk-core/pull/2824) to add this functionality. For now, we can leave auth out of this resource until the auth rule PR is merged.

MarkLark86:
@jerome-poisson I have updated your PR in your branch, to get the REST API working, along with a couple other fixes (like auth fix, cors headers and a couple issues in the model & service).

It works locally, **except** I am unable to edit an existing CV. I haven't done thorough testing there, but looks like from the front-end it's trying to use POST instead of PATCH when updating an existing item. Can you please investigate and find a solution

superdesk/vocabularies_async/service.py
```diff
@@ -58,14 +58,14 @@ class VocabulariesService(AsyncResourceService[VocabulariesResourceModel]):
...
     async def _validate_items(self, update: VocabulariesResourceModel) -> None:
         # if we have qcode and not unique_field set, we want it to be qcode
-        if update.schema_.get("qcode") and update.unique_field is None:
+        if update.schema.get("qcode") and update.unique_field is None:
             update.unique_field = "qcode"
 
         unique_field = update.unique_field
         vocabs = {}
-        if update.schema_ and update.items:
+        if update.schema and update.items:
```
jerome-poisson, line 66: I did use `schema_` because `schema` is already used in `ResourceModel`. With your change, I have that:
```
/home/goffi/dev/sourcefabric/venv/async_3.10/lib/python3.10/site-packages/pydantic/_internal/_fields.py:172: UserWarning: Field name "schema" in "VocabulariesResourceModel" shadows an attribute in parent "ResourceModel"
```

jerome-poisson:
I still can't run the client at all, I have this message in JS console:
```
Uncaught (in promise) 
Object { status: 404, resource: "vocabularies" }
```
And I can't even log in (in fact, I don't have logging screen at all).

If you could run the frontend with this PR I must have something wrong in my config or something, I'm investigating.

superdesk/vocabularies_async/service.py
jerome-poisson: I'm reverting to `schema_` and using:
```python
schema_: Annotated[dict[str, dict], Field(alias="schema")]
```

superdesk/factory/app.py
```diff
@@ -414,6 +416,24 @@ def setup_sentry(self):
 
         sentry_sdk.init(dsn=self.config["SENTRY_DSN"], integrations=[QuartIntegration(), AsyncioIntegration()])
 
+    def extend_eve_home_endpoint(self, links: list[dict]) -> None:
```
jerome-poisson, line 419: This method is never called.

MarkLark86: You'll need to grab latest eve lib (from our superdesk/eve/Async repo & branch)

jerome-poisson:
@MarkLark86 the client is using a wrong HTTP verb due to the `_links` field missing, it was done in older `Resource` with `build_custom_hateoas` at https://github.com/superdesk/superdesk-core/blob/develop/superdesk/resource.py#L44 .

jerome-poisson:
`_links` field depends on the work currently done by @eos87 , so I'm merging current PR.

PR #2818, opened 2025-01-16
[SDESK-7472] Add async support to ingest handlers

eos87:
### Purpose
Add async support to ingest handlers

### What has changed
- Adjusted base class to implement async `can_handle` and `apply_rule` methods
- Adjusted references handlers and references to async/await calls

Resolves: [SDESK-7472]

[SDESK-7472]: https://sofab.atlassian.net/browse/SDESK-7472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/rules/rule_handlers.py
+++ b/apps/rules/rule_handlers.py
@@ -8,9 +8,9 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-from typing import Dict, Any, Awaitable
 import logging
 
+from typing import Dict, Any
 from quart_babel import lazy_gettext, LazyString
 
 from superdesk.core import get_app_config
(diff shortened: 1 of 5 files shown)


apps/rules/rule_handlers.py
```diff
@@ -30,10 +29,10 @@ class RoutingRuleHandler:
     supported_configs: Dict[str, bool]
     default_values: Dict[str, Any]
 
-    def can_handle(self, rule, ingest_item, routing_scheme) -> bool:
+    async def can_handle(self, rule, ingest_item, routing_scheme) -> Awaitable[bool]:
```
MarkLark86, line 32: What is the need for `Awaitable[` on the return type?

MarkLark86 (approved):
:+1:  looks good

BrianMwangi21 (approved):

apps/rules/rule_handlers.py
eos87: Just a leftover when I was trying to support both sync and async overload from children. Fixed.

PR #2824, opened 2025-01-27
[SDESK-7460] Create async auth rules for user privileges

eos87:
### Purpose
To implement privilege-based authentication rules for async rest endpoints.

### What has changed
- Added new privilege rules and configurations to ensure that only users with the right permissions can access certain resources.
- Updated the test files to include scenarios for checking access to privileged resources.
- Introduced new utility functions to help with privilege and role management. Also another one to flat nested lists.
- Fixed json serialisation issue when saving session (`SecureCookieSessionInterface` not being able to handle `ObjectId` values).
- Made some small tweaks to the error handling and configuration files to support the new privilege system.
- Added tests to make sure the new rules work as expected.

### Pending
~- I will push some changes to fix all (**_unrelated_**) `mypy` issues tomorrow~. Done!

Resolves: [SDESK-7460]

Thanks for checking!

[SDESK-7460]: https://sofab.atlassian.net/browse/SDESK-7460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,5 +20,5 @@ exclude = '''
 [tool.pytest.ini_options]
 testpaths = ["tests", "superdesk", "apps", "content_api"]
 python_files = "*_test.py *_tests.py test_*.py tests_*.py tests.py test.py"
-asyncio_mode = "auto"
+asyncio_mode = "strict"
 asyncio_default_fixture_loop_scope = "function"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,5 +20,5 @@ exclude = '''
 [tool.pytest.ini_options]
 testpaths = ["tests", "superdesk", "apps", "content_api"]
 python_files = "*_test.py *_tests.py test_*.py tests_*.py tests.py test.py"
-asyncio_mode = "strict"
+asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "function"
--- a/superdesk/core/auth/privilege_rules.py
+++ b/superdesk/core/auth/privilege_rules.py
@@ -1,0 +1,62 @@
+from typing import Any, cast
+
+from quart_babel import gettext
+from superdesk.errors import SuperdeskApiError
+from superdesk.users.async_service import get_privileges, is_admin
+from superdesk.core.types import HTTP_METHOD, AuthRule, Request
+
+
+def required_privilege_rule(privilege: str) -> AuthRule:
+    """
+    Creates an authentication rule that restricts access based on a specific privilege.
+
+    Args:
+        privilege (str): The privilege required to access the resource.
+
+    Returns:
+        AuthRule: An asynchronous rule function that checks if the user has the required privilege.
+
+    The rule verifies:
+    - If the user is an admin, access is granted without further checks.
+    - Otherwise, the user's privileges are retrieved and checked against the required privilege.
+
+    Raises:
+        SuperdeskApiError: If the user does not have the required privilege.
+    """
+
+    async def internal_rule(request: Request):
+        current_user = request.user
+
+        # TODO-ASYNC: Check if for admin we need to pull all privileges instead
+        if is_admin(current_user):
+            return None
+
+        user_role = request.storage.request.get("role")
+        user_privileges = get_privileges(request.user, user_role)
+
+        if user_privileges.get(privilege, False):
+            return None
+
+        raise SuperdeskApiError.forbiddenError(message=gettext("Insufficient privileges for the requested operation."))
+
+    return internal_rule
+
+
+def privilege_based_rules(rules: dict[HTTP_METHOD, str]) -> dict[str, AuthRule]:
+    """
+    Generates a dictionary of authentication rules for HTTP methods based on privileges.
+
+    Args:
+        rules (dict[HTTP_METHOD, str]): A mapping of HTTP methods (e.g., GET, POST) to the required privileges
+        for those methods.
+
+    Returns:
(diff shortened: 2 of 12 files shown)


pyproject.toml
```diff
@@ -20,5 +20,5 @@ exclude = '''
 [tool.pytest.ini_options]
 testpaths = ["tests", "superdesk", "apps", "content_api"]
 python_files = "*_test.py *_tests.py test_*.py tests_*.py tests.py test.py"
-asyncio_mode = "strict"
+asyncio_mode = "auto"
```
MarkLark86: Why the change to `auto`? Pytest in Superdesk doesn't use the asyncio plugin for tests (instead we use the pytest provided one).
See the change [here](https://github.com/superdesk/superdesk-core/blob/async/.github/workflows/nose-tests.yml#L28)

superdesk/core/auth/privilege_rules.py
```diff
@@ -0,0 +1,62 @@
...
+    Raises:
+        SuperdeskApiError: If the user does not have the required privilege.
+    """
+
+    async def internal_rule(request: Request):
+        current_user = request.user
+
+        # TODO-ASYNC: Check if for admin we need to pull all privileges instead
+        if is_admin(current_user):
+            return None
```
MarkLark86, line 31: This should be fine, if the user is admin then they have all privileges by design

superdesk/core/auth/privilege_rules.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if user_privileges.get(privilege, False):
+            return None
+
+        raise SuperdeskApiError.forbiddenError(message=gettext("Insufficient privileges for the requested operation."))
+
+    return internal_rule
+
+
+def privilege_based_rules(rules: dict[HTTP_METHOD, str]) -> dict[str, AuthRule]:
```
MarkLark86: As this rule is for checking against a HTTP method / role dictionary, it might be good to indicate that in the name of the rule

superdesk/core/auth/privilege_rules.py
```diff
@@ -0,0 +1,62 @@
...
+        for those methods.
+
+    Returns:
+        dict[str, list[AuthRule]]: A dictionary where each HTTP method is mapped to a list of authentication
+        rules that restrict access based on the specified privilege.
+    """
+
+    auth_rules: dict[str, AuthRule] = {}
+    for method, privilege_name in rules.items():
+        auth_rules[method] = required_privilege_rule(privilege_name)
```
MarkLark86, line 59: Rather than returning a dictionary, which you then later have to flatten, what if you provide an internal rule like the other one.
This rule can then check the request method directly, so code that uses this rule doesn't have to check themselves.

i.e.
```python
async def internal_rule(request: Request):
    privilege_name = rules.get(request.method)
    if privilege_name:
        await required_privilege_rule(privilege_name)(request) 
```

Would something like that work?

superdesk/tests/__init__.py
```diff
@@ -479,8 +479,12 @@ async def setup_db_user(context, user):
 
         user["password"] = original_password
         auth_data = json.dumps({"username": user["username"], "password": user["password"]})
+
+        context.headers.append(["Content-Type", "application/json"])
```
MarkLark86: Although most requests are, not every request is "application/json". Where is this causing an issue, might need to be fixed there instead

tests/core/privilege_rules_test.py
```diff
@@ -0,0 +1,69 @@
+import pytest
+from unittest.mock import patch, MagicMock
+from superdesk.errors import SuperdeskApiError
+from superdesk.core.auth.privilege_rules import privilege_based_rules, required_privilege_rule
+
+
+@patch("superdesk.core.auth.privilege_rules.is_admin")
```
MarkLark86: These 3 would be skipped, as we're using `-p no:asyncio` in the pytest call (asyncio plugin was causing an issue, so I disabled it in superdesk tests).
Instead, you could use the test class' we have in superdesk.tests module

tests/core/resource_endpoints_test.py
```diff
@@ -422,3 +424,82 @@ async def test_endpoint(self, mock_utcnow):
...
+class PrivilegedResourceEndpointsTestCase(AsyncFlaskTestCase):
+    use_default_apps = True
+    app_config = {
+        "MODULES": [
+            "tests.core.modules.users",
+            "tests.core.modules.module_with_privileges",
+        ]
+    }
+
+    async def asyncSetUp(self):
```
MarkLark86: This seems to have no effect, and can be removed from this class

pyproject.toml
eos87: I see. I will fix it. And what's the reason we don't use it?

superdesk/core/auth/privilege_rules.py
eos87: I'll change the name to something like `http_method_privilege_based_rules` which feels a bit more meaningful.

tests/core/resource_endpoints_test.py
eos87: leftover. Fixed.

tests/core/privilege_rules_test.py
eos87: I converted the tests into unit test class

superdesk/tests/__init__.py
eos87: Fixed

MarkLark86 (approved):

PR #2825, opened 2025-02-03
[SDESK-7484] Create async Content API Subscriber Token and Auth

eos87:
### Purpose
Implement a new token-based authentication system specific to the ContentAPI, replacing the default Superdesk async auth with a custom SubscriberToken mechanism.

### What has changed
- Introduced a new `SubscriberToken` resource model and an accompanying async service.
- Created a `ResourceConfig` for the `SubscriberToken` resource, enabling its REST API. Added corresponding module.
- Implemented a new `SubscriberTokenAuth` class for the ContentAPI.
- Set the ASYNC_AUTH_CLASS config in the ContentAPI app to the new auth class.
- Removed the old `SubscriberTokenAuth` class and related resource code.
- Updated necessary imports and configurations to reflect the new changes.
- Implemented optionally generating nested HATEOAS information for items in async rest api.
- Allow deleting items via rest API by fixing issue with CORS and wrong methods (it was using `resource_methods` instead of `items_methods`).

[SDESK-7484]

[SDESK-7484]: https://sofab.atlassian.net/browse/SDESK-7484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


superdesk/core/resources/resource_rest_endpoints.py
```diff
@@ -563,10 +572,69 @@ def _build_resource_hateoas(self, req: SearchRequest, doc_count: Optional[int],
...
+
+    def _get_related_links_from_annotations(self, item: dict[str, Any]) -> dict[str, Any]:
+        """Extract related links from the `resource_config.data_class` annotations.
+
+        Examines the data class annotations for Annotated fields that contain AsyncValidator metadata.
+        For each field with a resource name specified in its validator, generates a related link
+        if the field has a value in the item.
+        """
+        related_links = {}
+        for field_name, annotation in get_annotations(self.resource_config.data_class).items():
```
MarkLark86: `get_annotations` does not provide annotations from mixin or super classes (known limitation). I came across this with validations. Instead we have the following:
https://github.com/superdesk/superdesk-core/blob/c39fc7c519d85a96f67adcf2ff18faf7d64cb985/superdesk/core/resources/model.py#L243-L249 

Might be good to put the above code in a util to be used by both places.

MarkLark86: Also, I fear this might cause a bit of a performance issue. If the response includes 200 items, then we're calculating the related fields for each item.

A quick solution might be to store the related fields metadata on the ResourceRestEndpoints class instance, generate it once (on `__init__` maybe), and use that list thereafter.

MarkLark86 (commented):
:+1: Looks good. A minor improvement regarding HATEOAS & related items.

As for the tests, there are some failing, and it would be very beneficial if we had tests for the item-level HATEOAS, especially considering it's a required feature for some front-end features to work.

superdesk/core/resources/resource_rest_endpoints.py
eos87: Good observation. I extracted part of the logic into a separate util function + relocated some of the things into the ResourceModel itself so the related attributes are calculated on class initialization and we do not need to do it over and over again. Please take another look.

eos87:
> As for the tests, there are some failing, and it would be very beneficial if we had tests for the item-level HATEOAS, especially considering it's a required feature for some front-end features to work.

I've fixed the web signal broken tests (unrelated) and I'm working on fixing the rest and implementing new tests cases for the new logic. I'll have it ready tomorrow.

superdesk/core/resources/resource_rest_endpoints.py
```diff
@@ -594,47 +594,7 @@ def _populate_item_hateoas(
         }
 
         # extract related links from the `resource_config.data_class` relations (annotations)
-        related_links = self._get_related_links_from_annotations(item)
-        if related_links:
+        if related_links := self.resource_config.data_class.get_related_links(item):
```
MarkLark86, line 597: :+1:  Don't see `:=` too often

superdesk/core/resources/model.py
```diff
@@ -148,6 +152,29 @@ def type(self) -> str:
     #: Datetime the document was last updated
     updated: Annotated[Optional[datetime], Field(alias="_updated")] = None
 
+    def __init_subclass__(cls) -> None:
```
MarkLark86, line 155: Nicely done :nerd_face:

superdesk/core/resources/resource_rest_endpoints.py
eos87: I saw it once, long ago when it was proposed, then I forgot it was already there in python 3. I really like it, I think I will be using it more often 😃

MarkLark86 (approved):

eos87:
Tests have been added. I wanted to add some more tests in the `content_api` to verify the functionality of the new Token Auth class but I don't want to spend more time in this ticket. Since the `content_api` will need much more work in terms of moving it to async, we can add those tests later on.

PR #2829, opened 2025-02-11
[SDESK-7466] Upgrade Publish resources to async

MarkLark86:
### Purpose
Upgrade database resources used by the publishing system to async.

### What has changed
* Upgrade `content_filter`, `filter_condition`, `products`, `publish_queue`, `subscribers` and `sequences` resources to async
* New `superdesk.publish_async` module, where these resources are all registered at once
* Implement a custom publishing cache, for use when checking items against subscribers/products/content filters
* Add clone, clone_with to dataclass and model
* Add `on_fetched` functions to the ResourceRestEndpoints (to provide similar functionality to old rest system)
* Fix Vocabularies async resource, and implement the on_fetched function in a custom rest endpoint class
* Update content filtering based tests to use the new async code


superdesk/core/resources/service.py
```diff
@@ -390,8 +390,8 @@ async def create(self, _docs: Sequence[ResourceModelType | dict[str, Any]]) -> L
             if self.config.versioning:
                 await self.insert_versioned_document(doc_dict)
 
-        await self.on_created(docs)
-        return ids
+        await self.on_created(instances)
+        return instances
```
eos87, line 426: This change will probably break `newsroom-core` as the returned value used to be a list of integers and we're using it there to fetch the newly created item from db.

superdesk/core/resources/service.py
```diff
@@ -477,6 +477,8 @@ async def update(
 
         await self.on_updated(updates, original)
 
+        return original.clone_with(updates_dict)
```
eos87, line 512: Good addition! 👍

superdesk/core/resources/model.py
```diff
@@ -103,6 +103,14 @@ def to_json(self, **kwargs) -> str:
         default_params.update(kwargs)
         return RootModel(self).model_dump_json(**default_params)
 
+    def clone(self) -> Self:
+        return self.from_dict(deepcopy(self.to_dict()))
```
eos87: Don't you think this has the same result as built-in `.model_copy`?

superdesk/core/resources/service.py
```diff
@@ -545,18 +547,30 @@ async def on_deleted(self, doc: ResourceModelType):
...
+        return [item async for item in self.get_all(lookup)]
+
+    async def get_all_list_raw(self, lookup: dict | None = None) -> list[dict]:
+        return cast(list[dict], [item async for item in self.get_all_raw(lookup)])
+
+    async def get_all_map(self, lookup: dict | None = None) -> dict[str | ObjectId, ResourceModelType]:
+        return {item.id: item async for item in self.get_all(lookup)}
+
+    async def get_all_map_raw(self, lookup: dict | None = None) -> dict[str | ObjectId, dict]:
+        return cast(dict[str | ObjectId, dict], {item[ID_FIELD]: item async for item in self.get_all_raw(lookup)})
```
eos87: I saw the use case for these. Just a thought, from the performance point of view. Should we add a comment or in the docstrings suggesting that we should avoid, when possible, getting the whole list of items and instead use the method that returns a generator?

superdesk/publish_async/subscribers/rest_api.py
```diff
@@ -0,0 +1,66 @@
...
+        self,
+        args: ItemRequestViewArgs,
+        params: ItemRequestUrlArgs,
+        request: Request,
+    ) -> Response:
+        """Processes a get single item request"""
+        response = await super().get_item(args, params, request)
+
+        body = response.body.to_dict() if isinstance(response.body, SubscribersResource) else cast(dict, response.body)
+        self._hide_config_fields(body, self.hide_fields)
```
eos87: Isn't it possible to use `projection` for this?

superdesk/publish_async/subscribers/utils.py
```diff
@@ -0,0 +1,147 @@
...
+    :param max_seq_number: default None, maximal possible value, None means no upper limit
+    :param min_seq_number: default 1, init value, sequence will start from the NEXT one
+    :returns: sequence number
+    """
+    if not key_name:
+        logger.error("Empty sequence key is used: {}".format("\n".join(traceback.format_stack())))
+        raise KeyError("Sequence key cannot be empty")
+
+    target_resource = SequencesResource.get_service()
+    # target_resource.mongo.find_one_and_update()
```
eos87:
```suggestion
```

tests/content_filters/filter_condition_tests.py
```diff
@@ -90,29 +111,79 @@ async def asyncSetUp(self):
...
+            #     "filter_conditions",
+            #     [{"_id": 4, "field": "urgency", "operator": "nin", "value": "1,2,3", "name": "test-2"}],
+            # )
+            # self.app.data.insert(
+            #     "filter_conditions",
+            #     [{"_id": 5, "field": "urgency", "operator": "in", "value": "2,5", "name": "test-2"}],
+            # )
+            # self.app.data.insert(
+            #     "content_filters", [{"_id": 1, "content_filter": [{"expression": {"fc": [1]}}], "name": "soccer-only"}]
+            # )
```
eos87: Do we need this?
```suggestion
```

eos87 (approved):
Looks good, I just left a few minor comments.

superdesk/core/resources/service.py
MarkLark86: I will create a separate PR to address this

superdesk/core/resources/service.py
MarkLark86: I have added the docs as suggested. And yes, the developer should be aware of performance and resource usage when using these. Maybe down the track we can add a resource config to disable these on a particular resource - to disallow getting all items without using an async generator)

MarkLark86: Even with async generate, if the collection is big there would be a timeout on the cursor. That is why we have this `get_all_batch` function, uses pagination and async generator together

MarkLark86:
Hey @eos87 I added a config option to rest api, ability to always exclude field(s), as an improved solution to subscriber config (API keys etc). Could you have another look please. There are tests as well

superdesk/core/elastic/sync_client.py
```diff
@@ -110,14 +110,23 @@ def search(self, query: Dict[str, Any], indexes: Optional[List[str]] = None) ->
...
         :param item_id: ID of the document to find.
+        :param projection: The field projections to be applied
         :return: The document found or None if no document was found.
         """
 
         try:
+            if (
+                projection
+                and isinstance(projection, dict)
+                and not set(projection.keys()).intersection({"_source", "_source_excludes"})
```
eos87, line 131: Just to try to understand, why is the projection only applied if there is no intersection with these two `"_source", "_source_excludes"`?

eos87 (approved):
Looks good! 🚀

superdesk/core/elastic/sync_client.py
MarkLark86: `_source_ and `_source_excludes` are [params specific](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/elastic/base_client.py#L28) to Elasticsearch, so the params have already been converted over to elastic format.

We cannot use `isinstance` here because we're using a union of multiple types, ProjectedFieldSources is a dict and ProjectedFieldArg is a [union](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/types/search.py#L11) of multiple types, which `isinstance` does not work.

So instead we check the existence of the ES specific fields (which aren't in the ProjectedFieldArg type), and cast there.
It's dirty, but it works

MarkLark86: I've added comments to the code to indicate this, for future use

PR #2831, opened 2025-02-12
Sdesk 7469

jerome-poisson:
This PR is still a WIP. The classes are quite extensive, and the methods calling each other in a layered fashion make it challenging to follow. I have several questions about how to best manage certain method or type porting that I would like to tackle.


superdesk/archive_async/module.py
```diff
@@ -0,0 +1,50 @@
...
+
+archive_resource_config = ResourceConfig(
+    name="archive",
+    data_class=ArchiveResourceModel,
+    service=ArchiveService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(
+                name="uri_1",
+                keys=[("uri", 1), ("background", 1)],
```
MarkLark86, line 25: [unique](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/types/mongo.py#L35) defaults to `True`, therefor you would need to add `unique=False` all these index options

superdesk/archive_async/module.py
```diff
@@ -0,0 +1,50 @@
...
+                name="processed_from_1",
+                keys=[("processed_from", 1), ("background", 1)],
+            ),
+            MongoIndexOptions(
+                name="assignment_id_1",
+                keys=[("assignment_id", 1), ("background", 1)],
+            ),
+        ],
+    ),
+    rest_endpoints=RestEndpointConfig(
```
MarkLark86: We will continue to use the old resource for the REST API, therefor we won't need `rest_endpoints` config here for now (this will be done at a later stage)

superdesk/archive_async/service.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    return bool(item.linked_in_packages is not None and sum(1 for _ in item.linked_in_packages))
+
+
+class ArchiveService(AsyncResourceService[ArchiveResourceModel], HighlightsSearchMixin):
+    packageService = PackageService()
+    mediaService = ArchiveMediaService()
+    cropService = CropService()
+
+    async def on_fetched(self, docs) -> None:
```
MarkLark86: These `on_fetched` don't belong on the new AsyncResourceService as they have nothing to do with the data layer (these are used on REST APIs). They can be removed from this service (will be added to a custom REST API class later for this resource)

superdesk/archive_async/service.py
```diff
@@ -0,0 +1,77 @@
...
+        self.enhance_items(docs[ITEMS])
+
+    async def on_fetched_item(self, doc) -> None:
+        self.enhance_items([doc])
+
+    def enhance_items(self, items) -> None:
+        for item in items:
+            handle_existing_data(item)
+
+    async def on_create(self, docs: list[ArchiveResourceModel]) -> None:
```
MarkLark86, line 213: Is there a way to make a base class that this and the original resource service can both use? The old Eve resources now support async `on_` functions (such as `def on_create`, `def on_update` etc), just not on the core method functions (such as `def create`, `def update` etc).

I've done this with the async publish endpoints to reduce the amount of code changes required. I used the [package service](https://github.com/superdesk/superdesk-core/blob/develop/apps/packages/package_service.py#L93) as an example (pure python class, implementing the `on_create` etc methods).

And then you would update both the old and this service to use those functions.
Keeping in mind, if the class makes changes to the dictionary (especially here in creation), then you would need to propagate those changes onto the Pydantic model instance), maybe just using the [clone_with](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/types/model.py#L127) function

superdesk/archived_async/module.py
```diff
@@ -0,0 +1,34 @@
...
+
+archived_resource_config = ResourceConfig(
+    name="archived",
+    data_class=ArchivedResourceModel,
+    service=ArchivedService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(
+                name="item_id_1",
+                keys=[("item_id", 1), ("background", 1)],
```
MarkLark86, line 25: Same here, will require `unique=False`

superdesk/archived_async/module.py
```diff
@@ -0,0 +1,34 @@
...
+            ),
+        ],
+    ),
+    rest_endpoints=RestEndpointConfig(
```
MarkLark86: And here as well, REST config not required (front-end will continue to use the old REST API for now)

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
+# This file is part of Superdesk
```
MarkLark86, line 1: We also have the [Content API resource model](https://github.com/superdesk/superdesk-core/blob/async/content_api/items/model.py). There might be a lot of common fields here.

The ContentAPI items are slightly different to archive items, and it is also used by Newshub. But hopefully you might be able to combine them (and maybe move some of those classes into a `superdesk.types` module for common dataclasses etc

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+class Author:
+    uri: str
+    parent: str
+    name: str
+    role: str
+    jobtitle: str
+    sub_label: str
+
+
+@dataclass
```
MarkLark86: We have this defined as a core field, [Geopoint](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L209), which provides elastic mapping, json schema and validation

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+class LinkedInPackage:
+    package: Annotated[str, validate_data_relation_async("archive")]
+    package_type: str | None = Field(None, deprecated=True)
+
+
+class MetadataResource(ResourceModel):
+    guid: str
+    uri: str
```
MarkLark86: Any string fields in the old resource that have `"mapping": not_analyzed"` need to use [fields.Keyword](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L209), that way the elastic mapping will be correct

MarkLark86: There is also `TextWithKeyword` and `HTML` fields defined in core fields

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    unique_id: int
```
MarkLark86: Any fields in the old resource that has `unique` validation, requires [validate_unique_value_async](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/validators.py#L204) from core validators.

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    version_creator: Annotated[str | None, validate_data_relation_async("users")] = None
+    firstcreated: datetime
+    versioncreated: datetime
+    firstpublished: datetime | None = None
+    ingest_provider: Annotated[str | None, validate_data_relation_async("ingest_providers")] = None
+    source: str
+    original_source: str
+    ingest_provider_sequence: str
+    usageterms: str | None = None
+    copyrightnotice: str | None = None
```
MarkLark86: Any fields in the old resource that has `"mapping": not_indexed` will need to be annotated with [fields.not_indexed()](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L205) so Elastic mapping will be correct.

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    copyrightholder: str | None = None
+    anpa_category: list[Subject] | None = None
+    subject: list[Subject]
```
MarkLark86: Any list fields that has `"type": "nested"` in the original schema needs to be [annotated](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L182) as such here as well, so the elastic mapping is correct.

In this case it would be:
```python
subject: Annotated[list[Subject] | None, fields.nested_list(include_in_parent=True)] = None
```

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+@dataclass
+class Broadcast:
+    status: str
+    master_id: str
+    rewrite_id: str
+
+
+@dataclass
+class Subject:
+    qcode: str
```
MarkLark86: Please go over all usage of `not_analyzed` in the original schema, as it's used a lot, and replace with `fields.Keyword`

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    original_source: str
+    ingest_provider_sequence: str
+    usageterms: str | None = None
+    copyrightnotice: str | None = None
+    copyrightholder: str | None = None
+    anpa_category: list[Subject] | None = None
+    subject: list[Subject]
+    genre: list[Genre] | None = None
+    company_codes: list[CompanyCodes] | None = None
+    item_type: ContentTypes = Field(ContentTypes.TEXT, alias="type")
```
MarkLark86: Using an enum is good, but we also need to mark it as `not_analyzed` so elastic mapping is correct. You can annotate it with [fields.keyword_mapping](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L170) in this case (I haven't tried it with an Enum field before, but it would work)

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    package_type: str | None = Field(None, deprecated=True)
+    language: str | None = None
+    abstract: str | None = None
```
MarkLark86: Any field that has `"mapping": string_with_analyzer` in the old schema needs to use the [fields.HTML](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L144) type, as it uses the same `html_field_analyzer`

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    headline: str | None = None
+    slugline: Slugline
```
MarkLark86: Slugline is a special case and needs custom elastic mapping. This can be achieved using the [fields.elastic_mapping](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L166) annotation. In this case it would be:

```python
Slugline: Annotated[
    str,
    fields.elastic_mapping(
        {
            "type": "string",
            "fielddata": True,
            "fields": {
                "phrase": {
                    "type": "string",
                    "analyzer": "phrase_prefix_analyzer",
                    "fielddata": True,
                },
                "keyword": {
                    "type": "keyword",
                },
                "text": {"type": "string", "analyzer": "html_field_analyzer"},
            }
        }
    )
]
```

This should hopefully give you the same elastic mapping as the original schema

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    anpa_take_key: str | None = None
+    correction_sequence: int | None = None
+    rewrite_sequence: int | None = None
+    # FIXME: duplicate of ItemSchema
+    rewrite_of: str | None = None
+    # FIXME: duplicate of ItemSchema
+    rewritten_by: str | None = None
+    # FIXME: duplicate of ItemSchema
+    sequence: int | None = None
+    keywords: list[str]
```
MarkLark86: You can also use fields.Keyword with lists, i.e. `list[fields.Keyword]`

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    word_count: int
+    priority: int | None = None
+    urgency: int | None = None
+    profile: str | None = None
+    state: ContentStates
```
MarkLark86: These 3 fields also need to be annotated by the fields.keyword_mapping, to make sure they aren't analyzed by Elastic

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+
+
+class PubStatuses(Enum):
+    USABLE = "usable"
+    HOLD = "withheld"
+    CANCELED = "canceled"
+
+
+@dataclass
+class Author:
```
MarkLark86: Have a look at the [ContentAuthor](https://github.com/superdesk/superdesk-core/blob/async/content_api/items/model.py#L74) type from the ContentAPI, might be able to be re-used here

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    state: ContentStates
+    revert_state: ContentStates | None = None
+    pubstatus: PubStatuses | None = None
+    signal: list[Subject]
+    byline: str | None = None
+    ednote: str | None = None
+    authors: list[Author] | None = None
+    description_text: str | None = None
+    archive_description: str | None = None
+    groups: list[GroupRef] | None = None
```
MarkLark86: Any fields with `"minlength"` or `"maxlength"` defined in the old schema can use the [validate_minlength](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/validators.py#L62) and/or [validate_maxlength](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/validators.py#L87) validators from core

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    tz: str
+    country_code: str
+    dateline: str
+    alt_name: str
+    state: str
+    city_code: str
+    country: str
+    code: str
+    scheme: str
+    location: list[Geopoint]
```
MarkLark86: Is `location` a list or a dict, seems to be a [dict](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/metadata/item.py#L141) in the original schema

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+
+
+@dataclass
+class Place:
+    code: str
+    name: str
+    qcode: str
+    scheme: str
+    feature_class: str
+    location: list[Geopoint]
```
MarkLark86: And again here, seems to be a dict not a list

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    ednote: str | None = None
+    authors: list[Author] | None = None
+    description_text: str | None = None
+    archive_description: str | None = None
+    groups: list[GroupRef] | None = None
+    deleted_groups: list[str] | None = None
+    body_html: str | None = None
+    body_text: str | None = None
+    dateline: Dateline | None = None
+    expiry: datetime
```
MarkLark86: Expiry may not always have a value, so we can make it optional with `datetime | None = None`

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    # FIXME: type file
+    media: str | None = None
```
MarkLark86: Hmmm, actually not sure if we currently support `"type": "file"` in the new async lib.

This is because it's a special field, where Eve will grab the file from the HTTP request, store the file, and replace this field's value with the ID of the uploaded file.

I will need to look into adding support for this. For now we can mark this with `fields.Keyword` and place a `# TODO-ASYNC: Add support for `file` and `media` field types

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    mimetype: str
+    poi: POI
+    renditions: dict[str, dict]
```
MarkLark86: Just an FYI, any field with a type annotated of `dict` will automatically have `"mapping": not_enabled` functionality applied (nothing to do here, just wanting to let you know that)

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+@dataclass
+class FeatureMedia:
+    _id: str
+    guid: str
+    unique_id: int
+
+
+@dataclass
+class Associations:
+    featuremedia: FeatureMedia | None = None
```
MarkLark86: This would cause a validation error for non featuremedia associations.

Maybe we could just use the fields.elastic_mapping feature to implement proper elastic mapping, and use simple dict as the type. i.e.

```python
Associations = Annotated[
    dict,
    fields.elastic_mapping(
        "type": "object",
        "dynamic": False,
        "properties": {
            "featuremedia": {  # keep indexing featuremedia - we do some filtering using it
                "type": "object",
                "dynamic": False,
                "properties": {
                    "_id": not_analyzed,
                    "guid": not_analyzed,
                    "unique_id": {"type": "integer"},
                },
            }
        },
    )
]

File: superdesk/types/archive.py [hunk 27]
```diff
@@ -0,0 +1,453 @@
...
+@dataclass
+class EntityMetadataItem:
+    name: str
+    qcode: str
+    scheme: str
+    source: str
+
+
+@dataclass
+class EntityMetadata:
```
MarkLark86: Is this correct, I don't see the `metadata` field in the original schema. The fields that use this seem to just be `list[EntityMetadataItem]`

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    alt_text: str | None = None
+    # FIXME: Place is slightly different ("location" is not a list here)
+    place: list[Place] | None = None
+    event: EntityMetadata | None = None
+    person: EntityMetadata | None = None
+    object: EntityMetadata | None = None
+    organisation: EntityMetadata | None = None
+    creditline: str
+    linked_in_packages: list[LinkedInPackage] | None = None
+    highlight: Annotated[str | None, validate_data_relation_async("highlights")] = None
```
MarkLark86: Should these use [fields.ObjectId](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L150), so serialisation works as expected.

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+
+
+@dataclass
+class EntityMetadata:
+    metadata: list[EntityMetadataItem] | None = None
+
+
+@dataclass
+class MarkedDesk:
+    desk_id: Annotated[str | None, validate_data_relation_async("desks")] = None
```
MarkLark86: Same here, should probably use fields.ObjectId. In fact, anything that uses `Resource.rel` in the old schema, where the `type` param is not changed, should use fields.ObjectId

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    lock_user: Annotated[str | None, validate_data_relation_async("users")] = None
+    lock_time: datetime | None = None
+    lock_session: Annotated[str | None, validate_data_relation_async("auth")] = None
+    lock_action: str | None = None
+    template: Annotated[str | None, validate_data_relation_async("content_templates")] = None
+    body_footer: str | None = None
+    flags: Flag | None = None
+    sms_message: str | None = None
+    format: str = FORMATS.HTML
+    auto_publish: bool
```
MarkLark86: This should probably have a default of `False`

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+@dataclass
+class Annotation:
+    id: int
+    type: str
+    body: str
+
+
+@dataclass
+class Attachment:
+    pass
```
MarkLark86: Missing `attachment` field, resource link to the `attachments` resource

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    # FIXME: duplicate of ItemSchema
+    broadcast: Broadcast | None = None
+    event_id: str
+    # FIXME: duplicate of ItemSchema
+    embargo: datetime | None = None
+    # FIXME: duplicate of ItemSchema
+    publish_schedule: datetime | None = None
+    # FIXME: duplicate of ItemSchema
+    schedule_settings: ScheduleSettings | None = None
+    used: bool
```
MarkLark86: This could probably be defaulted to `False`

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    used_count: int
```
MarkLark86: Could probably default to `0`, what do you think?

superdesk/types/archive.py
```diff
@@ -0,0 +1,453 @@
...
+    used_updated: datetime
+    metrics: dict[str, Any]
+    _type: str
```
MarkLark86: This is already defined in parent resource class, no need to re-define here

MarkLark86 (commented):
@jerome-poisson I have gone over the old schema and this one. There are some Elastic mapping differences between these models and the old ones. We need to make sure that the old elastic schema and new ones match, otherwise users may have search issues.

Maybe you can grab the elastic mapping of the old one, and create a unit test for this new one, to ensure the mappings are the same. Check [here](https://github.com/superdesk/superdesk-core/blob/async/tests/core/resource_model_test.py#L75) for an example test

superdesk/types/archive.py
jerome-poisson: There is `ContentAssociation` defined at `content_api/items/model.py` which does a similar thing, I'm using it.

jerome-poisson:
All review comments should have been addressed, there is now a base class `BaseContentItem` use for `ContentAPIItem` and `MetadataResource`.

MarkLark86 (approved):
Looks good to merge

PR #2832, opened 2025-02-14
[SDESK-7487] Add docstrings about usage of url in endpoints 

eos87:
### Purpose
~- Fix a bug that the `url_prefix` was not being added to the endpoint urls even when it was provided to the `EndpointGroup` definition.~

Update: I have found what was the problem on my local not picking up the the url prefix and api version. I turned this PR into just adding some docstrings regarding the usage of the `url_prefix` when registering an endpoint. 

[SDESK-7487]

[SDESK-7487]: https://sofab.atlassian.net/browse/SDESK-7487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/core/web/endpoints.py
+++ b/superdesk/core/web/endpoints.py
@@ -145,7 +145,7 @@ class EndpointGroup(EndpointGroupProtocol):
 
         def fdec(func: EndpointFunction):
             endpoint_func = Endpoint(
-                f"/{self.url_prefix}{url}" if self.url_prefix else url,
+                f"{self.url}/{url}" if self.url else url,
                 func,
                 methods=methods,
                 name=name,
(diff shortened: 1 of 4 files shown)


superdesk/core/web/endpoints.py
```diff
@@ -145,7 +145,7 @@ def endpoint(
 
         def fdec(func: EndpointFunction):
             endpoint_func = Endpoint(
-                f"{self.url}/{url}" if self.url else url,
+                f"/{self.url_prefix}{url}" if self.url_prefix else url,
```
MarkLark86: This might play up with the RestEndpoints, as it uses `url`. We can change the RestEndpoints to use `url_prefix` instead.

eos87: than means the `url_prefix` here is useless?

MarkLark86 (approved):

BrianMwangi21 (approved):

PR #2833, opened 2025-02-17
fix: Register endpoints as well with root api response

MarkLark86:
### Purpose
Enhance the API root endpoint by registering the async endpoints to API root for resources discovery.

### What has changed
- Modified the `extend_eve_home_endpoint` method to register the async endpoints


eos87 (approved):
I did some minor changes and tested this locally. It works as expected.

PR #2836, opened 2025-02-19
refactor websockets handling

petrjasek:
- drop unused server loop
- drop pings (it's handled via websockets library)
- setup sentry_sdk

SDCP-879


MarkLark86 (approved):
It looks fine to me, though I guess this is something that just needs to be tested rather than peer reviewed.

Does the newer sentry-sdk version work with our version of Sentry v9.0, or would we need to update that too?

petrjasek:
it does not, already talked with infra about that, there will be a new instance with up to date version

PR #2840, opened 2025-02-24
Async publish consumer

MarkLark86:
### Purpose
Implement the PublishConsumer side of the new publishing system.

### What has changed
* Added a new typed version of getting app config, i.e. `get_config(int, "SOME_SETTING")`
* Created a `SingletonInstance` base class, for easy re-use
* Created a `load_class_from_config` helper, that loads the class type based on key from app config
* Add new PublishConsumers (asyncio & celery)
* Remove old publish transmit code
* Update docs


MarkLark86:
@petrjasek @eos87 docs can be found on https://superdesk--2840.org.readthedocs.build/en/2840/publish.html

superdesk/core/resources/model.py
```diff
@@ -343,7 +343,7 @@ class ResourceModelWithObjectId(ResourceModel):
     """Base ResourceModel class to be used, if the resource uses an ObjectId for it's ID"""
 
     #: ID of the document
-    id: Annotated[ObjectId, Field(alias="_id", default_factory=ObjectId)]
+    id: ObjectId = Field(alias="_id", default_factory=ObjectId)
```
eos87, line 346: What's the reason for this change? (just curious)

superdesk/publish_async/commands.py
```diff
@@ -0,0 +1,37 @@
...
+    Asynchronous task for processing pending exchange tasks with specific
+    time constraints using Celery. This task is set up to have a soft time
+    limit and an expiration time, ensuring efficient task execution and
+    preventing runaway processes.
+
+    Args:
+        None
+
+    Returns:
+        None
```
eos87, line 34: nit: probably not needed?
```suggestion
```

superdesk/publish_async/controller/consumers/asyncio_consumer.py
```diff
@@ -0,0 +1,138 @@
...
+            logger.info(f"Transmit State is not pending/retrying for queue item {task.id}. Is in in {task.state}")
+            return False
+        elif task.destination is None:
+            logger.error(f"Destination not defined in queue item {log_msg}")
+            return False
+
+        try:
+            # Update the status of the task to in-progress
+            task_update = {"state": PublishQueueState.IN_PROGRESS, "transmit_started_at": utcnow()}
+            await PublishQueueResource.get_service().update(task.id, task_update, task.etag, task)
```
eos87, line 92: I feel like we should try to stick to using the async service class directly whenever is possible as `.get_service()` is not clear right away when reading. What do you think?

superdesk/publish_async/controller/consumers/asyncio_consumer.py
```diff
@@ -0,0 +1,138 @@
...
+                await response
+            logger.info(f"Transmit completed for queue item {log_msg}")
+            return True
+        except Exception as e:
+            logger.exception(f"Failed to transmit queue item {log_msg}")
+
+            max_retry_attempt = get_config(int, "MAX_TRANSMIT_RETRY_ATTEMPT")
+            retry_attempt_delay = get_config(int, "TRANSMIT_RETRY_ATTEMPT_DELAY_MINUTES")
+            try:
+                timeout = 2 ** min(6, task.retry_attempt or retry_attempt_delay)
```
eos87, line 115: I'm wondering if we should put the `2` in a setting too so it is also customizable and gets some meaning. Thoughts?

eos87 (approved):
It looks good to me. I just a couple of minor comments/questions.

superdesk/core/resources/model.py
MarkLark86: mypy and my IDE were complaining when I was creating an instance of `ResourceModelWithObjectId` without specifying an `id`

superdesk/publish_async/controller/consumers/asyncio_consumer.py
MarkLark86: I left a comment on Brian's Planning PR about the use of `Model.get_service`. ALL ResourceModel's have this method, and provides access to the data management service without needing to import it (reduce cyclic imports)

superdesk/publish_async/controller/consumers/asyncio_consumer.py
MarkLark86: We will be looking at this in the future, as we can improve how we handle this scenario better with the new code. Petr and I will review what is possible what's all publish code is updated

superdesk/publish_async/commands.py
MarkLark86: I'll update these in my next PR (to use RST as that is what has been historically used).
This was populated by AI - I will add args to it so it provides RST in the future

petrjasek:
looks good to me as well, just have to wrap my head around that async consumer, what happens in case the server would crash while processing an item, but I suppose with the publish queue it can work

petrjasek (approved):

PR #2845, opened 2025-02-25
[Content Types] Create and use async resource and service

jerome-poisson:
This patch create Async ContentTypes resource, and an async service which ports methods from the legacy ones.

SDESK-7482


superdesk/types/content_types.py
```diff
@@ -0,0 +1,45 @@
...
+logger = logging.getLogger(__name__)
+
+
+class WidgetConfig(ResourceModel):
+    widget_id: str
+    is_displayed: bool
+
+
+class ContentTypes(ResourceModel):
+    type_: Annotated[str | None, Field(alias="type")]
```
MarkLark86: The original has the [content_type_single_item_type](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/validator.py#L297) validation. It would be good to implement one for async as well (can be defined in this file only, as it would only be used here).

```python
from pydantic_core import PydanticCustomError
from quart_babel import gettext

from superdesk.core.resources.validators import AsyncValidator

async def _validate_content_type(item: ContentTypes):
    if not item.content_type or item.content_type.lower() == "text":
        return
    if await ContentTypes.get_service().count({"type": item.content_type}) > 0:
        raise PydanticCustomError("unique", gettext("Only 1 instance is allowed."))

validate_content_type = AsyncValidator(_validate_content_type)
```

And then on the model, annotate using that validator:
```python
content_type: Annotated[str | None, validate_content_type, Field(alias="content_type")]
```

**Note:** I have used `content_type` instead of `type_`, as the name then gives context as to what the value of this field is used for

superdesk/types/content_types.py
```diff
@@ -0,0 +1,45 @@
...
+    label: Annotated[str, validate_iunique_value_async("content_types", "label")]
+    icon: str
```
MarkLark86: Fields like `icon` that are not mandatory could be marked optional using `str | None = None`

MarkLark86: Otherwise validation would fail when getting items from the DB that don't have a value defined here

superdesk/types/content_types.py
```diff
@@ -0,0 +1,45 @@
...
+    description: str
+    schema: dict[str, Any] = Field(default_factory=dict)  # type: ignore[assignment]
+    editor: dict[str, Any] = Field(default_factory=dict)
+    widgets_config: list[WidgetConfig] = Field(default_factory=list)
+    priority: int = 0
+    enabled: bool = False
+    is_used: bool = False
+    embeddable: bool = False
+    created_by: str | None = None
```
MarkLark86: This is where you would use [validate_data_relation_async("users")](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/validators.py#L143) for these next 2

superdesk/types/content_types.py
```diff
@@ -0,0 +1,45 @@
...
+    updated_by: str | None = None
+    init_version: int
```
MarkLark86: As this is not mandatory, we should probably use `int | None = None` as well here

superdesk/content_types_async/service.py
```diff
@@ -0,0 +1,581 @@
...
+    async def on_create(self, docs: list[ContentTypes]) -> None:
+        for doc in docs:
+            doc.updated_by = await get_user_id()
+            await self._set_created_by(doc)
+
+    async def on_delete(self, doc: ContentTypes) -> None:
+        if doc.is_used:
+            raise SuperdeskApiError(status_code=202, payload={"is_used": True})
+        await remove_profile_from_templates(doc)
+        # FIXME: not async yet
```
MarkLark86: This is where we use the convention `TODO-ASYNC:`, that way we can search all comments starting with this to see what is outstanding, and when it could be done. Please update all comments like this from `# FIXME: ` to `# TODO-ASYNC: `

superdesk/content_types_async/service.py
```diff
@@ -0,0 +1,581 @@
...
+# cvs hardcoded in the app wich special use
+# and not supposed to be added to content profile
+HARDCODED_CVS = ("languages",)
+
+
+class ContentTypesService(AsyncCacheableService[ContentTypes]):
+    resource_name = "content_types"
+
+    async def _set_created_by(self, doc: ContentTypes) -> None:
+        doc.created_by = await get_user_id()
```
MarkLark86: Is `get_user_id` async yet, otherwise this would raise an error stating can't await as the function is not a coroutine

superdesk/content_types_async/service.py
```diff
@@ -0,0 +1,581 @@
...
+
+    async def on_update(self, updates: dict[str, Any], original: ContentTypes) -> None:
+        await self._validate_disable(updates, original)
+        updates["updated_by"] = await get_user_id()
+        await prepare_for_save_content_type(original, updates)
+        await self._update_template_fields(updates, original)
+
+    async def on_delete_res_vocabularies(self, doc: dict[str, Any]) -> None:
+        projection: dict[str, Literal[1]] = {"label": 1}
+        cursor = await self.find({"schema." + doc[ID_FIELD]: {"$type": 3}}, projection=projection)
```
MarkLark86: `projection` can also simply be a list, i.e. `projection=["label"]`, which would not require the `Literal` typing from above

superdesk/content_types_async/service.py
```diff
@@ -0,0 +1,581 @@
...
+        if cursor.count():
+            message = _("Vocabulary {vocabulary} is used in {count} content type(s)").format(
+                vocabulary=doc.get("display_name"), count=cursor.count()
```
MarkLark86: [cursor.count](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/cursor.py#L54) is async, would need to await it's response. Might also be good to call it once, store the result locally and use that in the `if` statement and this message string

superdesk/content_types_async/service.py
MarkLark86: Is `get_user_id` async yet, otherwise this would raise an error stating can't await as the function is not a coroutine

superdesk/content_types_async/service.py
```diff
@@ -0,0 +1,581 @@
...
+from copy import deepcopy
+from typing import Any, Literal, cast
+
+import bson
+from bson import ObjectId
+from quart_babel import gettext as _
+
+from apps.auth import get_user_id
+from apps.desks import remove_profile_from_desks
+from apps.desks_async.desks_async_service import DesksAsyncService
```
MarkLark86: Rather than importing the service directly, which might cause cyclic imports in the end, you can import the ResourceModel and use that instead. i.e.

```python
from superdesk.types import DesksResourceModel

async def func_test():
    all_desks = await DesksResourceModel.get_service().get_all_list()
```

MarkLark86 (commented):
Mostly looks fine. Just a few improvements and a couple bugs need looking into

superdesk/types/content_types.py
jerome-poisson: Is there any indicator to know which ones are optional? I've set the `|None = None` only when `required: False` or `nullable: True` was set on the legacy schema, but on some fields like `icon` here, there is nothing.

MarkLark86: It's actually the other way around, by default fields are not required unless stated otherwise.
Unfortunately in Pydnatic we don't have the ability to have `nullable: False` AND `required: False`, as Pydantic fields always need a value.

So I would say, unless the old schema has `required: True`, then we can assume it's an optional field

superdesk/content_types_async/module.py
```diff
@@ -0,0 +1,23 @@
...
+from superdesk.core.resources import ResourceConfig, MongoResourceConfig, MongoIndexOptions
+from superdesk.types.content_types import ContentTypes
+from .service import ContentTypesService
+
+
+content_types_resource_config = ResourceConfig(
+    name="content_types",
+    data_class=ContentTypes,
+    service=ContentTypesService,
+    mongo=MongoResourceConfig(
```
MarkLark86, line 21: Missing `rest_endpoints` config, otherwise no REST API will be exposed for this resource. Also missing the `default_sort` config

```python
from superdesk.core.resources import RestEndpointConfig
from superdesk.core.auth.privilege_rules import http_method_privilege_based_rules

CONTENT_TYPE_PRIVILEGE = "content_type"

content_types_resource_config = ResourceConfig(
    ...
    default_sort=[("priority", -1)],
    rest_endpoints=RestEndpointConfig(
        id_param_type=r'regex("[\w,.:-]+")',
        auth=http_method_privilege_based_rules(
            {
                "POST": CONTENT_TYPE_PRIVILEGE,
                "PATCH": CONTENT_TYPE_PRIVILEGE,
                "DELETE": CONTENT_TYPE_PRIVILEGE,
            }
        ),
    ),
)
```

superdesk/types/content_types.py
jerome-poisson: OK, good to know, thanks.

MarkLark86 (approved):

PR #2848, opened 2025-03-04
Support async in old eve mongo resource services

MarkLark86:
Async Elasticsearch support to be added later


superdesk/async_resource/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .mongo_async_datalayer import MongoAsync
```
petrjasek: I wonder if we need the async in all the names, would be enough for me to see it once:
`from superdesk.async.resource import Mongo` or
`from.superdesk.async.resource import AsyncMongo` if needed
but would put it everywhere in the namespace

petrjasek (approved):

superdesk/async_resource/__init__.py
MarkLark86: Cannot use `async` as a module name, conflicts with the `async` statement. Changed to `eve_async` to give context to the name

PR #2853, opened 2025-03-09
[SDESK-7497] Small adjustments for `superdesk-planning` behave tests

eos87:
Small changes required for `superdesk-planning` behave tests to pass. Some of them are just for debugging.

[SDESK-7497]

[SDESK-7497]: https://sofab.atlassian.net/browse/SDESK-7497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -242,9 +242,12 @@ class BasePublishService(BaseService):
                 # race condition.
                 published_service = get_resource_service(PUBLISHED)
                 assert published_service is not None
+
+                # TODO-ASYNC: superdesk-planning tests break because of difference between `_id` and `guid`
                 published_service.patch(id, {QUEUE_STATE: PUBLISH_STATE.PUSHED})
                 from apps.publish.enqueue import push_publish
 
+                # TODO-ASYNC: update this service so task below completes eagearly and tests pass
                 push_publish.apply_async(str(id))
 
             push_notification(
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -243,7 +243,6 @@ class BasePublishService(BaseService):
                 published_service = get_resource_service(PUBLISHED)
                 assert published_service is not None
 
-                # TODO-ASYNC: superdesk-planning tests break because of difference between `_id` and `guid`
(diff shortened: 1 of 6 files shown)


apps/publish/published_item.py
```diff
@@ -137,6 +137,17 @@ class PublishedItemService(BaseService, HighlightsSearchMixin):
...
     SEQ_KEY_NAME = "published_item_sequence_no"
 
+    def patch(self, id, updates):
+        """
+        FIXME: Temporary workaround until we find out why _id and guid are not in sync
+        when running tests from `superdesk-planning`. Without this, many tests fill fail.
+        """
+        item = self.find_one(req=None, guid=id)
+        if item:
+            id = item[ID_FIELD]
```
eos87: I could not find the root of this problem unfortunately. I just added this workaround with the full intention of finding why are the `_id` and `guid` out of sync. So far what I've found is that when the `published` item is created, the values are fine, but apparently there is some action happening in between that is modifying the `_id` with an `ObjectId` value.

MarkLark86: @eos87 The `_id` field on the `published` collection uses an ObjectId. There could be multiple entries for the one item (multiple versions). What we use instead is the `item_id` field, as that is what matches the `_id` or `guid` from the `archive` collection (and other content based collections).

Where specifically in the Planning module is this causing an issue?

eos87: Removed as we spoke earlier today.

apps/publish/content/common.py
```diff
@@ -242,9 +242,12 @@ def update(self, id, updates, original):
                 # race condition.
                 published_service = get_resource_service(PUBLISHED)
                 assert published_service is not None
+
+                # TODO-ASYNC: superdesk-planning tests break because of difference between `_id` and `guid`
```
MarkLark86: These 2 comments aren't needed, you can see the changes in my PR here:

https://github.com/superdesk/superdesk-core/blob/32fb1ba9c9a798b01a431d153b533983675443ff/apps/publish/content/common.py#L295-L308

MarkLark86 (approved):
Looks good

PR #2855, opened 2025-03-12
Publish exchange system

MarkLark86:


MarkLark86:
@petrjasek @eos87 Docs can be found [here](https://superdesk--2855.org.readthedocs.build/en/2855/publish.html)

superdesk/publish_async/exchanges/content_exchange.py
```diff
@@ -0,0 +1,232 @@
...
+
+            response = await super().send(request)
+
+            if not response.content_api_subscribers and request.publish_to_content_api and content_api.is_enabled():
+                try:
+                    # If there were no ContentAPI Subscribers, we push it there manually now
+                    get_resource_service("content_api").publish(request.item, [])
+                except Exception:
+                    logger.exception(
+                        f"Failed to queue item to API for item: {request.item_id} for action {request.operation}"
```
petrjasek: would be good to avoid unique log messages so sentry can group those, you can put item_id and operation in `extra` param

superdesk/publish_async/filters/base_exchange_filter.py
```diff
@@ -0,0 +1,546 @@
...
+        product_ids: set[ObjectId] = set(
+            chain(*[subscriber.products for subscriber in subscribers if subscriber.products])
+        )
+        api_product_ids: set[ObjectId] = (
+            set()
+            if not request.publish_to_content_api or not content_api.is_enabled()
+            else set(chain(*[subscriber.api_products for subscriber in subscribers if subscriber.api_products]))
+        )
+
+        response.matched_products = self.get_matching_products(request, product_ids)
```
petrjasek, line 115: I think it would be useful to store matched products on the item itself, we could later display how many items were published per product, and it could be also used by formatters

petrjasek (commented):
looks good to me, would be good to get a review from someone more async-ish

PR #2856, opened 2025-03-14
[SDESK-7482] fix ContentProfile async resource

MarkLark86:
### Purpose
Managing ContentTypes (aka ContentProfiles) through the front-end was not possible. The front-end code does not like when editor/schema keys have a `null` value, so we fix it up in the service.

### What has changed
* Connect `ContentType.data.on_created` signal to a function to create a Content Template
* Rename `schema` field on `ContentTypes` to an alias `content_schema` (due to clashes in Pydantic schema method)
* Make `description`, `content_expiry`, `source` and `language` optional fields in DeskResourceModel
* Make `description` optional in ContentType
* Use `ObjectId` for `created_by` and `updated_by` fields on ContentType
* Fix `ContentType.type` validation to exclude current item being checked

### DIFF
--- a/apps/templates/__init__.py
+++ b/apps/templates/__init__.py
@@ -12,10 +12,11 @@ from typing import Any
 import superdesk
 from superdesk import register_jinja_filter
 from superdesk.text_utils import get_text
+from superdesk.types import ContentTypes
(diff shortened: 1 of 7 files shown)


MarkLark86:
@jerome-poisson I have fixed up a few things, tested it locally and it seems to work now. Can you please review, and assign to QA once it's merged

jerome-poisson (approved):

PR #2860, opened 2025-03-19
[SDESK-7539] Upgrade Users & Preferences eve resource to async services

MarkLark86:
### Purpose
Upgrade the User & Preferences eve resources to async services, and update it's usage where possible.

### What has changed
* Converted User eve resource to async resource
* Converted Preferences eve resource to async resource
* Updated code to use UsersResourceModel where possible
* Placed TODO-ASYNC comments for future tasks

Extends from https://github.com/superdesk/superdesk-core/pull/2859 (the first few commits you can ignore reviewing from this PR)

### DIFF
--- a/apps/desks.py
+++ b/apps/desks.py
@@ -314,7 +314,7 @@ class DesksService(AsyncBaseService):
         :return dict: desk document
         """
         desk_name = ""
-        desk = get_resource_service("desks").find_one(req=None, _id=desk_id)
+        desk = self.find_one(req=None, _id=desk_id)
         if desk:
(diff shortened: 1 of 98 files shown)


BrianMwangi21 (approved):

apps/desks.py
```diff
@@ -246,6 +246,19 @@ async def on_delete_async(self, desk):
                 message=_("Cannot delete desk as it has article(s) or referenced by versions of the article(s).")
             )
 
+<<<<<<< HEAD
+=======
```
petrjasek: those merge tags should go away

PR #2868, opened 2025-03-26
[SDESK-7490][Search Providers] Make Search Providers Async

jerome-poisson:


jerome-poisson:
I'll switch this over after all the customer search providers are converted to async; they will be handled in a follow-up ticket. Client testing will also be done then. I'm just a bit hesitant about the `search_providers_proxy` endpoint, especially the `POST` handling. The legacy code's lack of type hints and documentation makes it a bit unclear.

apps/io/search_ingest.py
```diff
@@ -106,3 +110,89 @@ def _get_query(self, req):
         args = getattr(req, "args", {})
         query = json.loads(args.get("source")) if args.get("source") else {"query": {"filtered": {}}}
         return query
+
+
+class SearchIngestServiceAsync(AsyncBaseService):
```
MarkLark86, line 116: Is there a reason why you've kept the original `SearchIngestService` and added this one, instead of upgrading that service to async?

apps/io/search_ingest.py
```diff
@@ -106,3 +110,89 @@ def _get_query(self, req):
...
+            if not doc.get("desk"):
+                # if no desk is selected then it is bad request
+                raise SuperdeskApiError.badRequestError(_("Destination desk cannot be empty."))
+            try:
+                archived_doc = await self.fetch_async(doc["guid"])
+            except FileNotFoundError as ex:
+                raise ProviderError.externalProviderError(ex, provider)
+
+            # TODO-ASYNC: fetch_item is not async
+            dest_doc = fetch_item(archived_doc, doc.get("desk"), doc.get("stage"), state=doc.get("state"))
```
MarkLark86: [fetch_item](https://github.com/superdesk/superdesk-core/blob/async/apps/archive/common.py#L467) is now async. Might be good to pull the latest, fix merge conflicts

MarkLark86 (commented):
Looks fine. Just a rebase etc. I suppose it's one of those things that we muse upgrade a client repo/search provider to test

apps/io/search_ingest.py
jerome-poisson: yes, I want to keep the old code working until I do the full switch, but I want to test with few async search providers first.

MarkLark86 (approved):
Looks good

PR #2869, opened 2025-03-27
Update the metadata stored in the binaries not only in the original image, but also in all renditions [SDCP-910]

devketanpro:


apps/publish/content/common.py
```diff
@@ -928,31 +928,36 @@ def _inherit_publish_schedule(self, original, updates, associated_item):
...
         if not mapping or not renditions:
             return
         try:
-            media_id = renditions["original"]["media"]
-        except (KeyError, TypeError):
-            return
-        if not media_id:
-            return
+            for rendition_key, rendition_data in renditions.items():
+                media_id = rendition_data.get("media")
```
petrjasek, line 943: `rendition_data` could be null if some rendition can't be generated due to size, would be good to check that

apps/publish/content/common.py
```diff
@@ -928,31 +928,36 @@ def _inherit_publish_schedule(self, original, updates, associated_item):
...
+                picture = app.media.get(media_id)
+                binary = picture.read()
+                metadata = get_metadata_from_item(updated, mapping)
+
+                updated_binary = write_metadata(binary, metadata)
+                if updated_binary != binary:
+                    updated_media_id = app.media.put(
+                        updated_binary, content_type=picture.content_type, filename=picture.filename
+                    )
+                    updates.setdefault("renditions", deepcopy(renditions))[rendition_key].update(
```
petrjasek: would be good to do that `setdefault` and `deepcopy` just once

petrjasek (commented):
test pls

petrjasek (approved):

PR #2874, opened 2025-04-09
[SDESK-7557] Implement async media I/O

MarkLark86:
### Purpose
Implement async Mongo GridFS / Amazon S3 interfaces to the `app.media` object. Implemented in the `release/3.0` branch so it can be used with the Newshub `release/3.0` branch/release

### What has changed:
* Remove previous async GridFS implementation (in favour of the `app.media` object)
* Add `aioboto3` to libs (it also uses `aiohttp` which is also now used/available)
* Implement async interface for Mongo GridFS
* Implement async interface for Amazon S3
* Use the new `app.media` async methods where currently possible
* Fix `get_headers` signature for ASGI (includes scope as well as env)
* Fix lint and tests


superdesk/storage/amazon_media_storage.py
```diff
@@ -248,28 +364,68 @@ def put(
...
+        version=True,
+        folder=None,
+        **kwargs,
+    ):
+        """Save a new file using the storage system, preferably with the name specified.
 
-        found = self._check_exists(_id)
-        if found:
-            return _id
+        If there already exists a file with this name name, the
```
petrjasek: there is `name` 2x

petrjasek (approved):

PR #2894, opened 2025-04-28
Update the metadata stored in the binaries not only in the original image, but also in all renditions [SDCP-910]

devketanpro:


devketanpro:
will add tests

apps/publish/content/common.py
```diff
@@ -895,8 +895,24 @@ def _publish_associated_items(self, original, updates=None):
                         sync_associated_item_changes(associated_item, associated_item_updates)
                         continue
 
-                    if association_updates.get("state") not in PUBLISH_STATES:
-                        # There's an update to the published associated item
+                    if updates.get("state"):
```
petrjasek: isn't this always true?

apps/publish/content/common.py
```diff
@@ -895,8 +895,10 @@ def _publish_associated_items(self, original, updates=None):
                         sync_associated_item_changes(associated_item, associated_item_updates)
                         continue
 
-                    if association_updates.get("state") not in PUBLISH_STATES:
-                        # There's an update to the published associated item
+                    if (
+                        app.config.get("ENABLE_ASSOCIATED_ITEMS_UPDATE")
+                        or association_updates.get("state") not in PUBLISH_STATES
```
petrjasek: I wouldn't work with a new config, I think it should be based on changes in the `association_updates` values,
imo that should be standard behaviour.

features/content_publish.feature
```diff
@@ -4965,3 +4979,118 @@ Feature: Content Publishing
...
+      When we publish "5678" with "correct" type and "corrected" state
+      """
+      {
+          "headline": "corrected story headline",
+          "correction_sequence": "2",
+          "associations": {
+            "picture": {
+              "_id": "1234",
+              "slugline": "corrected picture",
+              "state": "corrected",
```
petrjasek: would UI actually send different state in this case?

apps/publish/content/common.py
```diff
@@ -895,8 +895,7 @@ def _publish_associated_items(self, original, updates=None):
                         sync_associated_item_changes(associated_item, associated_item_updates)
                         continue
 
-                    if association_updates.get("state") not in PUBLISH_STATES:
-                        # There's an update to the published associated item
+                    if association_updates.get("state") not in PUBLISH_STATES or updates.get("state"):
```
petrjasek: I would rather compare the data, eg headline, slugline, etc. Imo this is again always true.
what we want to do is publish the association_updates if there are any changes, but not if there were no changes, updating that image metadata and republishing all images is kind of expensive

features/content_publish.feature
devketanpro: nope, I removed this

apps/publish/content/common.py
```diff
@@ -895,14 +895,31 @@ def _publish_associated_items(self, original, updates=None):
...
+                    orig_associated_item = archive_service.find_one(req=None, _id=associated_item[config.ID_FIELD])
+                    if self.is_changed(orig_associated_item, associated_item):
                         remove_unwanted(association_updates)
                         publish_service.patch(id=associated_item[config.ID_FIELD], updates=association_updates)
 
             # When there is an associated item which is published, Inserts the latest version of that associated item into archive_versions.
             insert_into_versions(doc=associated_item)
         self._refresh_associated_items(original)
 
+    def is_changed(self, old: dict, new: dict) -> bool:
```
petrjasek, line 909: we would need it also check the extra fields, and we now use profile for picture type so I would just look for any changes in fields not stating with `_`

apps/publish/content/common.py
```diff
@@ -895,15 +895,40 @@ def _publish_associated_items(self, original, updates=None):
...
+        SIGNIFICANT_FIELDS = {"headline", "slugline", "byline", "description_text"}
+
+        if len(old) != len(new):
+            return True
+
+        item_type = old.get("type") or new.get("type")
+        schema_config = app.config.get("SCHEMA", {})
+        schema_fields = set(schema_config.get(item_type, {}).keys())
+
+        extra_fields = {key for key in old.keys() | new.keys() if not key.startswith("_") and key not in schema_fields}
```
petrjasek: extra fields should be checked all probably, with `_` prefix I was talking about fields on top level, eg. `_created` `_etag` etc

apps/publish/content/common.py
```diff
@@ -895,15 +895,40 @@ def _publish_associated_items(self, original, updates=None):
...
+        Return True if any significant field has changed.
+        """
+        SIGNIFICANT_FIELDS = {"headline", "slugline", "byline", "description_text"}
+
+        if len(old) != len(new):
+            return True
+
+        item_type = old.get("type") or new.get("type")
+        schema_config = app.config.get("SCHEMA", {})
+        schema_fields = set(schema_config.get(item_type, {}).keys())
```
petrjasek: we are not really setting this anymore, it's part of respective content profile. but I would just check everything besides fields starting with `_`

apps/publish/content/common.py
```diff
@@ -895,15 +895,32 @@ def _publish_associated_items(self, original, updates=None):
                         sync_associated_item_changes(associated_item, associated_item_updates)
                         continue
 
-                    if association_updates.get("state") not in PUBLISH_STATES:
-                        # There's an update to the published associated item
+                    orig_associated_item = archive_service.find_one(req=None, _id=associated_item[config.ID_FIELD])
+                    if self.is_changed(orig_associated_item, associated_item):
```
petrjasek: I wonder if it would publish media item in progress if there were no changes

devketanpro: should we explicitly skip calling `publish_service.patch(...`) for associated items that are still in `in_progress` (or other non-published) states?

petrjasek: I think the issue might be rather the other way around, that it wouldn't get called when it should be.
maybe we should have something like this:
```py
 if association_updates.get("state") not in PUBLISH_STATES or self._is_changed(...):
 ```
but I wonder if it actually gets there if no tests were failing

Comment (devketanpro) [reply apps/publish/content/common.py]: makes sense. do you have a specific test scenario in mind that we might be missing here? I can add that case to the tests and we’ll see if the current logic gets triggered and behaves as expected

Comment (petrjasek) [reply apps/publish/content/common.py]: just publishing an item with image in progress without any changes made during publishing

Comment (devketanpro) [reply apps/publish/content/common.py]: added a new test case

File: apps/publish/content/common.py [hunk 9]
```diff
@@ -895,15 +895,34 @@ def _publish_associated_items(self, original, updates=None):
...
             insert_into_versions(doc=associated_item)
         self._refresh_associated_items(original)
 
+    def is_changed(self, old: dict, new: dict) -> bool:
+        """
+        Compare all top-level fields except those starting with an underscore (_).
+        Return True if any such field has changed.
+        """
+
+        if len(old) != len(new):
```
petrjasek: would also check somewhere `if old is not None` because that would break the publishing

petrjasek (approved):

PR #2895, opened 2025-04-28
[SDESK-7582] Remove old publish system

MarkLark86:
### Purpose
Remove old publish system and it's eve resources that have been replaced with a Pydantic async resource.


apps/publish/publish_service_tests.py
```diff
@@ -21,115 +23,135 @@
...
-        self.app.data.insert("publish_queue", self.queue_items)
+        self.subscriber_service = SubscribersResource.get_service()
+        self.queue_service = PublishQueueResource.get_service()
 
-        init_app(self.app)
+        await self.subscriber_service.create(self.subscribers)
+        await self.queue_service.create(self.queue_items)
+
+        # TODO-ASYNC-PR: Is this ``init_app`` needed?
+        # init_app(self.app)
```
eos87: probably not

apps/publish/published_item.py
```diff
@@ -132,28 +135,29 @@ class PublishedItemResource(Resource):
     }
 
 
-class PublishedItemService(BaseService, HighlightsSearchMixin):
+# TODO-ASYNC-PR: Upgrade this to async
```
eos87: Is this comment still needed? it seems to me that the service is already async.

eos87 (approved):
I don't fully understand all the logic here, but code wise looks good.

PR #2901, opened 2025-04-30
[SDESK-7586] Upgrade system resources to async eve services

MarkLark86:
### Purpose
Upgrade system based resources to use async eve services

### What has changed
* Upgraded `languages` resource to async eve
* Upgraded `system_messages` resource to async eve
* Upgraded `activity` resource to async eve
* Upgraded `audit` resource to async eve
* Upgraded `config` resource to async eve
* Fixed `limit` attribute of the AsyncListCursor
* Moved `get_items` and `get_languages` out of the Pydantic service to a util file

### What was **not** changed
* `client_config` - no I/O to wait for
* `backend_meta` - uses local file input, endpoint not used often
* `data_updates` - only used by backend commands

**Note:** Commented out notifier on Ingest/Publish error, due to requiring async but the code runs inside an Exception constructor. Will do this task in a separate ticket, as it will be much bigger of a change. (see usage of [update_notifiers](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/errors.py#L31))

### DIFF
--- a/apps/archive/archive.py
+++ b/apps/archive/archive.py
@@ -414,6 +414,8 @@ class ArchiveService(BaseService, HighlightsSearchMixin):
                 msg = 'added new {{ type }} item about "{{ subject }}"'
             else:
                 msg = "added new {{ type }} item with empty header/title"
+
+            # TODO-ASYNC[activity]: Prefix this next line with `await ` when updating this module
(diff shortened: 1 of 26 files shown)


eos87 (approved):
Looks good to me

PR #2912, opened 2025-05-06
[SDESK-7619] Update Ingest/Publish exceptions to support async notifications

MarkLark86:
### Purpose
Original when an exception was raised during an Ingest or Publish cycle, a document would be added to the `activity` resource and email(s) would be sent off. This happened inside the constructor of the Exceptions.

Now that we have async `activity` and emails, we can no longer create these activity docs or send emails in the constructor, where async code is not supported.

### What has changed
* Add a new method for these Exceptions that creates the `activity` resource and sends email notifications
* Update ingest code where these exceptions are raised to send the notifications
* Update publish code where these exceptions are raised to send the notifications
* Update tests

### DIFF
--- a/superdesk/errors.py
+++ b/superdesk/errors.py
@@ -31,9 +31,9 @@ def add_notifier(notifier):
 
 async def update_notifiers(*args, **kwargs):
     for notifier in notifiers:
-        resposne = notifier(*args, **kwargs)
-        if isawaitable(resposne):
-            await resposne
+        response = notifier(*args, **kwargs)
+        if isawaitable(response):
+            await response
 
 
 def get_registered_errors(self):
--- a/superdesk/errors.py
+++ b/superdesk/errors.py
@@ -8,6 +8,8 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
+from typing_extensions import Self
+from inspect import isawaitable
 import logging
 
 from cerberus import DocumentError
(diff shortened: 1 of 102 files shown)


superdesk/errors.py
```diff
@@ -27,9 +29,11 @@ def add_notifier(notifier):
         notifiers.append(notifier)
 
 
-def update_notifiers(*args, **kwargs):
+async def update_notifiers(*args, **kwargs):
     for notifier in notifiers:
-        notifier(*args, **kwargs)
+        resposne = notifier(*args, **kwargs)
+        if isawaitable(resposne):
+            await resposne
```
eos87: nit typo:
```suggestion
        response = notifier(*args, **kwargs)
        if isawaitable(response):
            await response
```

eos87 (approved):
Looks good. Left one minor comment regarding a typo. 

Just a quick thought: couldn't we have something like an exception catcher and then send the notifications based in the information collected in the exception? that way we would not need to call `.send_notifications` everywhere and could also be async... something a la Sentry style. What do you think?

MarkLark86:
> Looks good. Left one minor comment regarding a typo.
> 
> Just a quick thought: couldn't we have something like an exception catcher and then send the notifications based in the information collected in the exception? that way we would not need to call `.send_notifications` everywhere and could also be async... something a la Sentry style. What do you think?

That sounds like a good idea. Would rather look into this at a later time as a general error/exception handling improvement. The ingest/publish needed to be updated to async anyways, which is what most of the changes in the PR does.

PR #2914, opened 2025-05-07
improve: Publish util to get publish subscribers for item

MarkLark86:
### Purpose
To expose a util function that provides the ability to get the list of subscribers for an item publish action


eos87 (approved):
Looks good!

PR #2915, opened 2025-05-07
Fix various async tests

MarkLark86:
### Purpose
Fix various async issues found in the code, get more tests to pass.

### What has changed:
* Re-using DB connections between tests was causing an issue, due to the event loop changing and DB client expecting the same loop. Instead We'll close any previous connections and open up new ones for every class test instance
* Fix various async cursor and await issues found
* Fix more unit tests


MarkLark86:
Getting there:
```
Pytest : 1219 passed,  45 failed, 26 skipped
Behave :  635 passed, 295 failed,  4 skipped
ProdApi:    5 passed,  18 failed,  0 skipped
-------------------------------------------
Total  : 1859 passed, 358 failed, 30 skipped
```

eos87 (approved):
👏

PR #2922, opened 2025-05-13
[SDESK-7623] - Filter condition creation required fields are empty

BrianMwangi21:
### Purpose
This PR fixes a bug that cause fields field to be empty on filter condition creation

## Changes made
1. Created `filter_conditions/parameters` endpoint to return Filter Condition Parameters

## **NOTE**
Still WIP. Trying to figure out why the endpoint is not being called by the client-core here : https://github.com/superdesk/superdesk-client-core/blob/0d1567f1f379998cb2ab8ab301e8686dbb81a8dc/scripts/apps/content-filters/services/ContentFiltersService.ts#L17

Solves: [SDESK-7623](https://sofab.atlassian.net/browse/SDESK-7623)

[SDESK-7623]: https://sofab.atlassian.net/browse/SDESK-7623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrianMwangi21:
Closed this as it was done in this [PR](https://github.com/superdesk/superdesk-core/pull/2925) to avoid duplication.

PR #2923, opened 2025-05-13
fix saml auth on async

petrjasek:
SDESK-7618

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -13,6 +13,7 @@ pytest-env==1.1.5
 python3-saml==1.16.0
 moto[sqs]==5.1.3
 pyexiv2==2.15.3; sys_platform == 'linux'
+python3-saml==1.16.0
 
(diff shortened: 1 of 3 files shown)


petrjasek:
with those changes it works locally @MarkLark86 , but I'm wondering if we need to ditch those imports from flask or it's ok?

MarkLark86:
> with those changes it works locally @MarkLark86 , but I'm wondering if we need to ditch those imports from flask or it's ok?

Importing `flask` works because of the `quart-flask-patch` module we have installed. Ideally we should remove this module once we no longer need it, so for now it would be better to import from our flask module [superdesk.flask](https://github.com/superdesk/superdesk-core/blob/async/superdesk/flask.py)

petrjasek:
it's actually importing from `superdesk.flask` , I missed that

MarkLark86 (approved):

PR #2924, opened 2025-05-13
Fix celery in eager mode, and pydantic service search where clause

MarkLark86:
### Purpose
Fixes/improvements found during fixing tests in the Planning module

### What has changed:
* Fix celery in tests: Don't wrap task in app context if in eager mode, as the caller is already in an app context when the task is called (was causing issues in tests)
* Fix search in Pydantic services, not supporting eve style where clause (such as `?where=event_id=="abcd123"`
* Support `force` param in the web request's `get_json` method
* Add more test fixtures for configuring publishing (filters, products & subscribers)


eos87 (approved):
Looks good

PR #2928, opened 2025-05-15
Async fix tests 4

MarkLark86:
### Purpose
Fix more behave tests

### What has changed
* Fix various issues with the archive and publish code
* Fix an issue with User resource usage in the Roles resource
* Expose Eve ContentTypes REST API instead of the Pydantic one (issues with Pydantic resource)
* Fix Eve HATEOAS with resources not registered with Eve
* Fix async issues in Spellcheckers
* Fix base test code - reset app config per test, refresh elastic indexes, use mock publish consumer in behave tests
* Fix behave tests


eos87 (approved):
Looks good to me

PR #2931, opened 2025-05-16
 Add support for planning template loading in Content API [STT-107]

devketanpro:
PR based on :- https://github.com/superdesk/superdesk-planning/pull/2315

### DIFF
--- a/content_api/app/__init__.py
+++ b/content_api/app/__init__.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8; -*-
+# -- coding: utf-8; --
 #
 # This file is part of Superdesk.
 #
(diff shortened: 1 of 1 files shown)


MarkLark86:
@devketanpro This should actually go into the Planning repo/module. We'd like to keep all custom Planning code out of core.
This is already done in the main Web API app, see [here](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/__init__.py#L284)

devketanpro:
> @devketanpro This should actually go into the Planning repo/module. We'd like to keep all custom Planning code out of core.
This is already done in the main Web API app, see [here](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/__init__.py#L284)

got it

PR #2934, opened 2025-05-19
Async fix tests 6

MarkLark86:
### Purpose
Fix more behave tests

### What has changed
* Fix issue with checking invisible stages from archive resource (checking ObjectId vs string)
* Fix issue on desk sending notifications (using dict not Pydantic resource/service)
* Fix async search ingest from latest changes there
* Fix issue in saved search not returning an async cursor
* Fix projection issue in search endpoint (use Eve service and not Pydantic service)
* Remove print statements and wip tags in behave tests


eos87 (approved):
Looks good

PR #2937, opened 2025-05-19
use soft delete for availability

petrjasek:
so it won't generate based on default next time

SDBELGA-956

### Steps to test

Resolves: SDBELGA-956


eos87 (approved):
Looks good

PR #2943, opened 2025-05-20
allow users to override other users availability

petrjasek:
if they have respective privilege

SDBELGA-950

### Purpose

it allows users with privileges change other users availability settings

### Steps to test

Resolves: SDBELGA-950


eos87 (approved):

eos87 (approved):
Looks good

PR #2944, opened 2025-05-20
fix(validation): Prefer Eve resource in data relation if available

MarkLark86:
### Purpose
Data relation validation fails from Pydantic resource.

### What has changed
* Prefer Eve resource when checking data validation
* Make sure to use ObjectId when checking against Pydantic resource


eos87 (approved):
Makes sense

PR #2945, opened 2025-05-20
allow to continue record from another record

petrjasek:
use `manage.py storage:record -n test3 --continue-from test2` to generate a record that will continue from `test2`

### Purpose

allows to continue data recording from previous recording, thus it avoids the need to create extra dumps

### Steps to test

Resolves: #[issue-number]


tomaskikutis (approved):
It does work, but before continuing recording we should 1. restore base dump 2. apply record 3. only then continue recording

tomaskikutis (approved):
:+1:

PR #2953, opened 2025-05-30
[SDESK-7630] Fix publishing of package items

MarkLark86:
### Purpose
Publishing of packages and their items was not working properly in the new publish async system. It required special handling on the ContentExchange to process the package items

### What has changed
* Update the `ContentPublishExchange` publish module to handle publishing of package items (code/logic copied from [develop branch](https://github.com/superdesk/superdesk-core/blob/develop/apps/publish/enqueue/enqueue_service.py#L99))
* Fix issue in `BasePublishExchangeFormatter` that previously included Wire Subscribers for non text items
* Don't cache formatted item in `BasePublishExchangeFormatter` if the item is a Package (formatted item might change based on subscriber permissions)
* Fix issue in publish endpoint that didn't publish nested packages
* Fix issue in `ContentPublishExchange` when polling is enabled and request comes from API (initiate celery task immediately, copies logic from old code)
* Moved some package utils into the `superdesk.publish_async.utils` module (shared utils across publish async module and publish endpoints)
* Init PublishCache in internal_destination, as it uses content matching with requires the cache to be available (fixes SDESK-7666]

**Note**: Extends on changes from https://github.com/superdesk/superdesk-core/pull/2951, as it was required as well. You can ignore the first commit in this PR ([[SDESK-7630] fix: Killing of archived items](https://github.com/superdesk/superdesk-core/commit/0fa289f39a7ada9a1001858804fb203960c36ead))

### DIFF
--- a/apps/archive/archive_media.py
+++ b/apps/archive/archive_media.py
@@ -74,7 +74,6 @@ class ArchiveMediaService:
                 doc["renditions"] = renditions
                 doc["mimetype"] = content_type
                 set_filemeta(doc, metadata)
-                # TODO-ASYNC[activity]: Prefix this next line with `await ` when updating this module
                 await add_activity(
                     "upload",
                     "uploaded media {{ name }}",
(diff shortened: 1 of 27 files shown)


MarkLark86:
This should also fix some failing tests in https://github.com/superdesk/superdesk-client-core/pull/4816

eos87 (approved):

PR #2956, opened 2025-06-04
handle contacts search on backend

petrjasek:
avoid sending elastic query from client,
there will be custom service for belga.

SDBELGA-951

### Purpose

avoid generating elastic query on client, only pass `q` param to server

### Steps to test

Resolves: #[issue-number]

### DIFF
--- a/apps/contacts/service.py
+++ b/apps/contacts/service.py
@@ -39,17 +39,21 @@ class ContactsService(Service):
                             }
                         },
                     ],
+                    "should": [],
+                    "filter": [],
                 }
             }
 
             if "all" not in req.args:
-                query["bool"]["should"] = [
-                    {"term": {"is_active": True}},
-                    {"term": {"public": True}},
-                ]
+                query["bool"]["should"].extend(
+                    [
+                        {"term": {"is_active": True}},
+                        {"term": {"public": True}},
+                    ]
+                )
 
             if req.args.get("contact_type"):
-                query["bool"]["filter"] = [{"term": {"contact_type": req.args.get("contact_type")}}]
+                query["bool"]["filter"].append({"term": {"contact_type": req.args.get("contact_type")}})
 
             args = req.args.copy()
             args.pop("q")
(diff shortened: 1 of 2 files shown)


apps/contacts/service.py
```diff
@@ -25,6 +26,49 @@ def get(self, req, lookup):
         if "all" not in req.args:
             lookup["is_active"] = True
 
+        if req and not req.args.get("source") and req.args.get("q"):
+            query = {
+                "bool": {
+                    "must": [
+                        {
+                            "query_string": {
+                                "query": req.args.get("q") + "*",
```
devketanpro: directly appending `*` could lead to malformed queries or performance issues, especially with user input.
we can sanitize input, or consider using `prefix` or `match_phrase_prefix` queries for more controlled behavior.
thoughts ?

apps/contacts/service.py
```diff
@@ -25,6 +26,49 @@ def get(self, req, lookup):
...
+                    ],
+                    "should": [
+                        {"term": {"is_active": True}},
+                        {"term": {"public": True}},
+                    ],
+                }
+            }
+
+            if req.args.get("contact_type"):
+                query["bool"]["filter"] = [{"term": {"contact_type": req.args.get("contact_type")}}]
```
devketanpro: to avoid unintentionally overriding other filters that might be added in the future, it would be better to append to the filter list like this:
 ```.py
query["bool"].setdefault("filter", []).append(
    {"term": {"contact_type": req.args.get("contact_type")}}
)
```
thoughts ?

apps/contacts/service.py
petrjasek: yep wanted to use `match_bool_prefix` search, but it doesn't support multiple fields. also tested multi match query with `bool_prefix` but that didn't work neither, so at the end I ended up with `query_string` with `*`, which is what the client was using anyhow

apps/contacts/service.py
petrjasek: yep might be good for type checking

apps/contacts/service.py
devketanpro: makes sense

devketanpro (approved):

PR #2972, opened 2025-06-18
Merge develop into async

MarkLark86:
**NOTE**: Not to be merged, but for review only. I will push the merge directly into the async branch


MarkLark86:
@petrjasek There failing tests are similar to those in async branch, but there is the availability tests failing (will look into that in another PR)

PR #2981, opened 2025-06-25
Photo incorrectly marked as corrected when only story is corrected [SDCP-953]

devketanpro:


features/publish_correction.feature
```diff
@@ -0,0 +1,109 @@
...
+    When we publish "story_1" with "publish" type and "published" state
+    Then we get OK response
+
+    # Correct story
+    When we publish "photo_1" with "correct" type and "corrected" state
+    """
+    {"headline": "Corrected Photo"}
+    """
+    Then we get OK response
+    When we publish "story_1" with "correct" type and "corrected" state
```
petrjasek: I think what we're trying to fix is when the photo is not corrected, only published once, and then the story where it's associated gets corrected. In that case the story should stay published and shouldn't get automatically corrected with the story.

petrjasek (approved):

PR #2984, opened 2025-06-30
Images scheduled 2 hours in advance when article is descheduled and scheduled again [SDESK-7688]

devketanpro:


petrjasek:
I don't see any rescheduling in the test, can you cover it? or it's the same issue if it's scheduled for the first time?

petrjasek:
I'm still not sure why it should be set now on update on all associated items, atm that's not the case. isn't the issue rather somewhere else?

petrjasek:
the test is not matching the current behaviour, the schedule on associations is only set when publishing the item, not when updating it. so I wonder still what you're trying to fix here, without using publishing service to publish the text item and then republishing it it's a usecase we don't have and I don't think we want to set the schedule to all associated items on update, some might be already published at that time, it's also updating an item which might be edited elsewhere, so something we want to avoid in general.

devketanpro:
thanks @petrjasek  for the feedback. The issue we're seeing is that when an article is descheduled and later rescheduled (without republishing explicitly via the publishing service), associated images still retain the original schedule or get an unexpected future schedule (e.g., +2 hours).

My initial approach was to update the associated items' schedules during the article update to fix this, but I now see this could conflict with published or concurrently edited items.

I’ll rework the logic so that schedule updates only occur via the publishing service, ensuring we don’t touch associated items during basic updates. I’ll also adjust the test to simulate the correct flow: `publish → deschedule → republish` via the publish endpoint.

Let me know if that aligns better with the intended architecture.

features/content_publish.feature
```diff
@@ -5218,4 +5218,150 @@ Feature: Content Publishing
...
+                  }
+              }
+          }
+      ]
+      """
+      When we publish "art-1" with "publish" type and "published" state
+      """
+      {
+          "publish_schedule": "2035-04-01T10:00:00+0000",
+          "schedule_settings": {"time_zone": "UTC"}
```
petrjasek: would rather test it some existing timezone, could be also an issue with that

apps/publish/content/common.py
```diff
@@ -967,7 +967,8 @@ def _inherit_publish_schedule(self, original, updates, associated_item):
         if self.publish_type == "publish" and (updates.get(PUBLISH_SCHEDULE) or original.get(PUBLISH_SCHEDULE)):
             schedule_settings = updates.get(SCHEDULE_SETTINGS, original.get(SCHEDULE_SETTINGS, {}))
             publish_schedule = updates.get(PUBLISH_SCHEDULE, original.get(PUBLISH_SCHEDULE))
-            if publish_schedule and not associated_item.get(PUBLISH_SCHEDULE):
+            if publish_schedule:
```
petrjasek: I wonder why this would fix it, seems like on deschedule it unsets the schedule on associations as well, doesn't it?

devketanpro:
from my investigation, the root cause was that the `publish_schedule` and `schedule_settings` on the associated media items within the body HTML were not being cleared during descheduling. As a result, when the article was scheduled again, the old values on the media items caused unintended behavior.

To fix this, I updated the `_handle_media_updates` method to explicitly clear the `publish_schedule` and `schedule_settings` of the associated media items when:

the parent article is being `descheduled`, or

the associated media item itself is being `descheduled`.

petrjasek (approved):

PR #2989, opened 2025-07-04
First publish not working with corrected related items [SDESK-7707]

devketanpro:


apps/publish/content/common.py
```diff
@@ -809,6 +809,9 @@ def _publish_associated_items(self, original, updates=None):
         archive_service = get_resource_service("archive")
 
         for associations_key, associated_item in associations.items():
+            if not associated_item or associated_item.get("state") == CONTENT_STATE.CORRECTED:
```
petrjasek: you're duplicating the next condition here, would make sense to move this one below and drop that `not associated_item` part

petrjasek (approved):

PR #2993, opened 2025-07-08
avoid hardcoding datetime offset

petrjasek:
### Purpose

atm server only accepts times with `+0000` offset and that forces client to send wrong data
when it needs to send local time. 

### Steps to test

Resolves: #[issue-number]


tomaskikutis:
@eos87 could you review this?

eos87 (approved):
Looks good

PR #2994, opened 2025-07-14
[STT-1241] - As an STT admin I like to add a start and end date to the subscriber config via the UI

BrianMwangi21:
### Purpose
Allow admins to schedule when a subscriber becomes active or inactive using start and end dates

### What has changed
- Added `schedule.start_date` and `schedule.end_date` fields to the subscribers schema.
- Implemented celery task `update_subscriber_activation_states` to auto-activate or deactivate subscribers based on current time.
- Added unit test to test logic

### Steps to test
1. Create or edit a subscriber and set schedule.start_date and schedule.end_date.
2. Ensure the subscriber is only active between those dates.
3. Confirm the task runs every day at midnight
4. Additionally, run SubscriberScheduleTestCase to verify logic.

Resolves: [STT-1241](https://sofab.atlassian.net/browse/STT-1241)

NOTES: UI part of ticket is in this [PR](https://github.com/superdesk/superdesk-client-core/pull/4884)

[STT-1241]: https://sofab.atlassian.net/browse/STT-1241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -378,7 +378,7 @@ CELERY_BEAT_SCHEDULE = {
     },
     "subscribers:schedule_update": {
         "task": "superdesk.publish.subscribers_schedule.update_subscriber_activation_states",
-        "schedule": crontab(minute="*"),
+        "schedule": crontab(minute=0, hour=local_to_utc_hour(0)),
     },
 }
 
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -379,7 +379,7 @@ CELERY_BEAT_SCHEDULE = {
     "subscribers:schedule_update": {
         "task": "superdesk.publish.subscribers_schedule.update_subscriber_activation_states",
         "schedule": crontab(minute="*"),
-    }
+    },
 }
 
 #: Sentry DSN - will report exceptions there
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -376,6 +376,10 @@ CELERY_BEAT_SCHEDULE = {
         "task": "apps.saved_searches.report",
         "schedule": crontab(minute=0),
     },
+    "subscribers:schedule_update": {
+        "task": "superdesk.publish.subscribers_schedule.update_subscriber_activation_states",
+        "schedule": crontab(minute="*"),
+    }
 }
 
 #: Sentry DSN - will report exceptions there
--- a/superdesk/publish/subscribers.py
+++ b/superdesk/publish/subscribers.py
@@ -141,7 +141,7 @@ class SubscribersService(CacheableService):
         self._validate_seq_num_settings(updates)
         subscriber = deepcopy(original)
         subscriber.update(updates)
-        
+
         self._apply_schedule_status(subscriber)
 
         # Apply the calculated status back to `updates`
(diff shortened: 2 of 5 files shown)


superdesk/publish/subscribers_schedule.py
```diff
@@ -0,0 +1,59 @@
...
+    this will check if the current time is within the start and end range
+    and update `is_active` accordingly.
+    """
+    now = utcnow()
+    service = get_resource_service("subscribers")
+
+    try:
+        # Query for subscribers that have a schedule defined
+        lookup = {
+            "$and": [
```
petrjasek: would be good to only get those that need some changes, so the query could be like:
```
today = get_local_today().date().isoformat()
{"$or": [
    {"schedule.startDate": today},
    {"schedule.endDate": today},
]}
```

superdesk/publish/subscribers.py
```diff
@@ -70,6 +70,13 @@ class SubscribersResource(Resource):
                 },
             },
         },
+        "schedule": {
+            "type": "dict",
+            "schema": {
+                "startDate": {"type": "datetime", "nullable": True},
```
petrjasek: would store date only, which would require type string

superdesk/default_settings.py
```diff
@@ -376,6 +376,10 @@ def local_to_utc_hour(hour):
         "task": "apps.saved_searches.report",
         "schedule": crontab(minute=0),
     },
+    "subscribers:schedule_update": {
+        "task": "superdesk.publish.subscribers_schedule.update_subscriber_activation_states",
+        "schedule": crontab(minute="*"),
```
petrjasek: once a day at local 0 should be enough

superdesk/publish/subscribers_schedule.py
```diff
@@ -0,0 +1,59 @@
...
+            # Determine desired state
+            is_after_start = not start or now >= start
+            is_before_end = not end or now <= end
+            should_be_active = is_after_start and is_before_end
+
+            if should_be_active != active:
+                service.system_update(subscriber["_id"], {"is_active": should_be_active}, subscriber)
+                updated = True
+
+            if updated:
```
petrjasek: not sure if this updated is needed, you could just continue in the same block

superdesk/publish/subscribers.py
```diff
@@ -70,6 +72,13 @@ class SubscribersResource(Resource):
                 },
             },
         },
+        "schedule": {
+            "type": "dict",
+            "schema": {
+                "startDate": {"type": "string", "nullable": True},
+                "endDate": {"type": "string", "nullable": True},
```
eos87: we should use snake case to keep consistent.
```suggestion
                "start_date": {"type": "string", "nullable": True},
                "end_date": {"type": "string", "nullable": True},
```

eos87 (changes requested):

superdesk/publish/subscribers.py
```diff
@@ -121,6 +130,7 @@ def get(self, req, lookup):
 
     def on_create(self, docs):
         for doc in docs:
+            self._apply_schedule_status(doc)
```
petrjasek, line 133: I'm not sure if it should actually do this on create or update, might be confusing that you select certain status explicitly and it would change it right away

BrianMwangi21: @petrjasek we do this after warning the user and they still go ahead and click save. That's why I was seeking clarity yesterday with Gideon. 

I was doing it on the front end then Tomas said I move this checks to the backend.

petrjasek (approved):

eos87 (approved):
Looks good

PR #2996, opened 2025-07-21
[STT-1241] - Clear schedule when it is in the past completely

BrianMwangi21:
### Purpose
Small improvement to clear schedule after deactivation to allow setting a new schedule easily.

### What has changed
- If a subscriber is deactivated because the end date is today, the schedule is also reset since it is now completely in the past. 

### Steps to test
1. Set a subscriber with `is_active = True` and `schedule.end_date = today`.
2. Let the schedule logic run.
3. Confirm the subscriber is deactivated and schedule fields are cleared.

Resolves: [#STT-1241](https://sofab.atlassian.net/browse/STT-1241)


eos87 (approved):
looks good

PR #3000, opened 2025-07-25
Default text content profile name returned as None in API responses [SDESK-7705]

devketanpro:


prod_api/content_types/service.py
```diff
@@ -16,9 +16,11 @@
 
 class ContentTypesService(superdesk.Service):
     def get_output_name(self, profile):
+        service = superdesk.get_resource_service("content_types")
         try:
             _id = bson.ObjectId(profile)
-            item = superdesk.get_resource_service("content_types").find_one(req=None, _id=_id) or {}
-            return format_content_type_name(item, str(_id))
+            item = service.find_one(req=None, _id=_id) or {}
```
petrjasek: using `item` here is a bit confusing, profile would be a better name probably, and the param could be `profile_id`

prod_api/content_types/service.py
```diff
@@ -15,10 +15,12 @@
...
+        service = superdesk.get_resource_service("content_types")
         try:
-            _id = bson.ObjectId(profile)
-            item = superdesk.get_resource_service("content_types").find_one(req=None, _id=_id) or {}
-            return format_content_type_name(item, str(_id))
+            _id = bson.ObjectId(profile_id)
+            profile = service.find_one(req=None, _id=_id) or {}
         except bson.errors.InvalidId:
-            return "None"
+            profile = service.find_one(req=None, _id=profile_id) or {}
```
petrjasek: it would be a bit more readable and avoid some duplication like:
```python3
try:
    _id = bson.ObjectId(profile_id)
except bson.errors.InvalidId:
    _id = profile_id

profile = ...
```

petrjasek (approved):

PR #3001, opened 2025-07-28
[STT-1244] Fix DataClass config issue, support generating OpenAPI docs

MarkLark86:
### Purpose
Main purpose in this PR is to fix serialisation issues with DataClasses, and support auto-generation of OpenAPI docs from async modules.

### What has changed
* Fix the DataClass serialisation not using model_config from parent classes
* Add DateWithOptionalTime (string in MongoDB, date/date-time in Elastic)
* Fix issue with search, supporting the entire elastic query from the `where` param
* Add OpenAPI doc generation & command capabilities


eos87 (approved):
looks good

PR #3002, opened 2025-07-28
implement collation handling in mongo find

petrjasek:
### Steps to test

Resolves: #[issue-number]

### DIFF
--- a/superdesk/core/resources/cursor.py
+++ b/superdesk/core/resources/cursor.py
@@ -8,7 +8,7 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-from typing import Any, Generic, TypeVar, Type, Optional
+from typing import Any, Generic, TypeVar, Type
 
 from pymongo.collation import Collation
 from motor.motor_asyncio import AsyncIOMotorCollection, AsyncIOMotorCursor
@@ -128,7 +128,7 @@ class MongoResourceCursorAsync(ResourceCursorAsync[ResourceModelType], Generic[R
         collection: AsyncIOMotorCollection,
         cursor: AsyncIOMotorCursor,
         lookup: dict[str, Any],
-        collation: Optional[Collation] = None,
+        collation: Collation | None = None,
     ):
         super().__init__(data_class)
         self.collection = collection
--- a/superdesk/core/resources/cursor.py
+++ b/superdesk/core/resources/cursor.py
@@ -8,8 +8,9 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-from typing import Any, Generic, TypeVar, Type
+from typing import Any, Generic, TypeVar, Type, Optional
 
+from pymongo.collation import Collation
 from motor.motor_asyncio import AsyncIOMotorCollection, AsyncIOMotorCursor
 
 from .model import ResourceModel
@@ -127,11 +128,13 @@ class MongoResourceCursorAsync(ResourceCursorAsync[ResourceModelType], Generic[R
         collection: AsyncIOMotorCollection,
         cursor: AsyncIOMotorCursor,
         lookup: dict[str, Any],
+        collation: Optional[Collation] = None,
     ):
         super().__init__(data_class)
         self.collection = collection
         self.cursor = cursor
         self.lookup = lookup
+        self.collation = collation
 
     async def __anext__(self) -> ResourceModelType:
         item = await self.next()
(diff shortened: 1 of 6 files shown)


superdesk/core/resources/service.py
```diff
@@ -978,6 +983,12 @@ async def system_update(self, item_id: ObjectId | str, updates: dict[str, Any])
         except ElasticNotConfiguredForResource:
             pass
 
+    def _get_collation(self) -> Optional[Collation]:
+        """Get collation for MongoDB queries if configured"""
+        if self.config.mongo is not None and getattr(self.config.mongo, "collation", False):
+            return Collation("en", strength=2)
+        return None
```
eos87: wouldn't it be better to allow "per query" collation instead? that would give us more flexibility. Or do you think it's good enough to care about one collation for now?

superdesk/core/resources/cursor.py
```diff
@@ -127,11 +128,13 @@ def __init__(
         collection: AsyncIOMotorCollection,
         cursor: AsyncIOMotorCursor,
         lookup: dict[str, Any],
+        collation: Optional[Collation] = None,
```
eos87: nit:
```suggestion
        collation: Collation | None = None,
```

superdesk/core/types/mongo.py
```diff
@@ -63,6 +63,9 @@ class MongoResourceConfig:
     #: Boolean determining if this resource supports versioning
     versioning: bool = False
 
+    #: Boolean determining if thise resource supports collation
```
eos87:
```suggestion
    #: Boolean determining if this resource supports collation
```

superdesk/core/resources/service.py
petrjasek: yep will probably make it rather per request, seems like it wouldn't be using even `_id` index when using different collation

eos87 (approved):
Looks good to me

superdesk/core/types/search.py
```diff
@@ -170,6 +170,9 @@ class SearchRequest(BaseModel):
 
     elastic: ESQuery = Field(default_factory=ESQuery)
 
+    #: If `True` it will make mongo query case-insensitive
+    collation: bool = False
```
petrjasek: I'm thinking about calling it `case_insensitive` instead, any feedback?

eos87: `case_insensitive` is more meaningful definitely. I'm still thinking about the fact that this would only affect mongo though. Should we make it so: if `case_insensitive` is `true` then it turns  `use_mongo` `true` by default?  or call it `mongo_case_insensitive`?

superdesk/core/resources/service.py
```diff
@@ -978,6 +983,12 @@ async def system_update(self, item_id: ObjectId | str, updates: dict[str, Any])
         except ElasticNotConfiguredForResource:
             pass
 
+    def _get_collation(self, search_request: SearchRequest) -> Optional[Collation]:
+        """Get collation for MongoDB queries if configured"""
+        if search_request.collation:
```
MarkLark86: IMO we should also provide a default on the resource config, so that requests from the front-end would also be case insensitive if configured. Something like:

```python
case_insensitive = (
    search_request.case_insensitive
    if search_request.case_insensitive is not None
    else self.config.case_insensitive
)
if case_insensitive:
    return Collation("en", strength=2)
return None
```

To support this, we would need `case_insensitive` on the SearchRequest to be `bool | None = None` so that the front-end can perform a case sensitive request when the resource defaults to case insensitive.

Also, then language param is `en`. Would this work for all languages, or does this need to be configurable somehow?

superdesk/core/types/search.py
MarkLark86: * `case_insensitive` is descriptive
* I would not default to using Mongo (this requires an index to be created for all possible search fields, otherwise it could be a performance hit).
* Elasticsearch is case insensitive, unless searching against Keywords (which is generally constructed based on CVs and such). I don't see this as too much of an issue.

In short, I think `case_insensitive` should suffice, without over-engineering the solution to use a lowercase normalizer/analyzer for Keywords only.

superdesk/core/resources/service.py
petrjasek: I'm not sure anymore if we should have it on resource level, that would make all requests use it and even searching by _id wouldn't use the index.

MarkLark86: Good point. Let's leave as it is now then

MarkLark86 (approved):
Looks good, but there are lint issues

petrjasek:
checked the mypy errors but none of them is related to those files, seems the same as on the base branch

PR #3009, opened 2025-08-11
[SDESK-7630] Fix unit tests

MarkLark86:
### Purpose
To get the remaining unit tests passing

### What has changed
* Remove `test_subscribers_secret_keys` unit test, as that is covered in the behave tests
* Fix setting/getting celery app, and defining celery tasks (should also fix issue with soft timeout incorrect value)
* Don't import planning in content_api settings, use importlib instead to see if it's available
* Use AsyncQuartTestCase to construct Quart instances instead of doing it manually (caused issues with get_config, get_app etc)
* Fix saml tests
* Use our test classes instead of conftest for availability (due to autocleanup, connection sharing etc that we already have)
* Upgrade failing tests to use async resources/functionality
* 


petrjasek (approved):

superdesk/celery_app/__init__.py
```diff
@@ -34,16 +34,7 @@
 BaseTaskClass: type[HybridAppContextTask] = HybridAppContextTask if IS_BEAT_PROCESS else HybridAppContextWorkerTask
 
 
-class SuperdeskCelery(Celery):
-    def task(self, *args, **opts):
-        # TODO-ASYNC: ``soft_time_limit`` is not working,
-        # for some reason it's converting the int to datetime instance
-        # such as: datetime.datetime(1970, 1, 1, 0, 10, tzinfo=tzutc())
-        opts.pop("soft_time_limit", None)
-        return super().task(*args, **opts)
```
eos87, line 43: how was this issue solved?

eos87 (approved):
Looks good!

superdesk/celery_app/__init__.py
MarkLark86: Setting the `Task` attribute on the Celery instance after construction and being used by task functions was causing various issues. As soon as I commented out the `celery.Task = ContextTask` in init_app, those issues went away and I have not came across this soft time limit issue since.

PR #3011, opened 2025-08-12
[NHUB-627] Refactor health endpoint logic into helper function

eos87:
### Purpose
Moved health status computation to a new `get_health_status` function for better reusability. The health endpoint now calls this helper, passing the application name from config. 

This is used from https://github.com/superdesk/newsroom-core/pull/1337

Belongs to: [NHUB-627]

[NHUB-627]: https://sofab.atlassian.net/browse/NHUB-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/system/health.py
+++ b/superdesk/system/health.py
@@ -17,7 +17,6 @@ import logging
 import superdesk
 
 from typing import Callable, List, Tuple
-
 from superdesk.core import get_app_config, get_current_app
 from superdesk.flask import Blueprint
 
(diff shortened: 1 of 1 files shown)


MarkLark86 (approved):
Looks good. In the future we might think about making this configurable, in terms of what checks to add, and this is a start

PR #3013, opened 2025-08-25
[SDESK-7744] - Added NoCredentialsError handling in the amazon sqs

BrianMwangi21:
### Purpose
Added retry-on-NoCredentialsError in the Amazon SQS FIFO transmitter to guard against the intermittent error.

### What has changed
- Refactor: Extract SQS setup and send into a separate function.
- Retry on missing credentials: First attempt uses `queue_item` credentials and immediately retry using the matching subscriber destination’s config. If both fail, then we finally raise the `credentialsError`

Resolves: #[7744](https://sofab.atlassian.net/browse/SDESK-7744)

### DIFF
--- a/superdesk/publish/transmitters/amazon_sqs_fifo.py
+++ b/superdesk/publish/transmitters/amazon_sqs_fifo.py
@@ -19,6 +19,7 @@ errors = [
     PublishAmazonSQSError.connectionError().get_error_description(),
     PublishAmazonSQSError.clientError().get_error_description(),
     PublishAmazonSQSError.sendMessageError().get_error_description(),
+    PublishAmazonSQSError.credentialsError().get_error_description(),
 ]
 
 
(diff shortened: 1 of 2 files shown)


superdesk/publish/transmitters/amazon_sqs_fifo.py
```diff
@@ -30,10 +31,24 @@ class AmazonSQSFIFOPublishService(publish_service.PublishService):
...
+
+        destinations = subscriber.get("destinations") or []
+        delivery_type = (target_destination or {}).get("delivery_type")
+        name = (target_destination or {}).get("name")
+
+        if not delivery_type or not name or not destinations:
+            return {}
+
+        for destination in destinations:
+            if delivery_type and destination.get("name") == name and destination.get("delivery_type"):
```
MarkLark86: Should the delivery type value be matched, i.e. `delievery_type == destination.get("delivery_type")`?

BrianMwangi21: My bad. I was doing some refactoring while fixing the tests. Must have missed that

MarkLark86 (approved):

PR #3015, opened 2025-08-25
always check metadata in binaries if need updates

petrjasek:
if the changes to item metadata were done in previous versions and it was published without changes affecting those it wouldn't update binaries.

SDCP-910

replaces https://github.com/superdesk/superdesk-core/pull/3014

### Steps to test

Resolves: #[issue-number]


MarkLark86 (approved):
It would be good to also include tests from the PRs that were reverted (see https://github.com/superdesk/superdesk-core/pull/3014)

petrjasek:
will make a separate PR for that, seems not to be related to the media metadata

PR #3020, opened 2025-09-03
Update availability prod api endpoint and specs [SDBELGA-976]

devketanpro:


prod_api/tests/test_user_availability.py
```diff
@@ -81,3 +81,43 @@ def test_readonly(prodapi_app_with_data, prodapi_app_with_data_client):
...
+        # Insert a user availability entry
+        superdesk_app.data.insert(
+            "user_availability",
+            [
+                {
+                    "user": user["_id"],
+                    "date": "2023-05-16",
+                    "status": "partial",
+                    "working_hours": [{"start_time": "12:00", "end_time": "17:00"}],
+                }
```
petrjasek: would be good to add availability for another day to test if that filtering actually works

petrjasek (commented):
pls also update the api documentation

petrjasek:
documentation to be updated is this one https://github.com/superdesk/superdesk-core/blob/develop/specs/productionapi.yaml

petrjasek:
can you also update in docs those working_hours samples times to match the api output?

petrjasek (approved):

PR #3021, opened 2025-09-05
tests: New, cleaner core test module

MarkLark86:
### Purpose
The original async test code in superdesk-core no longer works in Newshub v3.1. Instead of making further modificiations to existing sueprdesk-core test code, I came up with a new test module that is cleaner to use and understand.

### DIFF
--- a/superdesk/core/tests/app.py
+++ b/superdesk/core/tests/app.py
@@ -1,0 +1,228 @@
+import logging
+from copy import deepcopy
+import os
+from pathlib import Path
+from importlib import import_module
+
+from quart import Quart, Config
+
+from .common import TestAppContext, NotificationMock, TestFactory
+from .http_client import TestClient
+from .resources import close_all_db_connections, drop_all_db_databases, delete_all_db_documents, init_db_indexes
+from .mongo import get_mongo_uri
+
+
+logger = logging.getLogger(__name__)
+
+
+class TestAppFactory(TestFactory):
+    async def before_module(self, context: TestAppContext) -> None:
+        setup_logging()
+        await close_all_db_connections(context)
+        await create_app(context)
+        drop_all_db_databases(context)
+        init_db_indexes(context)
+
+        if context.factory.init_app_context:
+            context.app_context = context.app.app_context()
+            await context.app_context.push()
+
+        if context.factory.init_request_context:
+            context.test_context = context.app.test_request_context("/")
+            await context.test_context.push()
+
+    async def after_module(self, context: TestAppContext) -> None:
+        if context.test_context:
+            try:
+                await context.test_context.pop()
+            except LookupError:
+                pass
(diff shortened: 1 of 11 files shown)


superdesk/core/tests/app.py
```diff
@@ -0,0 +1,228 @@
...
+
+def setup_config(context: TestAppContext, config: dict) -> dict:
+    app_abspath = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+    app_config = Config(app_abspath)
+    app_config.from_object(context.factory.default_settings_module or "superdesk.default_settings")
+    app_config = deepcopy(app_config)
+
+    cwd = Path.cwd()
+    for p in [cwd] + list(cwd.parents):
+        settings = p / "settings.py"
```
petrjasek, line 130: personally I'm not sure if there is a usecase to read it from there, would rather read it from eg. `test_settings.py` if needed to avoid overriding some core test settings with instance config

petrjasek (approved):

superdesk/core/tests/app.py
```diff
@@ -0,0 +1,228 @@
+import logging
+from copy import deepcopy
+import os
+from pathlib import Path
```
eos87, line 4:
```suggestion
import os
import logging

from copy import deepcopy
from pathlib import Path
```

superdesk/core/tests/app.py
```diff
@@ -0,0 +1,228 @@
...
+            context.app.config.update(context.factory.config_backup)
+
+        async with context.app.app_context():
+            delete_all_db_documents(context)
+
+        # context.app_context = context.app.app_context()
+        # await context.app_context.push()
+
+        # context.test_context = context.app.test_request_context("/")
+        # await context.test_context.push()
```
eos87, line 62: probably we don't need this anymore

superdesk/core/tests/behave.py
```diff
@@ -0,0 +1,88 @@
+from typing import Callable, Any
+from functools import partial
+from os import environ
+import asyncio
+import logging
```
eos87, line 5:
```suggestion
import asyncio
import logging

from os import environ
from typing import Callable, Any
from functools import partial
```

superdesk/core/tests/pytest_functions.py
```diff
@@ -0,0 +1,164 @@
...
+#     # async with context.app.app_context():
+#     yield context.app
+#     # await app_context.pop()
+#     # async with context.app.app_context():
+#     #     yield context.app
+#
+#     if context.factory.signals.after_test:
+#         await context.factory.signals.after_test(context)
+#
+#
```
eos87, line 155: probably not needed either
```suggestion
```

eos87 (approved):
Looks good, just left some really minor comments.

PR #3026, opened 2025-09-09
exported advisories loose formatting upon reopening them [SDBELGA-972]

devketanpro:


superdesk/editor_utils.py
```diff
@@ -349,8 +349,16 @@ def render(self):
...
+                    normalized_text = []
+                    for ch in block["text"]:
+                        if ch == "\n":
+                            normalized_text.append("\u2028")  # Use Unicode LINE SEPARATOR
+                        elif unicodedata.category(ch) != "Cc":
+                            normalized_text.append(ch)
+                    block["text"] = "".join(normalized_text)
+
             html = self.exporter.render(content_state)
+            html = html.replace("\u2028", "<br/>")
```
petrjasek: would rather use just `<br>` there so we only use one form

superdesk/editor_utils.py
```diff
@@ -349,8 +349,16 @@ def render(self):
                 block.setdefault("entityRanges", [])
                 block.setdefault("inlineStyleRanges", [])
                 if block.get("text"):
-                    block["text"] = "".join(ch for ch in block["text"] if unicodedata.category(ch) != "Cc")
+                    normalized_text = []
+                    for ch in block["text"]:
+                        if ch == "\n":
+                            normalized_text.append("\u2028")  # Use Unicode LINE SEPARATOR
+                        elif unicodedata.category(ch) != "Cc":
```
petrjasek, line 356: what are you trying to filter out here? it might also remove tabs if I got it right

devketanpro: The filtering is intended to remove control characters (Unicode category C) that can break XML rendering in lxml. Without it, certain characters in the content (even invisible ones) were causing errors like ValueError: All strings must be XML compatible. We do allow valid formatting characters like \n (handled via LINE SEPARATOR) and can preserve tabs if needed. This ensures that normal text and Unicode (like qualité) render correctly without breaking the HTML export.

petrjasek: I see

petrjasek: can you add a comment to the code with that info?

petrjasek (approved):

PR #3028, opened 2025-09-11
Update availability prod api endpoint and specs [SDBELGA-976]

devketanpro:
Updated the logic to ensure that when the day parameter is provided, only the requested day's availability is returned. If no data exists for that day, an empty list is returned.
Updated the `format_hours` function to always return working hours in HH:MM format.

### DIFF
--- a/prod_api/tests/test_user_availability.py
+++ b/prod_api/tests/test_user_availability.py
@@ -91,8 +91,6 @@ def test_service_get_by_day(superdesk_app, prodapi_app_with_data):
     with superdesk_app.app_context():
         user = superdesk_app.data.find_one("users", req=None, username="admin")
         assert user
-        superdesk_app.data.remove("user_availability", {"user": user["_id"], "date": "2023-05-16"})
-        superdesk_app.data.remove("user_metrics", {"user": user["_id"], "date": "2023-05-16"})
         superdesk_app.data.insert(
             "default_user_availability",
             [{"_id": user["_id"], "enabled": True}],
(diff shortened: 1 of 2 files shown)


prod_api/tests/test_user_availability.py
```diff
@@ -102,29 +104,53 @@ def test_service_get_by_day(superdesk_app, prodapi_app_with_data):
...
+        assert len(availability) == 1
+        assert availability[0]["date"] == "2023-06-01"
+        assert availability[0]["status"] == "available"
+        assert availability[0]["working_hours"] == []
+
+        # Query for a day with no availability
+        resp = client.get("/prodapi/v1/user_availability?day=2023-05-16")
+        assert resp.status_code == 200
+        items = resp.json["_items"]
+        assert len(items) == 1
```
petrjasek: I wonder why it returns something if there is no data? is it just empty record for the day?

prod_api/user_availability/service.py
```diff
@@ -26,9 +26,13 @@ class AvailabilityData(TypedDict):
 
 
 def format_hours(availability_day):
+    def to_hhmm(t):
+        # Converts 'HH:MM:SS' or 'HH:MM' to 'HH:MM'
+        return t[:5] if t and len(t) >= 5 else t
```
petrjasek: you can use `datetime.time` it would be more readable imo:
```py
time.fromisoformat(t).strftime('%H:%M')
```
and would also call the function format_time or something

prod_api/user_availability/service.py
```diff
@@ -26,9 +26,16 @@ class AvailabilityData(TypedDict):
 
 
 def format_hours(availability_day):
+    def format_time(t):
+        # Converts 'HH:MM:SS' or 'HH:MM' to 'HH:MM'
+        try:
+            return time.fromisoformat(t).strftime("%H:%M")
+        except Exception:
```
petrjasek: can you catch something more specific like `ValueError`?

prod_api/user_availability/service.py
```diff
@@ -30,7 +30,7 @@ def format_time(t):
         # Converts 'HH:MM:SS' or 'HH:MM' to 'HH:MM'
         try:
             return time.fromisoformat(t).strftime("%H:%M")
-        except Exception:
+        except (ValueError, TypeError):
             return t[:5] if t and len(t) >= 5 else t
```
petrjasek: I wonder if this would work in case there would be any of those errors actually, what other value are you expecting there?

devketanpro:
The TypeError would only occur if the value is None or not a string, since in that case t[:5] would fail. I’ve updated the code to explicitly check for None and non-string inputs before parsing, and now it only catches ValueError for invalid time strings.

petrjasek (approved):

PR #3038, opened 2025-09-19
add xmp metadata parsing for videos

txl-cp:
SDCP-959

use exiftool to parse xmp tags

to be completed with [client-core feature](https://github.com/superdesk/superdesk-client-core/pull/4882)


apps/publish/content/common.py
```diff
@@ -994,15 +1003,28 @@ def _update_picture_metadata(self, updates, original, updated):
                 if not media_id:
                     continue
 
-                original_metadata = get_metadata_from_item(original, mapping)
-                metadata = get_metadata_from_item(updated, mapping)
+                picture = app.media.get(media_id)
```
petrjasek: this could be a video as well, would call it rather `media`

apps/publish/content/common.py
```diff
@@ -994,15 +1003,28 @@ def _update_picture_metadata(self, updates, original, updated):
...
+                binary = picture.read()
 
-                if original_metadata == metadata:
+                file_metadata = (
+                    video_read_metadata(binary) if updated[ITEM_TYPE] == "video" else image_read_metadata(binary)
```
petrjasek: I think it would simplify the usage in general if we would have a generic functions with type param, so this logic could be inside:
```py
file_metadata = read_metadata(binary, updated[item_type])
```
for all those related methods

scripts/tests_setup
```diff
@@ -2,7 +2,16 @@
 
 # Install python package dependencie
 sudo apt-get -y update
-sudo apt-get -y install libxml2-dev libxmlsec1-dev libxmlsec1-openssl
+sudo apt-get -y install libxml2-dev libxmlsec1-dev libxmlsec1-openssl perl make curl
```
petrjasek: wouldn't installing `libimage-exiftool-perl` be enough?

superdesk/media/video.py
```diff
@@ -42,3 +44,161 @@ def get_meta(filestream):
...
+        with ExifToolHelper() as et:
+            et.execute(*args, path)
+
+        with open(path, "rb") as r:
+            return r.read()
+    except ExifToolExecuteError as e:
+        logger.exception(e.stderr)
+        return bin
+    finally:
+        remove(path)
```
petrjasek: I guess you could do the whole thing inside that tempfile block,
calling `tmp.seek(0)` after `et.execute` and returning `tmp.read()` after

apps/publish/content/common.py
```diff
@@ -994,15 +1003,28 @@ def _update_picture_metadata(self, updates, original, updated):
...
+                )
+                metadata = (
+                    get_video_from_photo(get_metadata_from_item(updated, mapping))
```
petrjasek: wouldn't expose that video metadata is converted from photo metadata, that can be done inside the method,
so could be handled inside `get_metadata_from_item` with explicit type param

txl-cp: abstracted to get_metadata_from_item and moved related fns to metadata.py

scripts/tests_setup
txl-cp: I tried that initially but I thought it would be better if exiftool versions in core and client-core match. the libimage only installs 12.x

superdesk/media/video.py
```diff
@@ -42,3 +44,161 @@ def get_meta(filestream):
...
+            else {}
+        ),
+    }
+    tags = {k: vv for k, v in xmp.items() if (vv := v or photo.get(k))}
+    return cast(VideoMetadata, tags)
+
+
+def map_exiftool_args(video: VideoMetadata) -> list[str]:
+    args = ["-sep", ","]
+    args.extend(f"-{k}={','.join(v) if isinstance(v, list) else v}" for k, v in video.items())
```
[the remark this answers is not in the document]
txl-cp: don't think this is needed since we're working with standard tags and exiftool knows exacly what tag goes in what namespace

MarkLark86:
@txl-cp Does pyexiftool library provide async capabilities? My concern with Superdesk v3+ is that this will restrict an API instance from processing other requests until the metadata is written into the video file.

scripts/tests_setup
petrjasek: I'm not sure if that's an issue, I suppose most of those tags used are there for long time anyhow and it would simplify the installation in different environments

tests/media/image_metadata_test.py
```diff
@@ -25,37 +25,41 @@ def image_binary() -> bytes:
...
-        CountryCode="CAN",
-        Creator=["Paul Chiasson"],
-        CreatorsJobtitle="stf",
-        JobId="DPI755",
-        Instructions="EDS NOTE:A FILE PHOTO",
-        Title="MORT PIÉTONNE MONTRÉAL 20201014",
-        CopyrightNotice="",
-        CreditLine="The Canadian Press",
-        ProvinceState="PQ",
+        {
```
petrjasek, line 27: I like the previous version better, I think copilot got it wrong here, it should work just fine

superdesk/media/metadata.py
```diff
@@ -0,0 +1,54 @@
...
+    write_metadata as video_write_metadata,
+    VideoMetadata,
+)
+
+
+def read_metadata(bin: bytes, content_type: CONTENT_TYPE_LITERAL):
+    return video_read_metadata(bin) if content_type == "video" else image_read_metadata(bin)
+
+
+def get_metadata_from_item(item: Item, mapping: PhotoMetadataMapping, content_type: CONTENT_TYPE_LITERAL):
```
petrjasek: I wonder if this `PhotoMetadataMapping` should be renamed somehow to reflect it's not only used for photo, would you actually have same config for both photo and video metadata?

superdesk/media/metadata.py
```diff
@@ -0,0 +1,54 @@
...
+)
+from superdesk.media.video import (
+    read_metadata as video_read_metadata,
+    get_video_from_photo,
+    write_metadata as video_write_metadata,
+    VideoMetadata,
+)
+
+
+def read_metadata(bin: bytes, content_type: CONTENT_TYPE_LITERAL):
```
petrjasek: can you also specify return types in the file?

superdesk/media/metadata.py
```diff
@@ -0,0 +1,54 @@
...
+        value = get_item_value(item, src)
+        if value is not None:
+            metadata[dest] = value
+
+    if content_type == "video":
+        return get_video_from_photo(metadata)
+    return metadata
+
+
+def get_item_value(item, src: str):
```
petrjasek: would call it `_get_item_value` to make it "private"

txl-cp:
> @txl-cp Does pyexiftool library provide async capabilities? My concern with Superdesk v3+ is that this will restrict an API instance from processing other requests until the metadata is written into the video file.

It's not async. I can update the write fn so i/o is non-blocking. is it needed?

MarkLark86:
> > @txl-cp Does pyexiftool library provide async capabilities? My concern with Superdesk v3+ is that this will restrict an API instance from processing other requests until the metadata is written into the video file.
> 
> It's not async. I can update the write fn so i/o is non-blocking. is it needed?

In Superdesk v3.0+ it will be beneficial. Otherwise this task would stop an API/worker instance from processing other API requests while this media processing is going on. So not needed now, but beneficial in v3.0+

scripts/tests_setup
txl-cp: adjusted to use the libimage

superdesk/media/metadata.py
txl-cp: combined videometadata and photometadata and moved it over to metadata_mapping

superdesk/media/metadata_mapping.py
```diff
@@ -0,0 +1,69 @@
+from typing import Dict, List, Literal, TypedDict
+
+
+class Metadata(TypedDict, total=False):
```
petrjasek: would call it `MediaMetadata` probably, `Metadata` sounds too generic

petrjasek (commented):
looks good

petrjasek:
can you check the conflits pls? seems like I don't have permissions

PR #3057, opened 2025-11-04
fix elastic mapping for dynamic embedded object

petrjasek:
allow indexing object with dynamic properties

### Steps to test

Resolves: #[issue-number]


eos87 (approved):
Nice catch!

PR #3059, opened 2025-11-04
add date param to index from mongo

petrjasek:
SDESK-7791

### Steps to test

Resolves: SDESK-7791


eos87 (approved):
Looks good. We will need to port this one into the async version.

PR #3077, opened 2025-11-28
[STT-1484] - Configure publish tab to show the "sent to" as default 

BrianMwangi21:
### Purpose
This PR adds a config setting to allow choosing which tab the authoring actions side panel shows by default.

### What has changed
- Added `AUTHORING_ACTIONS_DEFAULT_TAB` to `default_settings.py` with validation and defaulting to “publish” 
- Send the setting through `apps/client_config.py` tot the frontend

### How to test
1. Start the server and check the `/api/client_config` endpoint. Verify `config.authoring_actions_default_tab` is “publish” by default.
2. Set `AUTHORING_ACTIONS_DEFAULT_TAB=send_to`, restart the server, and confirm `/api/client_config` now reports “send_to”.

**NOTE:** This PR is coupled with the client changes in this [PR](https://github.com/superdesk/superdesk-client-core/pull/5013).

Resolves: #[STT-1484](https://sofab.atlassian.net/browse/STT-1484)

[STT-1484]: https://sofab.atlassian.net/browse/STT-1484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -1370,14 +1370,14 @@ BROADCAST_ENABLED = strtobool(env("BROADCAST_ENABLED", "true"))
 
 CORRECTIONS_WORKFLOW = False
 
-#: Default tab shown in the authoring interactive actions panel
+#: Default tab shown in the authoring actions side panel
 #:
 #: .. versionadded:: 3.1
 #:
 AUTHORING_ACTIONS_DEFAULT_TAB = env("AUTHORING_ACTIONS_DEFAULT_TAB", "publish")
 if AUTHORING_ACTIONS_DEFAULT_TAB not in ("publish", "send_to"):
     logger.warning(
-        "Invalid AUTHORING_ACTIONS_DEFAULT_TAB '%s'. Falling back to 'publish'.",
+        "Invalid value for AUTHORING_ACTIONS_DEFAULT_TAB '%s'. Defaulting to 'publish'.",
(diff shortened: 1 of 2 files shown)


superdesk/default_settings.py
```diff
@@ -1369,3 +1369,15 @@ class PublishChannelConfig(TypedDict, total=False):
 BROADCAST_ENABLED = strtobool(env("BROADCAST_ENABLED", "true"))
 
 CORRECTIONS_WORKFLOW = False
+
+#: Default tab shown in the authoring actions side panel
+#:
+#: .. versionadded:: 3.1
+#:
+AUTHORING_ACTIONS_DEFAULT_TAB = env("AUTHORING_ACTIONS_DEFAULT_TAB", "publish")
```
MarkLark86, line 1377: If this setting is not used anywhere on the backend, we should probably add it to the frontend config file (in client-core) instead

petrjasek: I was thinking the same, but on the other hand if we want to move towards saas I would keep it on the backend so eventually we could change it via UI

MarkLark86: Good point :+1:

MarkLark86 (approved):

PR #3083, opened 2025-12-03
[SDESK-7826] fix: Add audit and WS notifications to pydantic based resources

MarkLark86:
### Purpose
With the newer Pydantic based async resources, we weren't including them in the audit log or the websocket resource notifications.

### What has changed
Add global data listeners to add audit log entries and sending of websocket notifications

Resolves: [SDESK-7826](https://sofab.atlassian.net/browse/SDESK-7826)

[SDESK-7826]: https://sofab.atlassian.net/browse/SDESK-7826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/audit.feature
+++ b/features/audit.feature
@@ -1,0 +1,63 @@
+Feature: Resource Audit Logs
+    @auth
+    @notification
+    Scenario: Pydantic resources create audit logs
+        When we post to "/filter_conditions" with success
+        """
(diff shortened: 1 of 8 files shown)


petrjasek:
would add some config to resource config if we want notifications so it's only doing it when needed, we're already sending too many notifications

petrjasek (approved):

PR #3085, opened 2025-12-05
avoid deepcopy in resource manager

petrjasek:
unnecessary overhead, it's used when generating links all the time

### Steps to test

Resolves: #[issue-number]


MarkLark86 (approved):

MarkLark86 (commented):
@petrjasek This is also [this one](https://github.com/superdesk/superdesk-core/blob/release/3.1/superdesk/core/mongo/mongo_manager.py#L67) from the Mongo factory. It's used every time in the resource service to access mongo. I looked at Elastic, it seems fine

PR #3086, opened 2025-12-05
run ingest cleanup once per hour instead of 4x

petrjasek:
### Steps to test

Resolves: #[issue-number]


MarkLark86 (commented):
Why the change in frequency here?

petrjasek:
I was thinking what might be causing some unnecessary load, but not sure if it's a problem actually

PR #3091, opened 2025-12-10
[STT-1516] Allow `on_fetched` and `aggs` for search endpoint requests to be disabled

MarkLark86:
### Purpose
To improve performance of requests, support disabling content `search` endpoint's use of `on_fetched` signals and generation of aggregations.

Resolves: [STT-1516](https://sofab.atlassian.net/browse/STT-1516)

[STT-1516]: https://sofab.atlassian.net/browse/STT-1516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/search/__init__.py
+++ b/apps/search/__init__.py
@@ -11,11 +11,11 @@
 import superdesk
 
 from copy import deepcopy
-from eve_elastic.elastic import set_filters, fix_query
+from eve.utils import ParsedRequest
+from eve_elastic.elastic import set_filters
 
-from superdesk.core import json, get_current_app, get_app_config, get_current_async_app
+from superdesk.core import json, get_current_app, get_app_config
 from superdesk.resource_fields import ITEMS
-from superdesk.types import ArchiveResourceModel
 from superdesk.eve_async import AsyncBaseService, ElasticAsyncEveCursor
 from superdesk import get_resource_service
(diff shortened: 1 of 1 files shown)


apps/search/__init__.py
```diff
@@ -249,12 +250,13 @@ async def get_async(self, req, lookup) -> ElasticAsyncEveCursor:
 
         cursor = await self.elastic_async.search(query, types, params)
 
-        app = get_current_app().as_any()
-        for resource in types:
-            response = {ITEMS: [doc async for doc in cursor if doc["_type"] == resource]}
-            await getattr(app, "on_fetched_resource").call_async(resource, response)
-            await getattr(app, "on_fetched_resource_%s_async" % resource).call_async(response)
-            await getattr(app, "on_fetched_resource_%s" % resource).call_async(response)
+        if not req or not req.args or req.args.get("run_signals", True):
```
petrjasek: I wonder if we should make it rather a kwarg in `get_async` because often we would be using it with
`..get_async(req=None)`, so `.get_async(req=None, signals=False)` would be easier than constructing request for that

petrjasek (approved):

PR #3096, opened 2025-12-17
[SDESK-7725] - Skip re-fetch on POST when no projection is defined

BrianMwangi21:
### Purpose
This PR skips the unnecessary re-fetch on POST requests for `auth_server_clients`. Previously after client creation, the flow would refetch from the db and return the hashed password instead of the plain text secret that was just generated thus overwriting it and it ends up on the UI. 

### What has changed
- Updated `resource_rest_endpoints.py` to only re-fetch if a POST projection is actually defined
- Confirmed no breaking changes for `users_async` and `subscribers` since they have POST projections defined already

### Steps to test
1. In the settings, go to Production API and create a new client
2. On save, confirm the client secret shown is the plain text version and not the hashed version
3. Locally, test that the credentials work by running for example:
Request:
```
CLIENT_ID=69429608e319ec9077d4859b
CLIENT_SECRET=ej9Vh87ospwD53q4ohrvlm0lMdDcrKhbUYt4nDoV

curl -u "CLIENT_ID:CLIENT_SECRET" \
  -X POST "http://localhost:5000/api/auth_server/token" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "grant_type=client_credentials"
```
Response should be:
```
{"access_token": "eyJhbG...", "expires_in": 86400, "token_type": "Bearer"}
```
NOTE: Follow the [guide](https://superdesk.readthedocs.io/en/latest/auth_server.html#testing) for local testing. 

Resolves: #[SDESK-7725](https://sofab.atlassian.net/browse/SDESK-7725)

[SDESK-7725]: https://sofab.atlassian.net/browse/SDESK-7725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/auth_server_clients.feature
+++ b/features/auth_server_clients.feature
@@ -151,3 +151,25 @@ Feature: Authorization Server Client Management
         """
         {"_status": "ERR", "_issues": {"scope": {"0": "Input should be 'ARCHIVE_READ', 'DESKS_READ', 'PLANNING_READ', 'CONTACTS_READ', 'USERS_READ', 'ASSIGNMENTS_READ' or 'EVENTS_READ'"}}}
         """
+
+     @auth
+     Scenario: Can register client and authenticate with returned secret
+         When we post to "/auth_server_clients"
+         """
+         [{
+             "name": "Test Client",
+             "scope": ["ARCHIVE_READ", "USERS_READ"]
+         }]
+         """
+         Then we get OK response
+         And we get existing resource
+         """
+         {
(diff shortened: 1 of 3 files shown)


petrjasek:
would be great to test it via .feature file, registering client there and use returned credentials to get token

petrjasek (approved):

PR #3109, opened 2025-12-30
fix nosql injection on sensitive data

petrjasek:
avoid filtering using user password and other
sentivive information

STT-1526

### Steps to test

Resolves: #[issue-number]


eos87 (approved):
Looks good

PR #3130, opened 2026-01-22
fix slow response on correction

petrjasek:
handle corrections via celery

SDCP-1054

### Steps to test

Resolves: #[issue-number]

### DIFF
--- a/superdesk/default_settings.py
+++ b/superdesk/default_settings.py
@@ -533,6 +533,15 @@ PUBLISH_CHANNELS: list[PublishChannelConfig] = [
     },
     {
         "operations": ["correct"],
+        "filter": lambda item: not len(item.get("associations") or {}),
+        "config": ExchangeConfig(
+            exchange="content",
+            filter="content:corrected",
+            router="asyncio",
+        ),
+    },
+    {
+        "operations": ["correct"],
         "config": ExchangeConfig(
             exchange="content",
             filter="content:corrected",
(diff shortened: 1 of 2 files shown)


superdesk/default_settings.py
```diff
@@ -531,15 +531,6 @@ class PublishChannelConfig(TypedDict, total=False):
             filter="resend",
         ),
     },
-    {
```
MarkLark86, line 534: After thinking about it, we should probably never use the AsyncioPublishConsumer. That would mean any request to publish something has the potential to need to wait for all transmissions to finish before responding.

 https://github.com/superdesk/superdesk-core/blob/36d0b2c55b5534788321ce83be002be998d3de8f/superdesk/publish_async/exchanges/exchange_factory.py#L284-L289

Removing the `is_async` clause in these `if` statements should solve the issue.

We wouldn't need this config change then either

petrjasek: I'm a bit confused, so when the consumer is CeleryPublish the asyncio router would still send it to celery?

MarkLark86: Yes, celery will be used to transmit if `CeleryPublishConsumer` is configured.
My point earlier was that the `AsyncioPublishConsumer` should never be used with `AsyncioPublishRouter`, as this will hold up the client request until transmission has been completed.

The config change you did fixes the correction, but there are other publish actions that could use asyncio Router & Consumer. To fix this, and disable `AsyncioPublishConsumer` for now (until we have a better overall solution) would be to disable the above if statement.

This would fix it so that no asyncio router will use asyncio consumer.

petrjasek:
will add it to release now, let me know if this needs changes @MarkLark86

MarkLark86:
> will add it to release now, let me know if this needs changes @MarkLark86

Hey @petrjasek, this looks good, no changes required

PR #3134, opened 2026-01-26
set user metrics based on authors field

petrjasek:
SDBELGA-1024

### Steps to test

Resolves: #[issue-number]

### DIFF
--- a/apps/publish/content/common.py
+++ b/apps/publish/content/common.py
@@ -177,8 +177,7 @@ class BasePublishService(BaseService):
         CropService().update_media_references(updates, original, True)
         signals.item_published.send(self, item=original, after_scheduled=False)
 
-        if original.get("original_creator"):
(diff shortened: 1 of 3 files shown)


apps/publish/content/common.py
```diff
@@ -434,6 +433,23 @@ def _process_publish_updates(self, original, updates):
...
+        Increments the published articles metric for each author with a user ID.
+        If no authors are present, falls back to the original_creator.
+
+        :param dict item: article item to track
+        """
+        authors = item.get("authors", [])
+        if authors:
+            for author in authors:
+                if isinstance(author, dict) and author.get("parent"):
+                    print("AUTHOR", author)
```
MarkLark86: `print` statement not required

MarkLark86 (approved):

PR #3144, opened 2026-02-04
[STT-1452] - Fix the STT "Autotranslate and publish" macro 

BrianMwangi21:
### Purpose
This PR adds an `_internal_destination` flag so downstream macros can handle publishing the routed copy correctly.

### What has changed
- Set a temporary `_internal_destination` flag on items before calling `send_to` in internal destination routing, then remove it afterward.

### How to test
1. Have 2 Desks. Desk A and Desk B.
2. Ensure internal-destination routing is configured to Desk B with the incoming macro `auto_translate_and_publish_macro` in one of the stages.
3. Create a non-English story with headline + body like using the Pikaplus profile
4. Publish the story or use the “Send to” flow
5. Verify the routed copy in the Auto Translation desk is translated and published.

**Note**:
- This change is intended to be consumed by the STT macro update in this corresponding [PR](https://github.com/superdesk/superdesk-stt/pull/682).

Resolves : [STT-1452](https://sofab.atlassian.net/browse/STT-1452)

[STT-1452]: https://sofab.atlassian.net/browse/STT-1452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/tasks.py
+++ b/apps/tasks.py
@@ -70,7 +70,9 @@ def compare_dictionaries(original, updates):
     return modified
 
 
-async def send_to(doc, update=None, desk_id=None, stage_id=None, user_id=None, default_stage="incoming_stage"):
+async def send_to(
+    doc, update=None, desk_id=None, stage_id=None, user_id=None, default_stage="incoming_stage", macro_kwargs=None
+):
     """Send item to given desk and stage.
 
     Applies the outgoing and incoming macros of current and destination stages
@@ -91,7 +93,7 @@ async def send_to(doc, update=None, desk_id=None, stage_id=None, user_id=None, d
     task = {"desk": desk_id, "stage": stage_id, "user": original_task.get("user") if user_id is None else user_id}
 
     if current_stage:
-        await apply_stage_rule(doc, update, current_stage, MACRO_OUTGOING)
+        await apply_stage_rule(doc, update, current_stage, MACRO_OUTGOING, macro_kwargs=macro_kwargs)
 
     if desk_id:
         # TODO-ASYNC[desks]: Use DesksResourceModel async service where when upgrading this module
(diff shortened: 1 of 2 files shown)


petrjasek:
can you do it on develop @BrianMwangi21 ? it's supposed to go to next version, will be later in release/3.4

superdesk/internal_destinations.py
```diff
@@ -81,11 +81,14 @@ async def handle_item_published(item, after_scheduled):
             item[SCHEDULE_SETTINGS] = {}
 
         new_item = deepcopy(item)
+        new_item["_internal_destination"] = {"desk": dest.get("desk"), "stage": dest.get("stage")}
```
petrjasek: I wonder why it's needed if the same data is already passed as kwargs, would it need to know it's an internal destination specifically? then we could add another param `internal_destination=dest`

BrianMwangi21: I needed a way for the macro to know the item came from an internal destination. The macro only receives the item, not the send_to data, so it can’t tell unless we mark it.

The _internal_destination flag is a small, localized way to carry that context because changing the macro execution flow across core would be a much bigger change.

Or could I be missing something?

petrjasek: I don't like adding additional data to item which is not related to the content, would be better to pass it to the macro as additional params. if there was typing added this wouldn't pass

apps/tasks.py
```diff
@@ -70,7 +70,9 @@ def compare_dictionaries(original, updates):
     return modified
 
 
-async def send_to(doc, update=None, desk_id=None, stage_id=None, user_id=None, default_stage="incoming_stage"):
+async def send_to(
+    doc, update=None, desk_id=None, stage_id=None, user_id=None, default_stage="incoming_stage", macro_kwargs=None
+):
```
petrjasek: can you add a return type to get this type checked?

petrjasek (approved):

PR #3158, opened 2026-02-26
add config for amazon s3 media id time prefix

petrjasek:
with options daily | hourly | none.

SDBELGA-1045

### Steps to test

Resolves: #[issue-number]


eos87 (approved):
Looks good. One of the tests run was cancelled for some reason.

PR #3163, opened 2026-03-06
[STT-1587] - Allow show_floating_count editor flag

BrianMwangi21:
### Purpose
Remove the hardcoded EDITOR_ATTRIBUTES whitelist validation from content profiles. This allows the frontend to add new editor configuration attributes without requiring backend changes.

### What has changed
- Removed the EDITOR_ATTRIBUTES constant that restricted which editor attributes could be saved
- Removed the clean_editor() function that stripped out "unknown" editor attributes
- Editor attributes are now passed through transparently from client to database

### Steps to test
1. In Content Profiles, enable Show Floating Character Count for an Editor3 field (e.g. body_html) and ensure the field has a character limit set.
2. Save the profile.
3. Verify the profile saves correctly without the backend stripping out the showFloatingCount attribute.
4. Open an article, add enough content to make the toolbar sticky, and confirm the floating character count appears.
5. Disable show_floating_count, save, and confirm the floating character count no longer appears when the toolbar is sticky.

Resolves: #[STT-1587](https://sofab.atlassian.net/browse/STT-1587)

[STT-1587]: https://sofab.atlassian.net/browse/STT-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


petrjasek:
might be good to get rid of this validation altogether, don't see much value in it

tests/content_types_test.py
```diff
@@ -126,3 +126,36 @@ async def test_get_output_name_for_text_profile(self):
 
             result = service.get_output_name("text")
             self.assertEqual(result, "Article")
+
+    @mock.patch.object(content_types, "get_fields_map_and_names", lambda: ({}, {}))
+    async def test_unknown_editor_attributes_preserved(self):
```
petrjasek: would rather add a behave test to check it end to end

petrjasek (approved):

PR #3170, opened 2026-03-12
[SDBELGA-1051] improve(s3-async): Use ObjectId and store filename in metadata

MarkLark86:
### Purpose
Using filename for the ID deviates from the way Mongo GridFS uses ObjectIDs. This can cause issues for other system when/if the system changes from GridFS to S3.

### What has changed
* Use ObjectIDs for IDs when using Amazon S3 storage (still support folder and version prefixes)
* Store the filename in the metadata of the S3 object (similar to how GridFS stores the filename in it's fs.files collection)
* Still support downloading files if they're using the older filename approach

Resolves: [SDBELGA-1051](https://sofab.atlassian.net/browse/SDBELGA-1051)
Note: This is the async version from https://github.com/superdesk/superdesk-core/pull/3169

[SDBELGA-1051]: https://sofab.atlassian.net/browse/SDBELGA-1051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/storage/amazon_media_storage.py
+++ b/superdesk/storage/amazon_media_storage.py
@@ -153,11 +153,19 @@ class AmazonMediaStorage(SuperdeskMediaStorage):
         return unidecode.unidecode(str(_id)).translate(get_translation_table())
 
     def media_id(self, version: bool | str = True) -> str:
-        """Get the ``media_id`` path for the given ``filename``.
+        """
+        Generates a media ID string using a configurable time-based prefix and a unique identifier.
 
-        if filename doesn't have an extension one is guessed,
-        and additional *version* option to have automatic version or not to have,
-        or to send a `string` one.
+        This method generates a media ID based on a configurable time granularity or provided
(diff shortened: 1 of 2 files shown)


superdesk/storage/amazon_media_storage.py
```diff
@@ -173,7 +169,7 @@ def media_id(self, filename, content_type=None, version=True):
         else:
             version = "%s/" % version.strip("/")
 
-        return "%s%s%s" % (version, self._make_s3_safe(filename), extension)
+        return f"{version}{ObjectId()}"
```
petrjasek: I'm just wondering if we should ditch the extension, and if we should use some other uuid method to generate the id that wouldn't match mongo objectid, it might get eventually converted to ObjectId in some places when the prefix is not used, but not sure if it's a problem

petrjasek (approved):

PR #3174, opened 2026-03-17
disable elastic-apm

petrjasek:
it doesn't support quart monitoring

### Steps to test

Resolves: #[issue-number]


MarkLark86:
@petrjasek Do we still need Elastic-APM, with the current observability stack available?
There is an [asgi-middleware](https://www.elastic.co/docs/reference/apm/agents/python/asgi-middleware) available if we do need it, but might require some work to create our own Quart wrapper around it

petrjasek:
we don't need it atm, sentry can handle performance monitoring too

MarkLark86 (approved):

PR #3175, opened 2026-03-18
[SDCP-1117] Add JWT token authentication for download endpoint

eos87:
### Purpose
Allow download URLs to work across subdomains by supporting JWT token authentication.

### What has changed
- Add `blueprint_auth_or_token` decorator that accepts either a JWT token query param or session auth
- Add `generate_url_with_token` helper in `superdesk/auth/utils.py` to generate signed URLs
- Update `/download` endpoint to use the new decorator
- Update export service to append JWT token to download URLs (7-day expiry)
- Add unit and behave tests

This enables export download URLs to work when the API and client are on different subdomains (e.g., `server1-uat.superdesk.pro` and `server1-uat-api.superdesk.pro`) without requiring cross-subdomain session cookies.

Resolves: [SDCP-1117]

[SDCP-1117]: https://sofab.atlassian.net/browse/SDCP-1117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/download.py
+++ b/superdesk/download.py
@@ -11,13 +11,15 @@
 """Download module"""
 import logging
 import superdesk
+
 from superdesk.errors import SuperdeskApiError
-from superdesk.auth.decorator import blueprint_auth
-from .resource import Resource
+from superdesk.auth.decorator import blueprint_auth_or_token
 from superdesk.eve_async import AsyncBaseService
 from superdesk.core import get_current_app, get_config
 from superdesk.flask import request, Blueprint, url_for
 from superdesk.storage.superdesk_file import generate_response_for_file, get_file_request_range
+
+from .resource import Resource
 
 bp = Blueprint("download_raw", __name__)
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


superdesk/download.py
```diff
@@ -11,21 +11,23 @@
...
+from .resource import Resource
+
 bp = Blueprint("download_raw", __name__)
 logger = logging.getLogger(__name__)
 
 
 @bp.route("/download/<id>", methods=["GET"], defaults={"folder": None})
 @bp.route("/download/<path:folder>/<id>", methods=["GET"])
-@blueprint_auth()
+@blueprint_auth_or_token()
```
petrjasek: I wonder if we would need both session and token, imo just token would be enough

petrjasek (approved):

superdesk/download.py
eos87: I thought so too. I removed the session part, now token would be required. Please take another look. Thanks

PR #3176, opened 2026-03-19
[STT-1640] fix(ws): Set socket keepalive & timeout options

MarkLark86:
### Purpose
The connection to Redis from our websocket server has reliability issues. Connection timeouts and half closed connections.

### What has changed
* Set TCP Keepalive to make sure the connection stays alive with traffic
* Set socket read/write timeout for 2 mintues

Resolves: [STT-1640](https://sofab.atlassian.net/browse/STT-1640)

[STT-1640]: https://sofab.atlassian.net/browse/STT-1640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/websockets_comms.py
+++ b/superdesk/websockets_comms.py
@@ -44,6 +44,7 @@ from superdesk.default_settings import (
     WS_REDIS_KEEPALIVE_COUNT,
     WS_REDIS_CONNECT_TIMEOUT,
     WS_REDIS_TIMEOUT,
+    WS_DEBUG,
 )
 
 
@@ -51,8 +52,12 @@ logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
-logging.getLogger("websockets").setLevel(logging.INFO)
-setup_logging(logging.INFO)
+if WS_DEBUG:
+    logging.getLogger("websockets").setLevel(logging.INFO)
+    setup_logging(logging.INFO)
+else:
+    logging.getLogger("websockets").setLevel(logging.WARNING)
+    setup_logging(logging.WARNING)
 
 
 class SocketBrokerClient:
@@ -272,7 +277,7 @@ class SocketCommunication:
         *,
         sentry_dsn: Optional[str] = None,
         sentry_traces_sample_rate: Optional[float] = None,
-        debug: bool = False,
+        debug: bool | None = None,
     ):
         self.host = host
         self.port = int(port)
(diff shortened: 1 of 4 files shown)


superdesk/websockets_comms.py
```diff
@@ -77,25 +77,36 @@ def connect(self):
...
         self.connection = Connection(
             self.url,
             heartbeat=WS_HEART_BEAT,
             transport_options={
-                "socket_connect_timeout": env("WS_REDIS_CONNECT_TIMEOUT", 2),
+                "socket_connect_timeout": float(env("WS_REDIS_CONNECT_TIMEOUT", 2)),
                 # If no message has been consumed in 10 minutes, allow health check to run
                 # otherwise consumer loop hangs indefinitely on read timeout
-                "socket_timeout": env("WS_REDIS_TIMEOUT", 600),
+                "socket_timeout": float(env("WS_REDIS_TIMEOUT", 600)),
```
petrjasek: I'm still not sure about the default here, seems like it would be only used when it sends the health check and then it would wait 10m instead of failing fast

MarkLark86: I believe its used in both the socket level and redis health checks.
https://github.com/redis/redis-py/blob/master/redis/connection.py#L1507

petrjasek: yep it is used for socket operations, but imo not getting a message won't trigger the timeout. seems like that was an issue in the past https://github.com/redis/redis-py/issues/225

MarkLark86: You're right. Though Kombu doesn't use that `listen` method, it does implement it's own iteration loop which is run every second by default.

I have use that `on_iteration` of theirs to implement a heartbeat back to the main thread. As I noticed when Kombu get's stuck, that `on_iteration` callback is never called. Normally this gets called every second, so if after 1 minute of nothing, we restart the Kombu thread.

Please have a look and let me know

MarkLark86:
@petrjasek I reverted that Kombu thread heartbeat, will implement in another PR. For now I focused on the transport params so we can get this merged and released.

The socket read/write is now set to 2mins as you suggested earlier

petrjasek:
not sure about those 2m, it might be too much https://redis.io/docs/latest/develop/clients/redis-py/produsage/#timeouts
seems like by default it would be 10s

petrjasek (approved):

PR #3177, opened 2026-03-19
chore: Bump Celery to latest 5.6 release

MarkLark86:
### Purpose
Celery 5.6 fixes some memory leaks when exceptions are raised, and also allows us to use a newer Redis library


petrjasek:
it might be good to first update the requirements in repos using develop, otherwise CI will start failing everywhere we use develop

petrjasek (approved):

PR #3180, opened 2026-03-23
fix behave test not failing when expecting empty list

petrjasek:
when asserting there is `"subject": []` it would pass with non-empty list.


eos87 (commented):
some behave tests are failing

petrjasek:
it's working after merging develop

eos87 (approved):

PR #3184, opened 2026-03-25
improve(s3): Improve exists by throwing exceptions

MarkLark86:
### Purpose
When checking if an item exists in S3, we were catching all errors and treating them if the item doesn't exist.

### What has changed
* Use the specific `ClientError` exception to catch specifically a 404, everything else should raise an exception

See: https://github.com/boto/boto3/issues/2442

### DIFF
--- a/superdesk/storage/amazon_media_storage.py
+++ b/superdesk/storage/amazon_media_storage.py
@@ -22,6 +22,7 @@ from io import BytesIO
 
 from os.path import splitext
 from urllib.parse import urlparse
+from botocore.exceptions import ClientError
 from botocore.client import Config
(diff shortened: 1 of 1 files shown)


petrjasek:
not sure what might be the side effects if we raise an exception now, might be good to log it to begin with, and maybe raise it only when `SUPERDESK_DEBUG` is enabled

MarkLark86:
> not sure what might be the side effects if we raise an exception now, might be good to log it to begin with, and maybe raise it only when `SUPERDESK_DEBUG` is enabled

Fair point. I have updated it as you suggested

petrjasek (approved):

PR #3187, opened 2026-03-26
improve(s3): Add timeout & retry settings

MarkLark86:
### Purpose
Previously we had no way of changing the S3 connection/read timeout and retry settings.

### What has changed
* Add and use the following settings/env. vars:
  * AMAZON_S3_CONNECT_TIMEOUT
  * AMAZON_S3_READ_TIMEOUT
  * AMAZON_S3_MAX_RETRIES
* Added another logger.exception call that was missing
* Use a non-async method to check if the client has been created yet or not (in case of race condition in the async event loop)

### Note
* Although `tcp_keepalive` is enabled, this will require sys-team to make changes on the OS level to make them effective, as boto3 doesn't provide a way to configure this on a per app basis. See https://github.com/boto/botocore/issues/2916
* I also set defaults for these 3 values in both `default_settings` and `amazon_media_storage` so we won't need to make changes in Newshub

### DIFF
--- a/superdesk/storage/amazon_media_storage.py
+++ b/superdesk/storage/amazon_media_storage.py
@@ -91,8 +91,6 @@ class AmazonMediaStorage(SuperdeskMediaStorage):
         super().__init__(app)
         self.client = boto3.client("s3", **self._get_connection_kwargs(False))
         self.session_async = aioboto3.Session()
-        self._client_async_context = None
-        self.client_async = None
         self.user_metadata_header = "x-amz-meta-"
 
     def _get_connection_kwargs(self, for_async: bool) -> dict:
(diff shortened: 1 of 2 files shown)


superdesk/storage/amazon_media_storage.py
```diff
@@ -196,8 +223,9 @@ async def call_async(self, method, **kw):
         if "Key" in kw:
             kw["Key"] = self.get_key(kw["Key"])
 
-        if not self.client_async:
-            self.client_async = await self.session_async.client("s3", **self.connection_kwargs).__aenter__()
+        if self._client_async_context is None:
+            self._client_async_context = self.session_async.client("s3", **self._get_connection_kwargs(True))
+            self.client_async = await self._client_async_context.__aenter__()
 
         return await getattr(self.client_async, method)(**kw)
```
petrjasek: I wonder if it would be safer for now to just use the context when calling the method, I don't think it would be a problem performance wise and could help with stability

PR #3194, opened 2026-04-10
fix ap ingest: ignore failing association parse errors

petrjasek:
Wrap each association parse in a try/except so a malformed association item does not abort the whole document parse. Failed associations are logged as warnings and skipped.

SDCP-1145

### Steps to test

Resolves: #[issue-number]


eos87 (approved):
Looks. Some tests are failing though.

petrjasek:
yep these were failing before on this branch

PR #3200, opened 2026-04-20
add clean up scripts

petrjasek:
@MarkLark86 won't be merging this, but can you pls check it if you see any issue

looks like it would clean ~4M+ docs in sd, ~2M in content api

### Steps to test

Resolves: #[issue-number]


scripts/mongo/remove_stale_archive_items.js
```diff
@@ -0,0 +1,165 @@
...
+        item_id: { $exists: true, $ne: null },
+        last_published_version: true,
+    };
+
+    var publishedDocs = db
+        .getCollection("published")
+        .find(lookup, { _id: 1, item_id: 1 })
+        .sort({ _id: 1 })
+        .limit(BATCH_SIZE)
+        .toArray();
```
MarkLark86, line 96: Using `toArray` will place all 1,000 items into memory. This might be okay if the machine has enough RAM, otherwise iterating the cursor would be needed.

scripts/mongo/remove_stale_archive_items.js
```diff
@@ -0,0 +1,165 @@
+/*
```
MarkLark86, line 1: Do we have javascript enabled in MongoDB for the desired instance?

MarkLark86 (approved):
Just a couple questions, otherwise looks good

scripts/mongo/remove_stale_archive_items.js
petrjasek: it's only _id and item_id, not full items

scripts/mongo/remove_stale_archive_items.js
petrjasek: yep seems working

PR #3211, opened 2026-05-05
[SDESK-7907] Localize user-facing strings with gettext

eos87:
### Purpose
Wrap user-facing error and status strings with `quart_babel.gettext` (alias `_`), add required imports, and replace hardcoded messages across multiple modules (archive, media, auth, ldap, content_api, core resources/services, upload, text_checkers, users, vocabularies, etc.).

[SDESK-7907]

[SDESK-7907]: https://sofab.atlassian.net/browse/SDESK-7907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/media/crop.py
+++ b/superdesk/media/crop.py
@@ -10,6 +10,7 @@
 import json
 
 from eve.utils import ParsedRequest
+from quart_babel import gettext as _
 
 import logging
 from copy import deepcopy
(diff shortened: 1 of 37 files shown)


superdesk/media/crop.py
```diff
@@ -123,27 +124,27 @@ def _validate_aspect_ratio(self, crop, doc):
...
+                    raise SuperdeskApiError.badRequestError(message=_("Wrong aspect ratio!"))
         elif crop.get("ratio"):
             ratio = crop.get("ratio")
             if type(ratio) not in [int, float]:
                 ratio = ratio.split(":")
                 ratio = int(ratio[0]) / int(ratio[1])
             if abs((width / height) - ratio) > 0.1:
                 raise SuperdeskApiError.badRequestError(
-                    message="Ratio %s is not respected. We got %f" % (crop.get("ratio"), abs((width / height)))
+                    message=_("Ratio %s is not respected. We got %f") % (crop.get("ratio"), abs((width / height)))
```
MarkLark86: I see a few places where modulo interpolation is used. Does Babel handle these correctly when it extracts the details (especially given word order differences between languages)?

eos87: Babel handles this fine. As for the word order, yeah I think it's probably better to use `.format()` with named placeholders. I'll do the change.

eos87: done

MarkLark86 (approved):

PR #3218, opened 2026-05-14
Update positions of other items affected by moving/adding/removing

IvanJelicSF:
## Summary

Bulk patches on `/content_lists/<id>/items` now keep positions contiguous and unique when items are added, moved, or removed. Previously, only the directly-touched item's `position` was updated, which could leave gaps after deletes or produce duplicate positions on inserts/moves.

The new behavior in `apps/content_lists/service.py`:

- **Add** at position `p`: shift every non-sticky item with `position >= p` up by 1 before inserting.
- **Move** from `old` to `new`: items strictly between the two endpoints shift by 1 in the opposite direction; the moved item itself is excluded from the shift.
- **Delete** at position `p`: shift every non-sticky item with `position > p` down by 1.
- Sticky items (`sticky: true`) are excluded from all reordering.
- Items with `position: None` are handled — add/delete still works, and a move from/to `None` degrades to a one-sided shift.

## Test plan

Four new behave scenarios in `features/content_lists.feature` exercise each path end-to-end via the bulk-patch endpoint and assert the final ordering:

- [ ] Move an item to an occupied position — other item shifts to keep positions unique
- [ ] Add an item at an occupied position — existing items shift down
- [ ] Move an item forward across several positions — items in between shift back by one
- [ ] Delete an item — items after it shift up to close the gap

Fixes: [SDESK-7928](https://sofab.atlassian.net/browse/SDESK-7928)

[SDESK-7928]: https://sofab.atlassian.net/browse/SDESK-7928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/content_lists/service.py
```diff
@@ -77,3 +85,54 @@ async def bulk_patch(self, list_id: ObjectId, data: dict) -> ContentList:
...
+            {"list_id": list_id, "sticky": {"$ne": True}, "position": {"$gte": position}},
+            {"$inc": {"position": 1}},
+        )
+
+    async def _shift_positions_down(self, list_id: ObjectId, position: int) -> None:
+        """Close the gap at ``position`` by decrementing the position of every item after it."""
+        await self.mongo_async.update_many(
+            {"list_id": list_id, "sticky": {"$ne": True}, "position": {"$gt": position}},
+            {"$inc": {"position": -1}},
+        )
```
eos87: what are the chances we need to enable elastic search in this resource? if we would need to, then we should probably use the service method (`self.update_many(...)`) instead of `mongo_async` directly, so this won't bypass ElasticSearch. Although, I'm not sure the `$inc` mechanism would work the same in Elastic.

apps/content_lists/service.py
```diff
@@ -77,3 +85,54 @@ async def bulk_patch(self, list_id: ObjectId, data: dict) -> ContentList:
...
+        if new_position < old_position:
+            await self.mongo_async.update_many(
+                {
+                    "list_id": list_id,
+                    "sticky": {"$ne": True},
+                    "_id": {"$ne": item_id},
+                    "position": {"$gte": new_position, "$lt": old_position},
+                },
+                {"$inc": {"position": 1}},
+            )
```
eos87: same here

eos87 (commented):
Looks good in general, just two concerns:

1. Atomicity: each action does a shift + a create/update/delete as two separate Mongo calls. If the second call fails after the shift succeeds, the list ends up in a broken state? 

3. Sticky items: I see they are being excluded intentionally but, we need to make sure the sticky items get their `position` value properly set. Like: when an item becomes sticky, set its position to `None`. When an item becomes non-sticky, it needs a real position assigned (probably appended to the end, or supplied by the client). Thoughts?

apps/content_lists/service.py
IvanJelicSF: Thanks for checking it out. Sticky items will follow, so yes, they are intentionally omitted from this commit. Not sure about Elastic, didn't consider it in this "PoC" phase, but it's a very good observation. On one hand, I see Content lists as an internal thing, but on the other, we'll need to deliver them pretty much on a same scale as articles. Let's think about it more and keep the PR open for now, as I have to think about the Mongo comment you made as well.

apps/content_lists/service.py
```diff
@@ -61,19 +63,59 @@ async def bulk_patch(self, list_id: ObjectId, data: dict) -> ContentList:
...
+            return (
+                pos is None,
+                pos if pos is not None else 0,
+                rank is None,
+                -rank if rank is not None else 0,
+                d["_id"],
+            )
+
+        docs.sort(key=sort_key)
+
```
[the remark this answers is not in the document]
IvanJelicSF: This is actually not valid as the feedback is not considering the way the client posts to the API. The most efficient workflow means client and the API working on cohesion regarding the type and structure of the payload for all actions

features/content_lists.feature
```diff
@@ -159,6 +159,172 @@ Feature: Content Lists
...
+        """
+
+    @auth
+    Scenario: Moving an item to a higher position shifts items in between
+        Given an archive item "article-a"
+        And an archive item "article-b"
+        And an archive item "article-c"
+        And an archive item "article-d"
+        When we post to "/content_lists"
+        """
```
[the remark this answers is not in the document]
IvanJelicSF: This actually never happens on the client side, but will be considered in the future. Sticky position is not used in manual content lists and will be removed from the dataset.

eos87 (approved):
Please check the CI tests failing. Thanks!

PR #3219, opened 2026-05-15
implement lifecycle timing tracking

petrjasek:
track time it takes to process article, either from ingest to transmission,
or from publishing to transmission for manually created articles.

SDCP-1157

### Steps to test

Resolves: #[issue-number]

### DIFF
--- a/apps/archive/common.py
+++ b/apps/archive/common.py
@@ -29,6 +29,7 @@ from superdesk.utc import utcnow, get_expiry_date, local_to_utc, get_date
 from superdesk import get_resource_service
 from superdesk.metadata.item import (
     metadata_schema,
+    lifecycle_timing_schema,
     ITEM_STATE,
     CONTENT_STATE,
     LINKED_IN_PACKAGES,
@@ -131,6 +132,7 @@ ARCHIVE_SCHEMA_FIELDS = {
     },
     "expiry_status": {"type": "string", "mapping": not_analyzed, "nullable": True},
     "original_id": {"type": "string", "mapping": not_analyzed},
+    "lifecycle_timing": lifecycle_timing_schema,
 }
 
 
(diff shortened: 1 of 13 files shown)


MarkLark86:
@petrjasek I'm not a fan of having those stats stored directly on the item. We already have stats storage in the Analytics module (if `ANALYTICS_ENABLE_ARCHIVE_STATS` is enabled).
See the [feature tests](https://github.com/superdesk/superdesk-analytics/blob/develop/server/features/archive_history_stats/timeline_stats.feature#L40) and [schema](https://github.com/superdesk/superdesk-analytics/blob/develop/server/analytics/stats/archive_statistics.py#L46).

If we use the stats feature, we can have the data and code over in the analytics module.

The extra information that's generated in this PR can then be stored in stats resource as well.
We may need new signals, such as:
* PublishConsumer: on_transmit
* Generic Ingest Lifecycle: like ETL lifecycles

Then using other signals, such as:
* `app.on_<insert|update>_archive_publish`
* `PublishQueueResource.get_signals().data.on_<create|update>`

You can also use [stats_type](https://github.com/superdesk/superdesk-analytics/blob/develop/server/analytics/stats/archive_statistics.py#L49) to store ingest data separately, and then merge that data into the newly fetched archive item.

You should then be able to hook into the same lifecycle in this PR. The data will be available in stats, and later available for front end widgets/charts/data access.

petrjasek:
would consider that for later, we should probably decide what we want in analytics and what stats/metrics we should have elsewhere, I see this more like system processing info, but analytics are more editorial/workflow oriented, but in theory this could be combined as well.

I see the benefit in have the basic timing on items so you can filter them based on processing time now as we're trying to figure out why specific items take longer.

MarkLark86 (approved):

PR #3225, opened 2026-05-29
[SDESK-7917] Feature: Celery async worker runtime

MarkLark86:
### Purpose
Originally celery workers run their tasks procedurally per thread/process, not starting the next task until the current one is finished. This causes a performance bottleneck, especially when many tasks are queued and async I/O is in use.

### What has changed
* Create `CeleryAsyncWorkerThread` thread class
  * tasks are executed on a separate asyncio event loop
  * limits the number of concurrent tasks on the asyncio event loop, and automatically continued once load has stabilized
  * When stopping the thread, makes sure to wait for the current tasks to complete
* Create `CeleryAsyncWorkerTask` celery task, extending from our existing ones, so `eager` still works with the newer runtime.
* Celery hooks, so the worker thread is started/stopped when required
* Test celery worker app

Resolves: [SDESK-7917](https://sofab.atlassian.net/browse/SDESK-7917)

[SDESK-7917]: https://sofab.atlassian.net/browse/SDESK-7917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarkLark86:
@petrjasek @eos87 I created a [test instance](https://sdpr-4158.test.superdesk.org/) so we can test it out before merging. So far publishing seems to work :smile: 

I can confirm it's using the new async worker thread, from from the [log entry](https://test.superdesk.org/logs/3/www/sdpr-4158/main.log):
```
May 29 05:23:47 sdpr-4158 sh[106]: 05:23:47 work.1 | on_worker_init[250] level=INFO process=ForkPoolWorker-1
```

docs/settings.rst
```diff
@@ -287,6 +287,42 @@ Default: ``None``
...
+
+The maximum number of active tasks in the async thread. If more tasks are added, task submission
+to the thread is paused (as the client level).
+
+``CELERY_ASYNC_THREAD_RESTART_TASKS``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Default: ``20``
+
+The maximum number of active tasks in the async thread, before task submission is resumed.
```
eos87, line 317: what determines these values? I'd assume the hardware but what's the relation?

superdesk/tests/worker.py
eos87, line 1: Let's rename this one to `worker_test.py` so it is easier to distinguish

tests/celery_app/async_worker_test.py
eos87, line 1: same with this one please `async_worker_test.py‎`

eos87 (approved):
Code looks good to me. Left a couple of minor comments.

superdesk/tests/worker.py
```diff
@@ -0,0 +1,60 @@
...
+            return pong is not None
+        except Exception:
+            return False
+
+    while True:
+        if is_celery_ready():
+            print("Celery is now ready!")
+            break
+        sleep(0.1)
+
```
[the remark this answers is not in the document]
eos87: not sure how likely this would happen but it could indeed hang the tests indefinitely

docs/settings.rst
MarkLark86: The default values are ones that I came up with. It is something we're going to have to try out, and might need different values depending on the environment (hardware capabilities, tasks/sec etc).

They're used for backpressure mitigation of the asyncio event loop. Without this, the event loop could become overloaded slowing down all tasks on the loop, essentially allowing unlimited number of concurrent tasks in the thread.

petrjasek (approved):

MarkLark86:
@petrjasek Not sure why, but there's a few email based tests failing in Python v3.10, though they pass in v3.12.
I ran out of time, could you have a look for me please?

PR #3231, opened 2026-06-10
Webhook and autorefresh support for Content lists

IvanJelicSF:
# Content Lists: Webhooks & Auto-refresh

This PR adds two notification mechanisms to the `content_lists` app:

- **Auto-refresh** — internal websocket notifications so open clients live-refresh when a list or its items change.
- **Webhooks** — external HTTP callbacks so third-party systems are notified when a list's items change.

Both are emitted side-by-side from the items `bulk_patch`; auto-refresh is in-app and synchronous, webhooks are out-of-band via Celery.

---

### Summary
Emit websocket push notifications whenever a content list or its items change, so every open client auto-refreshes the affected list (and the editor's own other tabs stay in sync) without a manual page reload.

### Background
Content lists are edited by multiple users and across multiple browser tabs. Previously a client only saw another user's edits after a manual refresh. Superdesk already has a websocket channel (`superdesk.notification.push_notification`); this work wires the content lists services into it.

### What was implemented
All changes in `apps/content_lists/service.py`.

**List-level events** — `ContentListsService`
- `content_list:created` — from `on_created`, one notification per new list, carrying `_id`.
- `content_list:updated` — from `on_updated` when list metadata (name, limit, …) changes, carrying `_id`.
- `content_list:deleted` — from `on_deleted`, carrying `_id`.

**Item-level event** — `ContentListItemsService`
- `content_list:items_updated` — from `bulk_patch` after an add/move/delete batch is persisted, carrying `list_id`.

### Out of scope / Notes
- Backend emission only; frontend subscribe/refresh handlers are a separate client-side ticket.
- Fire-and-forget over the existing websocket channel — no delivery guarantee or replay for offline clients.

---

### Summary
Add outbound webhook support so external systems are notified when a content list's items change. Subscribers are a first-class REST resource, and deliveries happen out-of-band via Celery so slow/failing endpoints never block the API request that triggered the change.

### Background
The auto-refresh work above covers internal clients. This adds **external** notifications: HTTP POST callbacks to third-party URLs registered by an operator.

### What was implemented

**1. Webhook resource & data model** — `apps/content_lists/models.py`
`ContentListWebhook` with:
- `url` (required, non-empty) — POST target
- `name` (optional) — operator label
- `enabled` (default `true`) — toggle without deleting
- `excluded_lists` (default `[]`) — list ids for which this webhook does **not** fire (opt-out model: webhooks fire for every list by default)

**2. REST endpoints** — `apps/content_lists/__init__.py`
- `content_list_webhooks` resource: `GET`/`POST` collection, `GET`/`PATCH`/`DELETE` item.
- Mongo index on `enabled`.

**3. Delivery pipeline** — `apps/content_lists/webhooks.py`
- `enqueue_webhook_deliveries(event, list_id)` — reads enabled, non-excluded webhooks and dispatches **one Celery task per webhook** (isolated retries — a failing endpoint never causes re-POST to endpoints that already succeeded).
- `deliver_content_list_webhook` Celery task — POSTs JSON `{"event": ..., "list_id": ...}`, 10s timeout, 30s soft limit, up to 3 retries with 60s backoff; logs and drops after exhausting retries.
- Payload encoded with `MongoJSONEncoder` to correctly serialize `ObjectId`/`datetime` (the Celery serializer re-casts hex ids back to `ObjectId` on the worker, which the default requests encoder cannot handle).

**4. Trigger** — `apps/content_lists/service.py`
- Webhooks fire on `content_list:items_updated`, emitted from the items `bulk_patch`.

**5. Tests & docs**
- Unit tests for payload serialization and enqueue filtering — `apps/content_lists/webhooks_tests.py`
- Behave CRUD scenarios — `features/content_lists.feature`
- OpenAPI spec — `specs/content_lists_openapi.yaml`

**Delivery behavior**
- [ ] When a list's items change (add/move/delete via bulk patch), an HTTP POST is sent to every enabled webhook whose `excluded_lists` does not contain that list id.
- [ ] Disabled webhooks (`enabled: false`) receive no delivery.
- [ ] A webhook that excludes the changed list receives no delivery for that list, but still receives deliveries for other lists.
- [ ] The POST body is valid JSON `{"event": "content_list:items_updated", "list_id": "<hex string>"}` with `Content-Type: application/json`; `list_id` is serialized as a hex string, not an ObjectId.

**Reliability / isolation**
- [ ] Delivery happens asynchronously (Celery) and does not block or fail the API request that triggered the change.
- [ ] Each webhook is delivered by an independent task; one endpoint failing does not affect delivery to others.
- [ ] A failing endpoint is retried up to 3 times (60s apart); after retries are exhausted the failure is logged and dropped (no infinite retry, no crash).

### Out of scope / Notes
- Webhooks fire **only** on item changes (`items_updated`), not on list metadata create/update/delete. List-level events would be a follow-up.
- No payload signing / HMAC and no per-webhook secret — endpoints cannot verify authenticity yet.
- No delivery audit log / operator-visible status (failures only hit server logs).

Delivers [SDESK-7951](https://sofab.atlassian.net/browse/SDESK-7951) [SDESK-7952](https://sofab.atlassian.net/browse/SDESK-7952)

[SDESK-7951]: https://sofab.atlassian.net/browse/SDESK-7951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/content_lists/models.py
```diff
@@ -33,3 +33,11 @@ class ContentListItem(ResourceModelWithObjectId):
     sticky: bool = False
     sticky_position: int | None = None
     enabled: bool = True
+
+
+class ContentListWebhook(ResourceModelWithObjectId):
+    url: Annotated[str, validate_not_empty()]
+    name: str | None = None
```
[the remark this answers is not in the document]
IvanJelicSF: We do validate `http://` and `https://` schemas, which is sufficient.

apps/content_lists/service.py
```diff
@@ -5,20 +5,34 @@
...
 
 
 class ContentListsService(AsyncResourceService[ContentList]):
     async def on_update(self, updates: dict[str, Any], original: ContentList) -> None:
         updates.pop("content_list_items_updated_at", None)
         await super().on_update(updates, original)
 
+    async def on_created(self, docs: list[ContentList]) -> None:
+        await super().on_created(docs)
+        # Let other open clients live-add the new list without a refresh.
```
eos87: nit:
```suggestion
```

apps/content_lists/service.py
```diff
@@ -5,20 +5,34 @@
...
+        for doc in docs:
+            push_notification("content_list:created", _id=str(doc.id))
+
+    async def on_updated(self, updates: dict[str, Any], original: ContentList) -> None:
+        await super().on_updated(updates, original)
+        # List metadata changed (name, limit, ...) — let open clients refresh it.
```
eos87: nit:
```suggestion
```

apps/content_lists/service.py
```diff
@@ -83,6 +97,15 @@ async def bulk_patch(self, list_id: ObjectId, data: dict) -> ContentList:
         )
         result = await lists_service.find_by_id(list_id)
         assert result is not None
+
+        # Tell every open client (including the editor's other tabs) that this
+        # list's items changed, so they can live-refresh instead of waiting for
+        # a manual page reload.
```
eos87: I feel like these comments are not needed.

apps/content_lists/webhooks.py
```diff
@@ -0,0 +1,66 @@
+import logging
+
+import requests
+from bson import ObjectId
+from eve.io.mongo import MongoJSONEncoder
```
eos87: I think one of the goals of the async framework was to get rid of `eve`. Do you think `from superdesk.json_utils import SuperdeskJSONEncoder` would work in this case? It inherits from `eve` but once we get rid of it, we'll simply modify that class and there won't be a reason to touch this part because of that once again. Thoughts?

apps/content_lists/__init__.py
```diff
@@ -49,8 +51,30 @@
...
+        ]
+    ),
+    rest_endpoints=RestEndpointConfig(
+        resource_methods=["GET", "POST"],
+        item_methods=["GET", "PATCH", "DELETE"],
+        auth=http_method_privilege_based_rules(
+            {
+                "POST": "content_lists",
+                "PATCH": "content_lists",
+                "DELETE": "content_lists",
```
eos87, line 73: should `GET` also be tight to the permission?

apps/content_lists/service.py
```diff
@@ -83,6 +97,15 @@ async def bulk_patch(self, list_id: ObjectId, data: dict) -> ContentList:
...
+        push_notification("content_list:items_updated", list_id=str(list_id))
```
eos87: we should extract this string `"content_list:items_updated"` into a constant variable so we use here and below in `enqueue_webhook_deliveries`

apps/content_lists/service.py
```diff
@@ -83,6 +97,15 @@ async def bulk_patch(self, list_id: ObjectId, data: dict) -> ContentList:
...
+
+        # Notify external subscribers (webhooks) of the item change out of band.
+        await enqueue_webhook_deliveries("content_list:items_updated", list_id)
```
eos87: I wondering if we should wrap this in a `try/except` to truly make it "fire-and-forget". I mean, if for some reason the broker/enqueue of the celery subtask fail it would break your http request. Instead, it should try, if it fails fine we log an error 


```suggestion
        try:
            await enqueue_webhook_deliveries("content_list:items_updated", list_id)
        except Exception:
            logger.exception("Failed to enqueue content list webhook deliveries for list %s", list_id)
```

eos87 (commented):
I left a few comments. One additional comment is that we should make translation ready the user facing strings.

apps/content_lists/webhooks.py
```diff
@@ -0,0 +1,66 @@
...
+    One task per webhook keeps retries isolated: a failing endpoint never
+    causes a re-POST to endpoints that already succeeded. After the retries
+    are exhausted the failure is logged and dropped.
+    """
+    try:
+        # Encode the body ourselves: the Celery serializer re-casts hex-string
+        # ids back into ObjectId on the worker, which the default requests JSON
+        # encoder cannot serialize. MongoJSONEncoder renders ObjectId/datetime.
+        body = MongoJSONEncoder().encode(payload)
+        response = requests.post(
```
MarkLark86: It would be good to use async requests, as this will hold up a Celery worker in the newer async runtime.

We, and other libraries we use, are using aiohttp and we have a [mixin class](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/web/http_client_session.py) to make it easier to use.

You could implement a small class here like the following:
```python
import aiohttp
from eve.io.mongo import MongoJSONEncoder
from superdesk.core.web import AsyncHttpClientSessionMixin


class WebhookContentListDelivery(AsyncHttpClientSessionMixin):
    http_timeout = aiohttp.ClientTimeout(connect=5, sock_read=10)

    async def post_list(self, url: str, payload: dict) -> None:
        # Encode the body ourselves: the Celery serializer re-casts hex-string
        # ids back into ObjectId on the worker, which the default requests JSON
        # encoder cannot serialize. MongoJSONEncoder renders ObjectId/datetime.
        body = MongoJSONEncoder().encode(payload)

        http_client = await self.http_session()
        async with http_client.post(url, json=body) as resp:
            resp.raise_for_status()
```

- The `http_timeout` class variable is automatically applied inside the mixin when the session is created
- The `"Content-Type": "application/json"` is automatically applied when passing in `json=`

specs/content_lists_openapi.yaml
```diff
@@ -0,0 +1,657 @@
+openapi: '3.0.2'
```
MarkLark86, line 1: How was this file generated, was it manual or automated somehow?
If it can help, we have an [openapi](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/openapi.py) module that helps to create specs from Pydantic models.

My concern with using OpenAPI specs (or anything like it), is how these docs can drift from the code. Meaning the code could change, new attributes/params added, but the docs aren't updated unless manually done so.

If we could automate this spec, that would be great

IvanJelicSF: Hey Mark, thanks for the feedback. Yes it was automated and mostly an experiment, so we'll handle it properly once this feature gets merged upstream :)

PR #3232, opened 2026-06-12
[SDESK-7943] Add archived unarchive command

BrianMwangi21:
### Purpose
This PR adds a new command that allows one to restore expired published articles from the archived store so they can be corrected and republished. This adds a CLI command for unarchiving single articles and packages.

### What has changed
- Added `archived:unarchive` CLI command.
- Restores archived article(s) back into `archive` and `published`.
- Handles composite packages atomically, including the individual package members.
- Preserves expiry on restored items using desk settings, config fallback, or a 7-day default.
- Added tests

### Steps to test
1. Expire a single archived article in MongoDB, reindex `archive`, then run `archive:remove_expired` to have it removed.
2. Confirm the article is removed.
3. Run `python manage.py archived:unarchive --item-id <article-id>`.
4. Confirm the article is back in `archive` and `published`.
5. Repeat the same flow with a package and verify the package children are restored too.

Resolves: [#SDESK-7943](https://sofab.atlassian.net/browse/SDESK-7943)

### DIFF
--- a/apps/archived/commands.py
+++ b/apps/archived/commands.py
@@ -1,0 +1,319 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+import logging
+from copy import deepcopy
+from datetime import timedelta
+
+import click
+from eve.utils import ParsedRequest
+
+from superdesk.commands import cli
+from superdesk.core import get_config
+from superdesk.metadata.item import CONTENT_STATE, CONTENT_TYPE, ITEM_TYPE, PUBLISH_STATES
+from superdesk.resource_fields import ETAG, ID_FIELD, VERSION
+from superdesk.utc import utcnow, get_expiry_date
+from superdesk.types.archived import QueueState
+from superdesk import get_resource_service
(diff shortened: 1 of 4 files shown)


apps/archived/commands.py
```diff
@@ -0,0 +1,319 @@
...
+@click.option("--item-id", "-i", multiple=True, required=True, help="Item ID(s) to unarchive")
+@click.option("--dry-run", "-d", is_flag=True, help="Simulate without making changes")
+async def cli_archived_unarchive(item_id, dry_run):
+    """Unarchive published articles so they can be edited/corrected.
+
+    Example:
+    ::
+
+        $ python manage.py archived:unarchive --item-id=urn:newsml:example.com:2023-01-01:1
+        $ python manage.py archived:unarchive -i item1 -i item2 --dry-run
```
petrjasek: can it work with positional args, might be easier, eg.:
```
 $ python manage.py archived:unarchive --dry-run item1 item2 
```

apps/archived/commands.py
```diff
@@ -0,0 +1,339 @@
...
+        try:
+            if desk_id:
+                expiry = await get_expiry(desk_id, stage_id)
+                if expiry:
+                    doc["expiry"] = expiry
+                    return
+        except Exception:
+            logger.warning("Failed to get expiry for desk %s stage %s", desk_id, stage_id)
+
+        doc["expiry"] = utcnow() + timedelta(days=DEFAULT_UNARCHIVE_EXPIRY_DAYS)
```
petrjasek: in general those items shouldn't expire, if they do they will go to archived again. would either set it much longer to begin with or set it to `null`. could be also cli param for the operation

petrjasek (approved):

PR #3234, opened 2026-06-17
fix: prevent publish queue items stuck in routing state

petrjasek:
SDCP-1179

### Steps to test

Resolves: #[issue-number]


MarkLark86 (approved):
Looks good

PR #3235, opened 2026-06-17
[SDESK-7945] - Resend action fails to get the destination credentials

BrianMwangi21:
### Purpose
This PR fixes publish queue resend failures when queued destination secrets were lost from the stored queue snapshot. Resend was reusing the queue item’s destination config as-is, so missing auth secrets caused `Auth error` on retry.

### What has changed
- Hydrates missing destination secret fields on publish queue create by looking up the live subscriber destination.
- Added a regression test

### Steps to test
1. Create or use a publish queue item whose `destination.config.password` is missing but the actual subscriber destination has the password.
2. Trigger resend from the publish queue UI.
3. Verify the resend succeeds.

Resolves: #[SDESK-7945](https://sofab.atlassian.net/browse/SDESK-7945)

[SDESK-7945]: https://sofab.atlassian.net/browse/SDESK-7945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/superdesk/publish/publish_queue.py
+++ b/superdesk/publish/publish_queue.py
@@ -99,63 +99,12 @@ class PublishQueueService(BaseService):
             if subscriber_id:
                 subscriber = subscriber_service.find_one(req=None, _id=subscriber_id)
 
-            self._hydrate_destination_secrets(doc, subscriber)
+            subscriber_service.hydrate_destination_secrets(doc.get("destination") or {}, subscriber)
             self._set_queue_state(doc, {})
             doc["moved_to_legal"] = False
 
             if "published_seq_num" not in doc and subscriber:
                 doc["published_seq_num"] = subscriber_service.generate_sequence_number(subscriber)
-
-    def _hydrate_destination_secrets(self, doc, subscriber):
-        destination = doc.get("destination") or {}
-        if not subscriber:
-            return
-
-        secret_fields = self._get_secret_fields()
-        subscriber_destination = self._match_subscriber_destination(destination, subscriber.get("destinations") or [])
-        if not subscriber_destination:
-            return
-
-        self._copy_destination_secrets(destination, subscriber_destination, secret_fields)
-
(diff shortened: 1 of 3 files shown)


superdesk/publish/publish_queue.py
```diff
@@ -96,48 +94,53 @@ def on_create(self, docs):
...
+        secret_fields = self._get_secret_fields()
         subscriber_destination = self._match_subscriber_destination(destination, subscriber.get("destinations") or [])
         if not subscriber_destination:
             return
 
-        self._copy_destination_secrets(destination, subscriber_destination)
+        self._copy_destination_secrets(destination, subscriber_destination, secret_fields)
+
+    def _get_secret_fields(self):
+        from superdesk.publish.subscribers import SECRET_FIELDS
```
petrjasek: if it's there to avoid circular dependency I would move it somewhere else and import it there from both. but that method to add missing secrets could be in subscribers service and be only called from here, so that whole secret handling is in one place

superdesk/publish/publish_queue.py
```diff
@@ -92,15 +92,34 @@ class PublishQueueResource(Resource):
 
 
 class PublishQueueService(BaseService):
+    def get(self, req, lookup):
```
petrjasek: afaik we don't want to hide secrets in publish queue, the problem was that those were not present

BrianMwangi21: Copilot flagged it, and it initially made sense since we also hide secrets on subscribers and maybe we just never got to hiding them in the queue too. But I have reverted that change.

petrjasek: yep makes sense in general, but wouldn't implement it now in the old version

petrjasek (approved):

PR #3236, opened 2026-06-18
[SDESK-7913] Replace flask-mail with aiosmtplib

MarkLark86:
### Purpose
Previously we used `flask-mail` for sending emails. This library does not use async I/O, and therefor will hold up a Celery Worker while sending an email.

### What has changed
* Remove `flask-mail` from dependencies
* Add `aiosmtplib` to dependencies
* Create new EmailFactory, that does not depend on other libraries or frameworks (no Flask required).
* Add `autoretry` to celery task for specific errors raised
* Create generic Email types in core
* Update `superdesk.emails` to use the new core EmailFactory (importing `send_mail` in this module also allows backwards compatibility in older code)
* Update email publish transmitter (mainly for the new EmailAttachment type)
* Create new MockEmailFactory, so we can record messages in tests and assert them later
* Fix/update tests

Resolves: [SDESK-7913](https://sofab.atlassian.net/browse/SDESK-7913)

[SDESK-7913]: https://sofab.atlassian.net/browse/SDESK-7913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarkLark86:
@petrjasek Tests are failing because Planning module is still using `flask_mail`, which is fixed in it's PR

MarkLark86:
@petrjasek some of the feedback focuses on Python 3.10 incompatibility.
Should we up our minimum version of Python to 3.12 for SD 3.6+

petrjasek:
we probably could, have to check with systems

petrjasek (approved):

eos87:
Since this was approved, I will merge because we need it to fix the build on CI.

PR #3271, opened 2026-07-20
avoid duplicates in publish queue

petrjasek:
when the task is cancelled check if the items
were already created and only restart it if
they are missing.

SDCP-1208

### Steps to test

Resolves: #[issue-number]


petrjasek:
it's not exactly clear what could be cancelling it, maybe request timeout or something, it's not visible in the logs

MarkLark86 (approved):
This looks good and should give us more insight into what's going on.
Good job 👍

PR #3283, opened 2026-08-05
[SDESK-7983] - Some feeding services are ingested but seem to be buggy

BrianMwangi21:
### Purpose
Ingested and uploaded pictures were saved with `renditions: null`, so images did not display. The async rendition task was dispatched from a request/ingest context where `_is_always_eager()` returns `True` (no bound `request.id`), forcing the eager branch that never retrieves the result — so renditions were dropped. 

### What has changed
- `superdesk/celery_app/context_task.py`: the dispatch decision in `apply_async` is now   config-driven via a new `_is_configured_always_eager()`. `_is_always_eager()` is unchanged and still governs the direct-call path, so calling a task directly stays  eager ( #3254  behaviour preserved). Tasks dispatched via `.delay()` now go to the  broker and their result is retrieved as intended.
- Added tests

### Steps to test
1. Set up an RSS ingest provider (e.g. BBC world feed) and run ingest.
   - Ingested picture items should now have populated `renditions`
     (`original`, `baseImage`, `thumbnail`, `viewImage`), not `null`, and thumbnails
     should render in the Ingest view.
2. In the editor, upload an image via the media icon, and separately drag an image from
   the Ingest list into `body_html`.
   - The image should render in the body; the created archive item's
     `associations.editor_0.renditions` should be fully populated.
3. Run the tests

Resolves: https://sofab.atlassian.net/browse/SDESK-7983

### DIFF
--- a/superdesk/celery_app/context_task.py
+++ b/superdesk/celery_app/context_task.py
@@ -122,13 +122,8 @@ class HybridAppContextTask(Task):
         # async with self.get_current_app().app_context():
         self.handle_exception(exc)
 
-    def _is_always_eager(self):
-        current_request = getattr(self, "request", None)
-        if current_request is None or not getattr(current_request, "id", None):
-            # If this task has not been bound to a request, then we're to process it directly
-            # This can happen when the task is called directly, instead of using `.delay` or `.apply_async`
-            return True
-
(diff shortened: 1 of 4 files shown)


MarkLark86 (approved):
Great work 🥇

PR #3287, opened 2026-08-14
Add Content lists feature

IvanJelicSF:
## Summary

Adds a **Content lists** feature (`apps/content_lists`) built on the async
framework, letting editors curate ordered lists of published content. The module ships three
resources — `content_lists`, `content_list_items`, and `content_list_webhooks` — and is
registered in `MODULES` in `superdesk/default_settings.py`.

### Content lists (`/content_lists`)
- CRUD REST endpoints for lists with `name`, `type` (`automatic` / `manual` / `bucket`),
  `description`, `limit`, `cache_life_time`, `filters`, and `enabled` fields.
- New `content_lists` privilege ("Manage Content Lists").
- `content_list_items_updated_at` timestamp maintained on the parent list whenever its items change.

### List items (`/content_lists/<id>/items`)
- Items reference `archive` content (validated data relation) and carry `position`, `sticky`,
  `sticky_position`, and `enabled`.
- A `bulk_patch` endpoint applies adds, updates, deletes, and moves in one request:
  - Positions of other items are renumbered automatically when items are added, moved, or removed.
  - Sticky items keep their pinned position across moves.
  - Duplicate content in the same list is rejected with a clear error (backed by a unique
    `(list_id, content)` index).
  - Add/delete pairs for the same content within one bulk request cancel out.
- Fetched items are enriched with article content (headline, thumbnail resolution including
  first `body_html` image fallback, etc.) for publisher-like consumption.

### Webhooks (`/content_list_webhooks`)
- CRUD for webhook subscriptions with `url`, `name`, `enabled`, and `excluded_lists`
  (webhooks fire for every list by default; listed ids are skipped).
- URLs are validated as HTTP(S) and stored in WHATWG-canonical form so equivalent spellings
  can't coexist as duplicates.
- Deliveries are enqueued as Celery tasks and posted via aiohttp (non-blocking); enqueue
  failures never fail the originating request.
- All webhook endpoints, including GET, require the `content_lists` privilege since webhook
  URLs may embed auth tokens.

### Notifications
- Websocket push notifications on content list and list item changes, including an
  `extension` field for client-side routing.

## Testing

- Behave scenarios covering CRUD, bulk operations, and webhooks (`features/content_lists.feature`, ~960 lines).
- Unit/REST tests for services, bulk patch, renumbering, article content enrichment,
  REST endpoints, and webhook delivery (`tests/content_lists/`).

## Docs

- OpenAPI spec for the new endpoints: `specs/content_lists_openapi.yaml`.

Resolves: [SDESK-7998](https://sofab.atlassian.net/browse/SDESK-7998)

[SDESK-7998]: https://sofab.atlassian.net/browse/SDESK-7998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/content_lists/service.py
```diff
@@ -0,0 +1,229 @@
...
+
+class ContentListItemsService(AsyncResourceService[ContentListItem]):
+    async def bulk_patch(self, list_id: ObjectId, data: dict) -> ContentList:
+        """Process a batch of add/move/delete actions on a content list's items.
+
+        Uses ``updatedAt`` in the payload for optimistic concurrency control instead of
+        the standard etag header. Updates ``content_list_items_updated_at`` directly via
+        MongoDB to bypass the ``on_update`` hook, which strips that field from all updates.
+        """
+        if not data:
```
MarkLark86: Beyond this initial key test I don't see much validation of this user provided actions. We could use Pydantic to provide this for us.

For example:
```python
from datetime import datetime
from enum import Enum
from superdesk.core.resources import BaseModel


class ContentListItemActionType(str, Enum):
    ADD = "add"
    DELETE = "delete"


class ContentListBulkAction(BaseModel):
    action: ContentListItemActionType
    contentId: str


class ContentListBulkActions(BaseModel):
    items: list[ContentListBulkAction]
    updatedAt: datetime


class ContentListItemsEndpoints(ResourceRestEndpoints):
    async def bulk_patch_items(self, request: Request) -> Response:
        ...
        data = ContentListBulkActions.from_dict(await request.get_json())
```

This would make sure that all the action data is validated before you start using it

MarkLark86 (commented):
Looks good, and thanks for using the new async functionality.
Just the 1 suggested improved from me, regarding validating client supplied request body.
 
I will let the others/copilot have a look over as well before approval.

specs/content_lists_openapi.yaml
```diff
@@ -0,0 +1,691 @@
+openapi: '3.0.2'
```
MarkLark86, line 1: Was this doc hand crafted, semi-automated or full-automated?

apps/content_lists/__init__.py
```diff
@@ -0,0 +1,83 @@
+from quart_babel import lazy_gettext as _
```
MarkLark86: I would probably also move this to a `apps/content_lists/module.py` file, and update the `MODULES` config to use that.

If any other part of the code needs something from this module (which it doesn't yet, but might in the future), `from apps.content_lists.<x> import <y>` would also import everything from this file (all configs, resource services, endpoints etc.

specs/content_lists_openapi.yaml
IvanJelicSF: Yeah, automated, but I went through it. It is actually based on the Publisher openapi spec as we have it over there for the whole API. I did it mostly for my own sake, seems I missed to note that

MarkLark86 (approved):

PR #3300, opened 2026-08-27
fix updating publish queue with old etag

petrjasek:
on error when setting retry

### Steps to test

Resolves: #[issue-number]


petrjasek:
@MarkLark86 pls review it and let me know, I'm merging it now so I can test it meanwhile

PR #3306, opened 2026-08-31
[SDESK-7994] Add the AI providers, actions and events module

eos87:
### Purpose

First production step of the "Superdesk goes AI" work ([SDESK-7994], [SDESK-7996]). Superdesk gets a generic way to connect language models and offer editorial AI actions: an admin registers a provider (OpenAI, OpenRouter, a self-hosted Ollama, anything speaking the OpenAI-compatible API) with their own key, defines actions like headline suggestions or a summary per content profile, and editors run them from the client with every run logged. Prompts, keys and provider calls all stay server side; the browser never talks to a model directly.

### What has changed

- New `superdesk.ai` module on the async framework, in `MODULES` by default. A deployment with no provider configured just has no actions to run.
- `ai_providers`: connection config with a write-only `api_key` (empty keeps it, null clears it, never returned by any response), an `available_models` shortlist, plus `POST .../test` and `GET .../models`.
- One provider client for the OpenAI-compatible chat completions protocol, over aiohttp, with typed error mapping (429/502/504) and an `AI_REQUEST_TIMEOUT` setting. It's also usable from server code directly, see the "Using a provider from code" section in the docs.
- `ai_actions`: per-action config (type, input/output fields, limits, content profiles, provider + model) and `POST .../run` that builds the prompt server side and returns suggestions. `suggestion` and `summary` for now; translation and rewrite come later.
- `ai_events`: a log entry per run (model, tokens, latency, user, desk, outcome reporting by the editor; never article text or keys). This is what the future AI dashboard reads from.
- Privileges: `ai_studio` to manage and read the log (equivalent to key access, documented), `ai` to run.
- Docs in `docs/ai.rst` with a curl walkthrough; behave features for the API contract; ~170 pytest tests.

### Steps to test

1. Follow the walkthrough in `docs/ai.rst`: register a provider (an OpenRouter free-tier key works, or point `base_url` at a local Ollama), test the connection, list its models.
2. Create a `suggestion` action for the Text profile and run it against any text item: you should get three headline suggestions back and an `ai_events` entry with token counts.
3. Report an outcome on the event (`accepted` + index) and check another `ai` user gets a 403 doing the same.
4. `pytest tests/ai` and `behave features/ai_providers.feature features/ai_actions.feature features/ai_events.feature` both green; provider HTTP is mocked, nothing touches the network.

Resolves: [SDESK-7994]

[SDESK-7994]: https://sofab.atlassian.net/browse/SDESK-7994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDESK-7996]: https://sofab.atlassian.net/browse/SDESK-7996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDESK-7994]: https://sofab.atlassian.net/browse/SDESK-7994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


superdesk/ai/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Configuration of the AI providers Superdesk talks to, and the actions built on top of them.
```
MarkLark86: If we keep this file empty (and use `superdesk.ai.module` in settings) it will improve issues with cyclic imports etc. Anything that imports `superdesk.ai.<x>` will currently import everything from `module` as well (which looks like maybe why the `TYPE_CHECKING` guard is needed in the `providers` sub-module)

superdesk/ai/models.py
```diff
@@ -0,0 +1,191 @@
...
+    TRANSLATION = "translation"
+
+
+#: Action types whose answers are a headline or a paragraph, short enough to keep on the run log
+#: next to the outcome a client reports later. A rewrite or a translation answers with the whole
+#: article, so storing its answers would make the log a second copy of the content.
+SHORT_OUTPUT_ACTION_TYPES = frozenset({AIActionType.SUGGESTION, AIActionType.SUMMARY})
+
+
+@dataclass
```
MarkLark86: I don't think we need the extra `@dataclass` decorator now, it should work with just the `Dataclass` base class

superdesk/ai/models.py
```diff
@@ -0,0 +1,191 @@
...
+    The article text is never part of an event, only ``input_chars`` and ``output_chars``. An entry
+    is written per run, so holding the text would make the log a copy of the content of every item
+    an action is run against.
+
+    ``user_id`` and ``desk_id`` are stored as text: they are copied from a session and from an item
+    rather than taken from a validated model, so the type they are stored with elsewhere is not
+    guaranteed. The action and provider ids keep the type they have on their own resources.
+    """
+
+    item_id: str
```
MarkLark86, line 149: This seems to be tying the AI functionality down to content only, where as I see possible uses elsewhere within the superdesk ecosystem. IMO it might be good to have content specifics inside a sub-module dedicated to content, and keeping these core models as generic as possible

superdesk/ai/prompts.py
```diff
@@ -0,0 +1,220 @@
...
+import json
+import re
+from typing import Any, Mapping
+
+from quart_babel import gettext
+
+from .errors import AIErrorKind, AIProviderError, NotSupportedError
+from .models import AIActionType
+
+SYSTEM_TEMPLATES: dict[AIActionType, str] = {
```
MarkLark86, line 17: This is what was mentioned in Show & Tell, rather than being hard-coded it would be good to be manageable through a resource (to be done later I suppose)

superdesk/ai/rest_endpoints.py
```diff
@@ -0,0 +1,204 @@
...
+from superdesk.users.async_service import get_privileges, is_admin
+
+from .errors import AIProviderError, to_api_error
+from .models import AIAction, AIEvent, AIProvider, RunActionPayload
+from .privileges import AI_PRIVILEGE, AI_STUDIO_PRIVILEGE
+from .providers import get_client
+from .service import AIActionsService
+
+
+def get_request_user_id(request: Request) -> str | None:
```
MarkLark86, line 20: These 2 functions seem quite generic, and could be better placed under `superdesk.core.web`

superdesk/ai/rest_endpoints.py
```diff
@@ -0,0 +1,204 @@
...
+                auth=[required_privilege_rule(AI_PRIVILEGE)],
+                parent=self,
+            )
+        )
+
+    async def run(self, request: Request) -> Response:
+        action = await self._get_action(request)
+
+        try:
+            payload = RunActionPayload.model_validate(await request.get_json() or {})
```
MarkLark86: I know `model_validate` is the standard Pydantic way, but we tend to use `from_dict` in our codebase

superdesk/ai/service.py
```diff
@@ -0,0 +1,490 @@
+import logging
```
MarkLark86: This file is big, it would be good to separate out the implementations into separate files

superdesk/default_settings.py
```diff
@@ -1144,6 +1145,9 @@ class PublishChannelConfig(TypedDict, total=False):
 #: default timeout for email connections
 EMAIL_TIMEOUT = 10
 
+#: seconds to wait for a response from an AI provider before the request is aborted
```
MarkLark86, line 1148: It would be good to include a mention in the docs about request timeout vs AI request timeout. Nginx/Hypercorn can timeout before the AI request times out.

MarkLark86 (commented):
Looks good 👍 Tested locally and got:
```json
{
    "event_id": "6a9612cfa1905c9eb8989cb3",
    "model": "qwen3.6:27b",
    "provider": "6a960dc22086ac99c8f416c8",
    "suggestions":[
        {
            "over_limit": false,
            "text": "Original Star Wars Trilogy Still the Cinematic Peak"
        },
        {
            "over_limit": false,
            "text": "First Three Star Wars Films Remain Franchise Peak"
        },
        {
            "over_limit": false,
            "text": "Original Star Wars Trilogy Sets Unmatched Standard"
        }
    ],
    "usage":{"input_tokens":3172,"output_tokens":37}
}
```

docs/ai.rst
```diff
@@ -32,3 +32,160 @@ So far, Superdesk supports the following AI Services:
...
+client writes an accepted answer to:
+
+.. code:: sh
+
+    curl -X POST $API/ai_actions -H "$AUTH" -H "$JSON" -d '{
+        "name": "Headline suggestions",
+        "action_type": "suggestion",
+        "input_fields": ["body_html"],
+        "output_field": "headline",
+        "suggestions_count": 3,
```
petrjasek: might be cleaner to have something like:
```
"action_params": {
    "suggestions_count": 3,
}
```
so we don't have to add new fields to api for every action

docs/ai.rst
```diff
@@ -32,3 +32,160 @@ So far, Superdesk supports the following AI Services:
...
+Add an action for it. ``input_fields`` are read from the item, ``output_field`` is the field the
+client writes an accepted answer to:
+
+.. code:: sh
+
+    curl -X POST $API/ai_actions -H "$AUTH" -H "$JSON" -d '{
+        "name": "Headline suggestions",
+        "action_type": "suggestion",
+        "input_fields": ["body_html"],
+        "output_field": "headline",
```
petrjasek, line 125: not sure if this shouldn't be a list as well, eg. suggesting headline and slugline in single call. or different metadata with single prompt

docs/ai.rst
```diff
@@ -32,3 +32,160 @@ So far, Superdesk supports the following AI Services:
...
+.. code:: sh
+
+    curl -X POST $API/ai_actions/68af1c0e2b1a4c0f9a3d1e02/run -H "$AUTH" -H "$JSON" -d '{
+        "item_id": "urn:newsml:localhost:2026-08-27T10:00:00:1234",
+        "source": "authoring"
+    }'
+
+    {
+        "suggestions": [
+            {"text": "Council backs the budget", "over_limit": false},
```
petrjasek, line 145: if we support multiple fields, this could use `field` param

features/ai_actions.feature
```diff
@@ -0,0 +1,229 @@
...
+            "input_fields": ["body_html"],
+            "output_field": "headline",
+            "max_characters": 60,
+            "provider": "#ai_providers._id#"
+        }
+        """
+        Then we get OK response
+        When we run the AI action at "/ai_actions/#ai_actions._id#/run"
+        """
+        {"item_id": "article-1", "source": "authoring"}
```
petrjasek, line 99: I would rather send the actual content to the api, otherwise it requires the current value user might be working on to be saved as a version on the server first

superdesk/ai/models.py
eos87: yeah I agree. Although I would keep it this simple for now until the next use case is clear, then we can adjust it accordingly.

superdesk/ai/prompts.py
eos87: agreed. I will keep it like this just for the MVP purposes and I will create a ticket with what I think it's missing or should be implemented next (this included).

superdesk/ai/service.py
eos87: split now

superdesk/ai/rest_endpoints.py
eos87: kept `get_request_user_id` here and moved `request_user_has_privilege` into `privilege_rules.py` that felt like a better home than `superdesk.core.web`

docs/ai.rst
eos87: yeap, good idea. Actually we already have `parameters` so I will put it there instead.

docs/ai.rst
eos87: kinda makes sense but it would introduce complexity. For example, how would you define in the action `max characters` per each output field? I guess we could add one row per field but I'm not sure. 

Image

petrjasek: it can probably get it from the content profile, but yep would be more complex

features/ai_actions.feature
eos87: actually this is already supported we just didn't add a proper test. Added in eaa2fed

docs/ai.rst
eos87: I did the content profile part now in b6421033: when the action names no `max_characters` it takes `maxlength` from the profile for the output field. It can be overridden from the action value. I'm leaving the multi-field output for another PR though. 
Please take another look when you have time.

petrjasek (approved):

PR #3316, opened 2026-09-09
[CPCN-1217] - Add default_max_results option to ResourceConfig

BrianMwangi21:
### Purpose
Allow a resource to set its default page size via `ResourceConfig`, applied to every search on that resource, instead of subclassing `ResourceRestEndpoints` and overriding `search_items`.

### What has changed
- Added `default_max_results` to `ResourceConfig`, next to `default_sort`.
- `SearchRequest.max_results` is now `int | None` and defaults to `None`, meaning "not set".
- `AsyncResourceService.find` resolves a missing `max_results` from the resource config, falling back to the elastic config default and then 25. The mongo path applies the same fallback, matching what the elastic client already does. An explicit `max_results` always wins, and resources without the option keep the current 25.
- The REST GET endpoint no longer handles the default itself; it passes `None` through and the service fills it in.
- Added tests covering the service and the REST API, and updated existing expectations for the new default.

### Steps to test
1. Set `default_max_results` on a resource's `ResourceConfig`.
2. Call `find` on its service without `max_results`, or GET the resource without a `max_results` param, with more items than the configured value. The number of items and `_meta.max_results` match the configured value.
3. Repeat with an explicit `max_results`. The explicit value is used.

Resolves : https://sofab.atlassian.net/browse/CPCN-1217

Works with : https://github.com/superdesk/newsroom-core/pull/1561

### DIFF
--- a/superdesk/core/resources/resource_rest_endpoints.py
+++ b/superdesk/core/resources/resource_rest_endpoints.py
@@ -115,9 +115,6 @@ class RestEndpointConfig:
 
     additional_lookup: AdditionalLookupConfig | None = None
 
-    #: Page size for the resource GET endpoint when the request has no ``max_results``, defaults to 25
-    default_max_results: int | None = None
-
 
 def get_id_url_type(data_class: type[ResourceModel]) -> str:
     """Get the URL param type for the ID field for route registration"""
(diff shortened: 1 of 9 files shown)


superdesk/core/resources/resource_rest_endpoints.py
```diff
@@ -115,6 +115,9 @@ class RestEndpointConfig:
 
     additional_lookup: AdditionalLookupConfig | None = None
 
+    #: Page size for the resource GET endpoint when the request has no ``max_results``, defaults to 25
+    default_max_results: int | None = None
```
MarkLark86: This would only apply default page size for API requests. In the ResourceConfig we have [default_sort](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/resources/resource_config.py#L48) already, maybe we can have `default_max_results` here instead.

This, though, would require changes to the [AsyncResourceService](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/resources/service.py#L837) and [SearchRequest](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/types/search.py#L153-L155) type.

This way a default max results would be applied across the entire resource

MarkLark86 (approved):

PR #3321, opened 2026-09-11
sanitize html in ninjs parser

petrjasek:
and remove class and style attributes

SDCP-1143

### Steps to test

Resolves: #[issue-number]

### DIFF
--- a/superdesk/io/feed_parsers/ninjs.py
+++ b/superdesk/io/feed_parsers/ninjs.py
@@ -17,6 +17,7 @@ from superdesk.io.registry import register_feed_parser
 from superdesk.io.feed_parsers import FeedParser
 from superdesk.utc import utc
 from superdesk.metadata.utils import generate_tag_from_url
+from superdesk.text_utils import sanitize_html
 from typing import Optional, Dict, List, Any
 from superdesk import get_resource_service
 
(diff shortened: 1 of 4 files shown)


superdesk/io/feed_parsers/ninjs.py
```diff
@@ -115,12 +120,9 @@ def _transform_from_ninjs(self, ninjs):
...
         for key, associated_item in ninjs.get("associations", {}).items():
             if associated_item:
-                self.items.append(self._transform_from_ninjs(associated_item))
-                if associated_item.get("type") == "picture" and associated_item.get("body_text"):
-                    associated_item["alt_text"] = associated_item.get("body_text")
-                if associated_item.get("versioncreated"):
-                    associated_item["versioncreated"] = self.datetime(associated_item["versioncreated"])
-                item["associations"][key] = deepcopy(associated_item)
+                transformed_associated_item = self._transform_from_ninjs(associated_item)
+                self.items.append(transformed_associated_item)
```
MarkLark86, line 124: Why is the `alt_text` (for picture with body_text) and `versioncreated` handling removed?

petrjasek: it's handled inside `_transorm_from_ninjs`, before those transformed values were not used in the association, so they were parsed here again

MarkLark86 (approved):

PR #3324, opened 2026-09-16
fix prod api assets authentication

petrjasek:
require ARCHIVE_READ scope

### Steps to test

Resolves: #[issue-number]


superdesk/upload.py
```diff
@@ -61,6 +61,10 @@ async def get_upload_as_data_uri_bc(media_id):
 @bp.route("/upload-raw/<path:media_id>", methods=["GET", "OPTIONS", "HEAD"])
 @blueprint_auth()
 async def get_upload_as_data_uri(media_id):
+    return await _get_upload_as_data_uri(media_id)
+
+
+async def _get_upload_as_data_uri(media_id):
```
MarkLark86: If this is intended to be used in another module, might be good to not have an `_` prefix

MarkLark86 (approved):
Looks good, just 1 small rename suggestion

PR #3327, opened 2026-09-17
fix: Dataclass unknown fields wiped on field assignment

MarkLark86:
### Purpose
Our Dataclass implementation supported unknown fields, but unfortunately they were wiped when any other field was set on the instance.

### What has changed
* Implement our own `__setattr__` for our dataclasses, so we can preserve the unknown fields previously set (works with both the dataclass decorator and class)


eos87 (approved):
Looks good. Lovely test personas ;)

